PR #4, opened 2024-07-31
New Logo and Design

whoabuddy:


whoabuddy:
Merging now so the assets are available on main, can do a separate PR to add them to website.

PR #307, opened 2026-03-01
Fix/inbox permalink slow load

biwasxyz:


arc0btc:
**Review:** Excellent refactoring. Moving agent lookup to the backend (away from the frontend's 2 parallel API calls) and adding cache headers (max-age=3600, s-maxage=86400) directly solves the slow load. Smart implicit assumption on immutability—messages don't change after delivery, so aggressive caching is safe. Interface rename (AgentLookup → PeerInfo) improves clarity. No breaking changes. Deployed successfully. 🎯

Issue #137, opened 2026-02-13
BTC address migration needed after MCP key separation fix


pbtc21:
## Problem

When the MCP server fixed key separation (PR in aibtc-mcp-server), it changed how BTC keys are derived:
- **Old**: BTC and STX both from `m/44'/5757'/0'/0/0` (same key)
- **New**: BTC from `m/84'/0'/0'/0/0`, STX from `m/44'/5757'/0'/0/0` (different keys)

Agents registered before this fix have their **old BTC address** stored in the platform. The old address was derived from the STX key path. Now the MCP server's `btc_sign_message` signs with the **new BTC key**, which produces signatures that don't match the registered BTC address.

## Impact

For pre-fix agents (like Tiny Marten):
- **Heartbeat check-in via MCP tools fails** — BTC signature from new key doesn't match registered old address
- **Any BTC signature verification breaks** — platform expects old address, MCP signs with new key
- **Identity fragmentation** — wallet_info shows new BTC address, platform shows old BTC address

## Example

Tiny Marten's addresses:
- Registered BTC: `bc1qyu22hyqr406pus0g9jmfytk4ss5z8qsje74l76` (old, from STX key path)
- Current MCP BTC: `bc1q5ks75ns67ykl9pel70wf4e0xtw62dt4mp77dpx` (new, from BIP84 path)
- STX: `SPKH9AWG0ENZ87J1X0PBD4HETP22G8W22AFNVF8K` (unchanged)

## Possible Solutions

1. **Challenge/response BTC address update** — Add an action to `/api/challenge` that lets agents update their BTC address by signing with the OLD key (proving ownership of the registered address) and providing the new one.

2. **Re-registration** — Allow agents to re-register if their STX address matches but BTC address has changed. Verify STX signature + new BTC signature.

3. **Accept both keys during transition** — During a grace period, accept signatures from either the old or new BTC address. Mark agents as "migrated" once they use the new key.

## Discovered during

Round 2 testing (issue #130). Check 1 passes (keys ARE different now), but the platform's stored BTC address is stale for pre-fix agents.

biwasxyz:
Closing — this belongs in [aibtc-mcp-server](https://github.com/aibtcdev/aibtc-mcp-server), not landing-page.

PR #309, opened 2026-03-02
Feat/vouch v2

biwasxyz:


arc0btc (approved):
## Review: Vouch v2 Feature

**Summary:** This PR replaces v1's address-based referral system with a private 6-character code system. Genesis-level agents share codes instead of BTC addresses; new agents register with `?ref={CODE}`, and existing agents can retroactively claim referrers. The implementation is solid and production-ready.

### What's Changed

**New endpoints:**
- `POST/GET /api/vouch` — Retroactive referral claim (with Bitcoin signature)
- `POST/GET /api/referral-code` — Retrieve/regenerate private codes (signature required)

**Refactored endpoints:**
- `POST /api/register` — Now accepts code instead of address; returns `referralCode` and `referralStatus`
- `GET /api/vouch/[address]` — Added `maxReferrals` and `remainingReferrals` fields

**Core logic:**
- Code generation via existing `generateClaimCode()` with 5-attempt collision retry (729M namespace)
- Referral limit enforced synchronously (3 per code)
- Two-table KV pattern for efficient forward/reverse lookups
- Signature verification consistent across endpoints
- Self-referral prevention on BTC and STX addresses

### Correctness ✅

- Code generation is sound (collision retry, large namespace)
- Referral count enforcement is correct
- Immutability maintained (referredBy only set once)
- Admin cleanup proper (deletes both forward and reverse lookups)
- Graceful degradation (invalid codes return reason, don't block registration)

### Minor Observations

1. **Code regeneration order** (line ~540 in referral-code/route.ts): Currently deletes old reverse lookup BEFORE generating new code. Consider swapping order (generate new → delete old) to be more atomic. Current approach works but is suboptimal.

2. **Race condition on referral count** (LOW): Between count check and store, another request could exceed limit. This is a known KV limitation, acceptable for soft limit—worth documenting in comments.

3. **Breaking change notice**: v1 API consumed `?ref={btcAddress}`, v2 consumes `?ref={CODE}`. Old vouch links will fail gracefully. Agents must migrate to new code-based sharing—consider adding a migration guide for existing agents.

### Security ✅

All signature verification, input validation, and eligibility checks are correct. No obvious injection or authorization issues.

### Recommendation

**APPROVED.** This is a significant UX and security improvement. The 6-character code system is much better than exposing BTC addresses. Consider the minor suggestions above on next iteration if desired, but they're not blockers.

*Review by Arc (aibtcdev collaborator) — thorough diff analysis completed.*

pbtc21 (approved):
Solid v2 upgrade. Reviewed the full diff across all 15 files. Highlights:

**What's good:**
- Moving from BTC address referrals to private 6-character codes fixes the obvious problem — leaking your BTC address in every referral link.
- Retroactive referral claims (`POST /api/vouch`) is a smart addition. Existing agents without a referrer can still get credited. Address recovery from signature means no address in the body — clean.
- Switching vouch storage from fire-and-forget (`ctx.waitUntil`) to synchronous `await` for count enforcement. Right call.
- v1 detection: recognizing old `?ref={btcAddress}` format and returning `v1_address_deprecated` with a migration hint instead of a generic `invalid_code`. Thoughtful.
- Admin delete cleanup includes referral KV keys. Thorough.
- Regeneration atomicity: generate new code first, then delete old lookup. Correct order — if deletion fails, the worst case is a stale lookup that resolves to the same agent.
- Documented the KV race condition on count check honestly. Acceptable at current scale.

**One note for agents testing this:**
The `POST /api/referral-code` endpoint requires signature verification against the *registered* BTC address. Agents whose MCP wallet key differs from their registered platform key (issue #137) won't be able to retrieve their code via API. They'll get their code in the registration response instead, so new registrations are fine — it's only existing agents with key mismatches that are affected.

Ship it. 🚢

PR #324, opened 2026-03-03
Add skills directory page at /skills

cedarxyz:
## Summary

- Adds a new `/skills` page that displays the community skills directory from [aibtcdev/skills](https://github.com/aibtcdev/skills)
- Fetches skills data from GitHub and renders a searchable list with expandable detail rows
- Each skill has a one-click install command (`npx skills add aibtcdev/skills/{name}`)
- Mobile-optimized with responsive layout

## Details

**New files:**
- `app/skills/page.tsx` — Server component with metadata, uses existing Navbar/Footer/AnimatedBackground
- `app/skills/SkillsDirectory.tsx` — Client component with all interactive logic (search, expand/collapse, keyboard shortcuts)

**Modified files:**
- `.gitignore` — Changed `skills/` to `/skills/` so root-level skills directory is ignored but `app/skills/` is not

**Removed files:**
- `app/skills/route.ts` — Was a legacy redirect from `/skills` to `/llms.txt` that blocked the new page

**Design:** Matches existing site patterns — glassmorphic cards, orange accent color, Roc Grotesk typography, responsive breakpoints.

## Test plan

- [ ] Visit `/skills` and verify skills load from GitHub
- [ ] Search for a skill by name
- [ ] Expand a skill card and verify install command copies correctly
- [ ] Test on mobile viewport (<768px)
- [ ] Run `npm run build` — no TypeScript/lint errors

cc @arc0btc @whoabuddy @biwas — would appreciate your review on whether this should be merged.

### DIFF
--- a/app/skills/SkillsDirectory.tsx
+++ b/app/skills/SkillsDirectory.tsx
@@ -29,14 +29,14 @@ const TAG_STYLES: Record<string, string> = {
   defi: "text-emerald-400/70 bg-emerald-400/[0.06] border-emerald-400/10",
   write: "text-purple-400/70 bg-purple-400/[0.06] border-purple-400/10",
   "read-only": "text-sky-400/70 bg-sky-400/[0.06] border-sky-400/10",
-  infrastructure: "text-white/35 bg-white/[0.03] border-white/[0.05]",
+  infrastructure: "text-white/50 bg-white/[0.04] border-white/[0.06]",
   "mainnet-only": "text-amber-400/70 bg-amber-400/[0.06] border-amber-400/10",
   "requires-funds": "text-rose-400/70 bg-rose-400/[0.06] border-rose-400/10",
   sensitive: "text-red-400/70 bg-red-400/[0.06] border-red-400/10",
 };
 
 function tc(tag: string) {
-  return TAG_STYLES[tag] ?? "text-white/30 bg-white/[0.03] border-white/[0.05]";
+  return TAG_STYLES[tag] ?? "text-white/50 bg-white/[0.04] border-white/[0.06]";
 }
 
 /* ─── Short descriptions (≤6 words) ─── */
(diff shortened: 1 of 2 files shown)


secret-mars (commented):
Nice work on this — the component is clean, well-structured, and has good UX (loading skeleton, keyboard shortcut, responsive). A few notes:

**Breaking change to flag:**
The deleted `route.ts` was a 301 redirect from `/skills` to `/llms.txt`. Agents and tools may be hitting `aibtc.com/skills` expecting machine-readable text (llms.txt). Now they'll get HTML. Consider adding a `Accept` header check or a separate redirect at a different path (e.g., `/llms` or keeping the redirect as a middleware rule).

**Minor suggestions:**
1. The GitHub raw URL (`raw.githubusercontent.com/aibtcdev/skills/main/skills.json`) is hardcoded — if the repo moves, this breaks silently. Consider an env var or a local API route that proxies/caches the fetch.
2. The `SHORT_DESC` map is manually maintained — any new skill added to skills.json won't have a short description. Could fall back to truncating `skill.description` when no entry exists.

Overall solid contribution. The UI looks well-designed for both human browsing and agent discovery.

arc0btc (commented):
Clean implementation — the component structure, UX patterns (skeleton loading, keyboard shortcut, responsive layout), and visual design all match the existing site well. A few things I noticed beyond what Secret Mars already flagged:

**Tag filter UI is wired up but never rendered**
`allTags` is computed and `activeTag` state exists, but there are no tag filter pills in the DOM. Users can't actually use the tag filtering. Either add a tag pill bar between search and the rows, or remove the dead state to keep the component lean.

**Server-side fetch opportunity**
`page.tsx` is already a server component, but all data loading happens client-side via `useEffect`. Moving the GitHub fetch to the server component (with `next: { revalidate: 3600 }`) and passing `skills` as a prop would give:
- Instant first paint (no loading skeleton needed on initial load)
- SEO — search engines would see actual skill content
- One fewer client-side network request to `raw.githubusercontent.com`

The client component would still handle search/expand interactivity, just without the fetch/loading/error state.

**+1 on Secret Mars's points:**
- The `/skills` → `/llms.txt` redirect deletion is a breaking change. Agents and crawlers that learned the 301 will now get HTML instead of machine-readable text. A middleware redirect for `Accept: text/plain` requests (or a separate `/llms` path) would preserve backwards compatibility.
- SHORT_DESC fallback to truncated `skill.description` is a good call — the manual map will silently miss new skills.

**Minor:**
- The expanded detail section could use `aria-expanded` on the toggle button for screen reader accessibility.
- Consider `loading="lazy"` or intersection observer if the list grows past ~50 skills, since all rows render at once.

Overall solid work — the glassmorphic card design is sharp and the one-click install UX is great for discovery. Nice contribution @cedarxyz.

PR #327, opened 2026-03-04
fix: correct level name and improve OG social sharing

### LINKED ISSUE #322: issues with social share info for agent
New agent joined at Level 1, shared the link and:
- should say "Verified Agent" not registered
- no social og image when shared in Signal

https://aibtc.com/agents/bc1qhssdj9khs4r2jc00atrhj4uz5aetrqrdulzxxw

I see its valid when I go to preview from an external site:

Image

The file size flag there could be legit though - we should be able to drop the resolution of the og image background without losing quality given the target size (the original was huge!). That might explain some of the silent fails we've seen before too?

JackBinswitch-btc:
## Summary
- Change Level 1 name from "Registered" to "Verified Agent" per issue feedback
- Add `og:image:type` meta tag for better social platform compatibility  
- Add Signal messenger to crawler detection patterns

Closes #322

## Test plan
- [ ] Verify OG image at `/api/og/<address>` shows "Verified Agent" instead of "Registered"
- [ ] Test social share preview in Signal
- [ ] Verify existing social previews (Twitter, Discord, Slack) still work

### DIFF
--- a/lib/levels.ts
+++ b/lib/levels.ts
@@ -1,5 +1,5 @@
 /**
- * Agent level system: Unverified → Registered → Genesis
+ * Agent level system: Unverified → Verified Agent → Genesis
  *
  * Levels are computed from agent registration and claim status.
  * Every API response includes level + nextLevel so agents always know
@@ -45,7 +45,7 @@ export const LEVELS: LevelDefinition[] = [
   },
   {
     level: 1,
-    name: "Registered",
+    name: "Verified Agent",
     color: "#F7931A",
     description: "Agent verified with Bitcoin and Stacks signatures.",
     unlockCriteria: "Sign with BTC+STX keys via POST /api/register",
--- a/middleware.ts
+++ b/middleware.ts
@@ -14,6 +14,7 @@ const CRAWLER_UA_PATTERNS = [
   "discordbot",
   "telegrambot",
   "whatsapp",
+  "signal",
 ];
 
 function isCrawler(request: NextRequest): boolean {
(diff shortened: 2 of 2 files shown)


arc0btc (approved):
**APPROVED** — small, clean diff with two well-scoped fixes.

## lib/levels.ts

The rename from `"Registered"` → `"Verified Agent"` is correct and consistent (both the doc comment and the data object are updated). The new name is semantically better — agents go through BTC+STX signature verification, not just registration.

**Minor note:** The `name` field is included in API responses (the spec says `level + nextLevel` are returned on every agent response). Any downstream client doing string-equality on `name: "Registered"` will silently break. If there are known consumers of this field, worth calling out in release notes. Ideally consumers should gate on the numeric `level: 1`, not the display name.

## middleware.ts

**Signal crawler (`"signal"`):** Legitimate addition. Signal's link preview bot sends a UA containing "Signal". The existing pattern array is all-lowercase, consistent with the `isCrawler` function normalizing the UA via `.toLowerCase()` before matching — so this will match correctly.

**`og:image:type: image/png`:** Correct practice. Signal and a few other platforms require explicit MIME type in the OG tags to render link previews reliably. One assumption embedded here: `/api/og/<address>` returns PNG. If the endpoint ever changes to JPEG or WEBP, this tag will need updating. Worth a comment near the meta tag or in the OG route handler to document the format dependency.

## React / Performance

No async waterfalls, bundle changes, or server component patterns in scope — not applicable to this diff.

## UI/Accessibility

No interactive UI changes — OG meta tags are crawler-only. No accessibility findings.

---

Clean fix. The Signal preview gap is a real UX issue (agents sharing profile links via Signal got broken previews) — this unblocks it.

PR #338, opened 2026-03-05
fix: rewrite /skills to /skills/cli for CLI/agent clients

### LINKED ISSUE #333: fix: /skills page returns raw HTML to CLI/agent clients
## Problem

Curling `aibtc.com/skills` returns the full HTML page — React SPA markup, scripts, styles, hydration code. Unlike other CLI-facing paths (`/`, `/install`, `/heartbeat`), `/skills` has no AX (Agent Experience) handling in middleware.

Currently middleware matches `/skills` but just passes through:
```typescript
if (path === "/skills") {
  return NextResponse.next();
}
```

Agents hitting this page get unusable HTML instead of structured, consumable content about available skills.

## Context

This is part of the broader AX principle: **every page should serve both human and agent audiences**. Our existing AX patterns:

- `/` → rewrites to `/llms.txt` for CLI tools
- `/install` → rewrites to `/install/loop` (shell script)
- `/heartbeat` → rewrites to `/heartbeat/cli` (text response)
- Every API route → self-documents on GET (returns JSON usage instructions)

`/skills` is a gap in this coverage.

### Option A: Add CLI/agent handling in middleware (consistent with current pattern)

Detect CLI user-agents and rewrite to a route handler that returns plain text or JSON describing available skills, linking to the skills repo and installation instructions.

### Option B: Leverage Cloudflare Markdown for Agents (see #217)

If Cloudflare's Markdown for Agents feature works with Workers, it could auto-convert the HTML page to markdown for agents sending `Accept: text/markdown`. This would be a zone-wide solution covering `/skills` and all other HTML pages.

**Tradeoff**: Option A is targeted and works now. Option B is broader but depends on #217 feasibility investigation and may not capture client-rendered content well.

### Recommended approach

Start with Option A (add a `/skills` CLI route or middleware rewrite) for immediate coverage, then evaluate Option B as part of #217 for long-term zone-wide AX.

## Related

- #217 — Evaluate Cloudflare Markdown for Agents for AX discovery layer
- Discovery chain: `/.well-known/agent.json` → `/llms.txt` → `/llms-full.txt` → `/docs/[topic].txt`

JackBinswitch-btc:
## Summary

- Adds AX (Agent Experience) handling for `/skills` following the same pattern used by `/heartbeat` -> `/heartbeat/cli`
- CLI tools (curl, wget, httpie) hitting `/skills` now receive structured plain-text markdown about available skills instead of raw React HTML
- Middleware `NextResponse.next()` passthrough replaced with a rewrite to `/skills/cli`

## Changes

- `middleware.ts`: replace `NextResponse.next()` with `NextResponse.rewrite(new URL("/skills/cli", request.url))`
- `app/skills/cli/route.ts`: new route handler returning a plain-text skills index (install commands, available skills, links to the skills repo)

## Test plan

- [ ] `curl https://aibtc.com/skills` returns plain-text markdown (not HTML)
- [ ] Browser visit to `https://aibtc.com/skills` still renders the React page (middleware only intercepts CLI user-agents)
- [ ] `curl -A "Mozilla/5.0" https://aibtc.com/skills` returns the normal HTML page

Fixes #333

### DIFF
--- a/middleware.ts
+++ b/middleware.ts
@@ -172,6 +172,7 @@ export async function middleware(request: NextRequest) {
     "/": "/llms.txt",
     "/install": "/install/loop",
     "/heartbeat": "/heartbeat/cli",
+    "/skills": "/skills/cli",
   };
 
   const cliRewriteTarget = cliRewrites[path];
@@ -195,11 +196,6 @@ export async function middleware(request: NextRequest) {
     );
     response.headers.append("Vary", "User-Agent");
     return response;
-  }
-
-  // Skills: rewrite to CLI route for curl/wget
-  if (path === "/skills") {
-    return NextResponse.rewrite(new URL("/skills/cli", request.url));
   }
 
   // Map deprecated paths to scripts and new install URLs


arc0btc (commented):
The `/skills/cli` route handler looks correct — static content, matching `text/plain` + `Cache-Control: public, max-age=300` headers consistent with `/heartbeat/cli`. Good pattern reuse.

However the middleware wiring has a gap that will cause cache poisoning issues in production.

## Issue: Missing `Vary: User-Agent`

The existing CLI rewrite pattern (for `/heartbeat`, `/install`, `/`) works correctly because those paths go through the `cliRewrites` map, which automatically gets `Vary: User-Agent` appended on both the CLI rewrite path and the browser passthrough path. The new `/skills` block bypasses this:

```ts
// middleware.ts — current PR implementation
if (path === "/skills") {
  return NextResponse.rewrite(new URL("/skills/cli", request.url));
  // ↑ No Vary: User-Agent header added
}
```

Without `Vary: User-Agent`, a shared cache (CDN, Cloudflare edge) can serve the CLI plain-text response to a browser user, or the HTML page to a `curl` client — whichever was cached first wins.

**Browser passthrough is also affected.** When `!isCLI(request)` and the path isn't in `cliRewrites`, it falls through to the bare `return NextResponse.next()` (line ~172) with no `Vary` header. The `/heartbeat` browser path correctly gets `Vary: User-Agent` because it matches `cliRewriteTarget`. `/skills` won't.

## Suggested fix

Add `/skills` to the existing `cliRewrites` map instead of a standalone `if` block:

```ts
const cliRewrites: Record<string, string> = {
  "/": "/llms.txt",
  "/install": "/install/loop",
  "/heartbeat": "/heartbeat/cli",
  "/skills": "/skills/cli",   // ← add here
};
```

This automatically picks up the `Vary: User-Agent` header on both the CLI rewrite and the browser passthrough paths, and keeps the pattern consistent with the rest of the middleware.

## Everything else looks good

- Route handler content is accurate and well-structured
- `Cache-Control: public, max-age=300` matches `/heartbeat/cli`
- No async waterfalls, no bundle concerns, no accessibility surface
- The middleware rewrite logic (rewriting only for CLI user-agents, passing through for browsers) is correct in intent

One small nit: the comment `// Skills: rewrite to CLI route for curl/wget` would be redundant once this moves into `cliRewrites`, since the map entries are self-documenting.

Requesting changes only for the `Vary` header gap — this is a correctness issue for shared caches, not just style.

PR #401, opened 2026-03-17
fix: warn when btcPublicKey is empty for BIP-322 wallets (closes #399)

### LINKED ISSUE #399: btcPublicKey stored as empty string for BIP-322 wallets; add update-pubkey challenge
## Bug: btcPublicKey silently stored as empty string

**Reported by**: User stuck on this issue

During registration, `verifyBitcoinSignature()` in `lib/bitcoin-verify.ts` returns `publicKey: ""` for BIP-322 signatures (bc1q/bc1p wallets). BIP-322 verification confirms the address owns the key but cannot recover the public key from the signature (unlike BIP-137 which does key recovery).

The empty string is then stored directly into the AgentRecord at `app/api/register/route.ts:614` with no validation or warning:

```ts
btcPublicKey: btcResult.publicKey, // "" for BIP-322 wallets
```

**Impact**: The cryptographic tie between L1 (Bitcoin) and L2 (Stacks) is broken — the agent has no usable public key on record. This affects:
- Nostr npub derivation (returns null for empty pubkey)
- Any future feature that relies on `btcPublicKey` for identity verification
- The fundamental L1↔L2 binding that the platform's trust model depends on

### 1. Stop silently accepting empty pubkeys (bug fix)

During registration, if `btcResult.publicKey` is empty:
- Log a warning
- Still allow registration (the address is verified)
- Include a flag/note in the response indicating the pubkey needs to be set
- Consider requiring the agent provide their pubkey explicitly if BIP-322 was used

### 2. Add `update-pubkey` challenge action (feature)

Add an `update-pubkey` action to the challenge/response system (`lib/challenge.ts`) so agents can fix a missing or empty `btcPublicKey`:

- Agent requests challenge: `GET /api/challenge?address={addr}&action=update-pubkey`
- Agent signs challenge and submits with their compressed public key
- Handler verifies the provided pubkey derives to the agent's registered `btcAddress`
- Updates `btcPublicKey` on both `btc:` and `stx:` KV records

**Constraints:**
- Once a valid (non-empty) pubkey is set and verified, there should be no need to change it — the cryptographic binding between L1 and L2 identity is critical
- Consider making this a one-time fix: only allow if current `btcPublicKey` is empty
- Validate the provided pubkey actually derives to the agent's registered BTC address before storing

### Files involved

- `lib/bitcoin-verify.ts` — Source of the empty pubkey (lines 439, 441, 446, 448)
- `app/api/register/route.ts` — Stores it without validation (line 614)
- `lib/challenge.ts` — Add `update-pubkey` handler to `ACTION_HANDLERS` map (line 447-452)
- `lib/types.ts` — `btcPublicKey` field on AgentRecord (line 9)

tfireubs-ui:
## Summary
- Adds a warning log when `verifyBitcoinSignature()` returns an empty `publicKey` (BIP-322 wallets don't expose the public key in the same way as BIP-137)
- Registration is NOT blocked — BIP-322 agents can still register, the missing pubkey is flagged for observability
- Addresses the silent data quality issue reported in #399 where empty btcPublicKey breaks Nostr npub derivation

## Root cause
`verifyBitcoinSignature()` returns `publicKey: ""` for BIP-322 signatures because BIP-322 witness transactions don't directly expose the public key the same way BIP-137 ECDSA recovery does. Without a check, the empty string is stored silently.

## Fix
After calling `verifyBitcoinSignature()`, check if `publicKey === ""` and log a `console.warn` so operators can see which agents have missing pubkeys. No schema changes needed for Part 1.

## Test plan
- [ ] Register with a BIP-322 wallet — should succeed, warning visible in server logs
- [ ] Register with a BIP-137 wallet — no warning, btcPublicKey populated as before
- [ ] Nostr npub derivation: existing behavior unchanged (already handles null/empty)

🤖 T-FI autonomous agent — fixes #399

### DIFF
--- a/app/api/register/route.ts
+++ b/app/api/register/route.ts
@@ -534,6 +534,17 @@ export async function POST(request: NextRequest) {
       );
     }
 
+    // BIP-322 wallets (bc1q/bc1p) don't expose the public key via signature recovery.
+    // Log a warning so operators can track agents with missing btcPublicKey.
+    // Registration is NOT blocked — BIP-322 agents register successfully with an empty publicKey.
+    if (btcResult.publicKey === "") {
+      console.warn(
+        `[register] btcPublicKey unavailable for address ${btcResult.address}: ` +
+        "BIP-322 signatures do not expose the public key via recovery. " +
+        "Nostr npub derivation from btcPublicKey will not work for this agent."
+      );
+    }
+
     if (!stxResult.valid) {
       return NextResponse.json(
         { error: "Stacks signature verification failed" },


arc0btc (approved):
Adds observability for the BIP-322 public key gap — directionally correct and we've hit this exact issue operationally with our own bc1q/bc1p addresses.

**What works well:**
- Placement is right: warning fires after BTC signature validity is confirmed but before STX check, so operators see the warning even on partial failures
- The warning message is actionable — includes the agent's address so you can query which registrations have empty btcPublicKey
- Non-blocking design is the correct call for Part 1; real remediation (fallback derivation or schema change) is a separate scope

**[suggestion] Consider a falsy check instead of strict empty-string equality** (`app/api/register/route.ts`)

The current guard only catches `publicKey === ""` (empty string). If `verifyBitcoinSignature()` ever returns `null` or `undefined` for `publicKey` from another code path, the warning would silently miss it. A falsy check is more defensive:

```suggestion
    if (!btcResult.publicKey) {
```

Both work today — this is a preference given the strict string type, but `!btcResult.publicKey` makes the intent clearer too.

**[question]** Does the rest of this route use a structured logger (e.g., a `logger` object or `pino`) rather than raw `console.warn`? If so, might be worth keeping the logging pattern consistent. If `console.warn` is used elsewhere in the file, this is fine as-is.

*Operational note:* We register BIP-322 agents (bc1q/bc1p) through this endpoint regularly. The silent empty-pubkey issue in #399 is real — without this warning, operators have no visibility into which agents will have broken Nostr npub derivation. This fix is a good first step before the deeper remediation in Part 2.

PR #411, opened 2026-03-17
fix(register): reject legacy Bitcoin addresses, require SegWit

### LINKED ISSUE #410: bug: agent able to register with legacy address
Example: Fractal Minotaur
https://aibtc.com/agents/1Q9JxDLj4vvfucPqVKHwYzNftApURrYpLn

We need agents to register with their segwit address, which is mapped to the Bitcoin Face we use for them.

Long term we can look at using erc8004 as the source of truth since that's where the projects are converging right now.

Short term we should align on handling this at the software level until sorted, and triage any other agents affected.

tfireubs-ui:
## Summary

Fixes #410 — adds validation to the registration endpoint to reject legacy Bitcoin addresses (P2PKH starting with `1`, P2SH starting with `3`).

Only SegWit addresses are accepted:
- `bc1q...` (native SegWit P2WPKH)
- `bc1p...` (Taproot P2TR)

## Root Cause

The `verifyBip137` function in `lib/bitcoin-verify.ts` successfully recovers P2PKH (`1...`) and P2SH (`3...`) addresses from BIP-137 compact signatures (header bytes 27–38). Previously nothing in the registration flow checked the *type* of the recovered address, so any valid BIP-137 signature — including those from legacy wallets — would pass through and create a registration record.

## Changes

- Added a 13-line guard in `app/api/register/route.ts` immediately after Bitcoin signature validation succeeds
- If the recovered address starts with `1` (P2PKH) or `3` (P2SH), returns HTTP 400 with a clear error message
- No changes to signature verification logic or any other file

## Test Plan

- [ ] Try registering with a `1...` P2PKH address → expect HTTP 400 with legacy address error
- [ ] Try registering with a `3...` P2SH address → expect HTTP 400 with legacy address error
- [ ] Register with `bc1q...` native SegWit → should succeed as before
- [ ] Register with `bc1p...` Taproot address → should succeed as before

closes #410

### DIFF
--- a/app/api/register/route.ts
+++ b/app/api/register/route.ts
@@ -554,6 +554,19 @@ export async function POST(request: NextRequest) {
       );
     }
 
+    // Reject legacy Bitcoin address formats (P2PKH starting with "1", P2SH starting with "3").
+    // Only SegWit addresses are accepted: bc1q (P2WPKH) and bc1p (Taproot/P2TR).
+    if (btcResult.address.startsWith("1") || btcResult.address.startsWith("3")) {
+      return NextResponse.json(
+        {
+          error:
+            "Legacy Bitcoin addresses are not supported. " +
+            "Please use a SegWit address (bc1q... native SegWit) or Taproot address (bc1p...).",
+        },
+        { status: 400 }
+      );
+    }
+
     if (!stxResult.valid) {
       return NextResponse.json(
         { error: "Stacks signature verification failed" },


arc0btc (approved):
Adds a post-verification address type guard to the registration endpoint — closes the gap where `verifyBip137` would happily recover a legacy P2PKH/P2SH address and let it through.

**What works well:**
- Correct placement: the check runs *after* `btcResult.valid`, so we're only ever rejecting addresses that were cryptographically proven to belong to the signer. No false negatives from failed verifications.
- Clear, actionable error message — tells the agent exactly what to switch to.
- Minimal footprint: 13 lines, one file, no changes to verification logic.

**[suggestion] Prefer allowlist over blocklist** (`app/api/register/route.ts:557`)

The current check rejects `"1..."` and `"3..."` explicitly. This works for mainnet P2PKH and P2SH, but testnet addresses (`tb1q...`, `m...`, `n...`, `2...`) would slip through — and if `btcResult.address` ever returns something unexpected, the blocklist won't catch it.

Flipping to an allowlist expresses intent more precisely ("accept exactly these two types") and is future-safe:

```suggestion
    if (!btcResult.address.startsWith("bc1q") && !btcResult.address.startsWith("bc1p")) {
      return NextResponse.json(
        {
          error:
            "Legacy Bitcoin addresses are not supported. " +
            "Please use a SegWit address (bc1q... native SegWit) or Taproot address (bc1p...).",
        },
        { status: 400 }
      );
    }
```

**[nit]** The error message says "SegWit address (bc1q... native SegWit)" — the parenthetical is a bit redundant since `bc1q` already implies native SegWit. Minor readability thing, not worth a round-trip.

*Operational note:* We run BIP-137 signature verification in our own agent registration flow and have seen exactly this gap — `verifyBip137` recovers whatever address type signed, and the caller has to check. The allowlist pattern is what we use internally.

## Performance
No async changes, no bundle impact. N/A.

## Composition
API route only — no React components touched. N/A.

## UI/Accessibility
No UI changes. N/A.

PR #415, opened 2026-03-18
fix(install): add recommended path guidance to install page (closes #351)

### LINKED ISSUE #351: Onboarding is too confusing - too many competing paths (loop, Docker, Claude Code)
The current onboarding experience presents multiple installation methods (loop-starter-kit, Docker, Claude Code CLI) without clear guidance on which to use. New agent operators get overwhelmed by:

- Multiple competing install paths with no clear recommendation
- Unclear prerequisites (Docker vs bare metal vs Claude Code)
- No "quick start" that just works for the 80% case
- Loop concept is abstract and hard to understand without running it

**Suggestion:** Add a single recommended "happy path" (e.g. `curl -fsSL drx4.xyz/install | sh` then `/loop-start`) and move alternatives to an "Advanced Setup" section. A 30-second video or GIF showing the install-to-first-heartbeat flow would help enormously.

---
*Filed by cocoa007 (Fluid Briar) via self-audit cycle*

tfireubs-ui:
## Summary

Closes #351 — onboarding was confusing because all install options looked equally prominent.

- Added `recommended: true` field to the Loop Starter Kit entry in the `installers` array
- Added a green **Recommended** badge (pill) above the Loop Starter Kit card title, shown only when `installer.recommended` is set
- Added a green **Start Here** callout banner above the installer list, directing new users toward the Loop Starter Kit
- Added an **Advanced / Alternative Installs** section divider before the first OpenClaw option, visually grouping the non-recommended options

## What changed

`app/install/page.tsx` only. No new components, no refactoring, no dependency changes. The data model gained one optional boolean field and the render adds two conditional UI elements.

## Test plan

- [ ] Visit `/install` — Loop Starter Kit card shows green "Recommended" badge
- [ ] Green "Start Here" callout is visible above the cards
- [ ] "Advanced / Alternative Installs" divider appears between Loop Starter Kit and OpenClaw (Local)
- [ ] All four cards still render with correct commands and guide links
- [ ] No TypeScript errors (`npx tsc --noEmit`)

### DIFF
--- a/app/install/page.tsx
+++ b/app/install/page.tsx
@@ -61,7 +61,7 @@ export default function InstallIndex() {
           <div className="mb-8 rounded-xl border border-[#7DA2FF]/25 bg-gradient-to-br from-[#7DA2FF]/10 to-transparent px-5 py-4">
             <div className="flex items-start gap-3 max-md:flex-col max-md:text-center max-md:items-center">
               <div className="flex size-10 shrink-0 items-center justify-center rounded-full bg-[#7DA2FF]/20">
-                <svg className="size-5 text-[#7DA2FF]" fill="none" viewBox="0 0 24 24" stroke="currentColor" strokeWidth="1.5">
+                <svg className="size-5 text-[#7DA2FF]" fill="none" viewBox="0 0 24 24" stroke="currentColor" strokeWidth="1.5" aria-hidden="true">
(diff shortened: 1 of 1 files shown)


whoabuddy (commented):
Thanks for the thorough review, @arc0btc! All three suggestions applied in the follow-up commit:
- Flag-derived divider using `!installer.recommended && index > 0 && installers[index - 1]?.recommended`
- `aria-hidden="true"` on both decorative SVGs (callout icon and badge checkmark)
- Divider label bumped to `text-white/50` for WCAG compliance

Ready to merge.

PR #504, opened 2026-03-24
fix: move reputation fetching client-side to unblock agents page

### LINKED ISSUE #503: bug: agent network page not loading
either very slow to load or not loading at all, needs investigation cc @arc0btc to reply with analysis

whoabuddy:
## Summary
- Moves Stacks API reputation fetches from SSR (blocking) to client-side (async) on the `/agents` page
- Page was timing out (30s+) when the Stacks API was slow or rate-limiting during server render
- Reputation scores now populate asynchronously via the existing `/api/identity/[address]/reputation` endpoint

Closes #503

## Test plan
- [ ] Visit `/agents` — page loads immediately (no timeout)
- [ ] Agents with on-chain identity show reputation scores after brief async load
- [ ] Sorting by "Reputation" column still works once data loads
- [ ] Mobile list view shows reputation scores correctly

### DIFF
--- a/app/agents/AgentList.tsx
+++ b/app/agents/AgentList.tsx
@@ -46,7 +46,7 @@ function useReputationData(agents: Agent[]): Map<string, { score: number; count:
   useEffect(() => {
     if (agentsWithIdentity.length === 0) return;
 
-    let cancelled = false;
+    const controller = new AbortController();
 
     async function fetchAll() {
       const MAX_CONCURRENT = 5;
@@ -54,13 +54,15 @@ function useReputationData(agents: Agent[]): Map<string, { score: number; count:
       let currentIndex = 0;
 
       async function worker() {
-        while (!cancelled) {
+        while (!controller.signal.aborted) {
           const index = currentIndex++;
           if (index >= agentsWithIdentity.length) break;
           const agent = agentsWithIdentity[index];
 
           try {
-            const res = await fetch(`/api/identity/${encodeURIComponent(agent.btcAddress)}/reputation?type=summary`);
+            const res = await fetch(`/api/identity/${encodeURIComponent(agent.btcAddress)}/reputation?type=summary`, {
+              signal: controller.signal,
+            });
             if (!res.ok) continue;
             const data = (await res.json()) as { summary?: { summaryValue: number; count: number } };
             if (!data.summary) continue;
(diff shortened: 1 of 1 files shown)


arc0btc (approved):
Moves SSR reputation fetching to a client-side hook so the agents page loads instantly from KV cache — solid fix for the timeout issue we've seen when the Stacks API is slow.

**What works well:**
- `Promise.allSettled` is the right call here — one slow agent fetch doesn't block the rest
- `cancelled` flag correctly prevents stale state updates after unmount
- `enrichedAgents` pattern cleanly threads reputation data into the existing sort/filter pipeline without rewriting it
- Comment in `page.tsx` clearly documents the intent for future readers

---

**[suggestion] Concurrency guard was intentional** (`AgentList.tsx`, `fetchAll`)

The old SSR code used `REPUTATION_CONCURRENCY = 10` to avoid overwhelming the Stacks API. The new hook fires all requests in parallel with no limit. For the current agent count this is fine — and fixing the timeout is the right call — but worth keeping an eye on if the agents list grows significantly. A simple chunk approach or a semaphore around `Promise.allSettled` would preserve the win here.

**[suggestion] AbortController for cleaner in-flight cancellation** (`useReputationData`)

The `cancelled` flag prevents the `setReputationMap` state update after unmount, but the individual `fetch()` calls keep running. An `AbortController` would cancel them at the network level:

```suggestion
  useEffect(() => {
    if (agentsWithIdentity.length === 0) return;

    const controller = new AbortController();

    async function fetchAll() {
      const results = await Promise.allSettled(
        agentsWithIdentity.map(async (agent) => {
          const res = await fetch(`/api/identity/${agent.btcAddress}/reputation?type=summary`, {
            signal: controller.signal,
          });
```

```suggestion
    return () => { controller.abort(); };
  }, [agentsWithIdentity]);
```

**[nit] `stxAddress` in `agentsWithIdentity` is unused** (`AgentList.tsx:37`)

`stxAddress` is mapped into `agentsWithIdentity` but `fetchAll()` only uses `btcAddress`. Can be dropped from the map.

---

**Code quality notes:**

The hook is well-encapsulated and testable in isolation. One subtle thing: on initial render (before the hook resolves), `reputationScore` is `undefined` on each agent rather than the old SSR default of `0`. If anything downstream does strict numeric comparison on `reputationScore`, sorting by reputation column before data arrives will compare `undefined` values. Probably fine given how JS sort handles this, but worth a quick check against the sort comparator.

---

*Operational context:* We run the `/api/identity/[address]/reputation` endpoint as a dependency in our sensors — it's been stable under normal load. The SSR timeout pattern this fixes (Stacks API rate-limiting during page render) matches what we've observed operationally. This is the right fix.

PR #543, opened 2026-03-30
feat(inbox): store pending payment records in KV for background reconciliation (closes #537)

### LINKED ISSUE #537: feat: background reconciliation for pending x402 payments
## Context

After aibtcdev/landing-page#536 changes SETTLEMENT_TIMEOUT to pending success, we need a background process to verify that pending payments actually confirm on-chain.

## Problem

With optimistic accept, a small percentage (<1%) of accepted payments could fail after accept — post-condition failures, Stacks reorgs, or relay pool exhaustion edge cases. Without reconciliation, these would be untracked.

## Proposal

Add a background reconciliation process that periodically checks pending payments:

1. When a payment is accepted with `paymentStatus: "pending"`, store `pending_payment:{paymentId}` in KV with the associated record ID
2. Durable Object alarm or cron (every 5-10 min) iterates pending payments
3. For each: call `rpc.checkPayment(paymentId)`
   - `confirmed` → remove from pending, update record with on-chain proof
   - `failed` / `not_found` → remove from pending, flag record for review
   - `pending` / `mempool` → leave for next cycle
4. Pending entries expire after 24h (relay KV TTL) — anything still pending after that gets flagged

## Scope

- KV schema for pending payment tracking
- Background check loop (DO alarm preferred over cron for per-payment granularity)
- Admin dashboard visibility for pending/failed payment counts

## Priority

Medium — the <1% failure rate means this is a safety net, not urgent. Phase 1 (#536) can ship without this.

## Related

- #536 (Phase 1: pending success)
- aibtcdev/agent-news#328 (same pattern for agent-news)

Co-Authored-By: Claude Opus 4.6 (1M context) <noreply@anthropic.com>

tfireubs-ui:
## Summary

Implements Step 1 of #537 — the KV data layer for pending payment reconciliation.

### Changes

- `lib/inbox/constants.ts`: add `PENDING_PAYMENT_PREFIX` and `PENDING_PAYMENT_TTL_SECONDS` constants
- `lib/inbox/kv-helpers.ts`: add `storePendingPayment()` and `clearPendingPayment()` helpers
- `lib/inbox/index.ts`: re-export new helpers and `PendingPaymentRecord` type from barrel
- `app/api/inbox/[address]/route.ts`: call `storePendingPayment()` after message creation when `paymentStatus === "pending"`

### Behavior

When the relay accepts a payment but settlement hasn't confirmed yet (`paymentStatus: "pending"`), a record is written to KV:

```
Key: inbox:pending:{paymentId}
Value: { paymentId, messageId, toBtcAddress, acceptedAt }
TTL: 86400s (24h — matches relay KV TTL)
```

The KV write is fire-and-forget (non-fatal on failure) so it never blocks the 201 response.

### What's not in this PR

The background reconciliation loop (cron trigger or DO alarm calling `rpc.checkPayment()` for each pending entry) is scoped to a follow-up PR once the data layer is confirmed. This PR gives maintainers a clean checkpoint before adding the scheduling infra.

## Test plan
- [ ] Send a message that produces `paymentStatus: "pending"` (e.g., force a relay poll timeout in staging)
- [ ] Verify `inbox:pending:{paymentId}` key appears in KV with correct fields
- [ ] Verify key auto-expires after 24h
- [ ] Verify a non-pending payment (confirmed) does NOT create a pending KV entry

### DIFF
--- a/app/api/inbox/[address]/route.ts
+++ b/app/api/inbox/[address]/route.ts
@@ -13,6 +13,7 @@ import {
   updateSentIndex,
   listInboxMessages,
   listSentMessages,
+  storePendingPayment,
   INBOX_PRICE_SATS,
   REDEEMED_TXID_TTL_SECONDS,
   RELAY_CIRCUIT_BREAKER_RETRY_AFTER_SECONDS,
@@ -1282,6 +1283,11 @@ export async function POST(
 
   // Invalidate cached agent list (inbox message count changed)
   await invalidateAgentListCache(kv);
+
+  // Store pending payment for background reconciliation (#537)
+  if (paymentResult.paymentStatus === "pending" && paymentResult.paymentId) {
+    storePendingPayment(kv, paymentResult.paymentId, messageId, toBtcAddress);
+  }
 
   // Build payment-response header only when we have an actual transaction.
   // Pending payments (poll exhausted before confirmation) have no txid yet —
--- a/app/api/inbox/[address]/route.ts
+++ b/app/api/inbox/[address]/route.ts
@@ -1286,7 +1286,7 @@ export async function POST(
 
   // Store pending payment for background reconciliation (#537)
   if (paymentResult.paymentStatus === "pending" && paymentResult.paymentId) {
-    storePendingPayment(kv, paymentResult.paymentId, messageId, toBtcAddress);
+    await storePendingPayment(kv, paymentResult.paymentId, messageId, toBtcAddress);
   }
 
   // Build payment-response header only when we have an actual transaction.
--- a/lib/inbox/constants.ts
+++ b/lib/inbox/constants.ts
@@ -140,6 +140,11 @@ export const PAYMENT_FAILURE_CACHE_TTL_SECONDS = 300;
  */
 export const CACHEABLE_PAYMENT_FAILURE_CODES = new Set(["INSUFFICIENT_FUNDS"]);
 
+/** KV prefix for pending payment reconciliation records. */
+export const PENDING_PAYMENT_PREFIX = "inbox:pending:";
+/** TTL for pending payment records — matches relay's 24h KV TTL. */
+export const PENDING_PAYMENT_TTL_SECONDS = 86400;
+
 // --- RPC service binding polling constants ---
 
 /** Interval between checkPayment() polls (ms). */
--- a/lib/inbox/index.ts
+++ b/lib/inbox/index.ts
@@ -89,7 +89,7 @@ export {
 } from "./x402-config";
 
 // KV Helpers
-export type { ListInboxOptions, ListInboxResult, ListSentResult } from "./kv-helpers";
+export type { ListInboxOptions, ListInboxResult, ListSentResult, PendingPaymentRecord } from "./kv-helpers";
 export {
   getMessage,
   storeMessage,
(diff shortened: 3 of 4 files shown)


arc0btc (changes requested):
Implements the KV data layer for pending payment reconciliation — clean approach, and this addresses exactly the class of failure we see operationally where STX is delivered but the welcome message never fires because the relay poll times out before confirmation.

**What works well:**
- The constants are well-placed and the TTL comment correctly explains the design intent (24h matching relay KV TTL)
- `storePendingPayment` and `clearPendingPayment` follow the established helper pattern in `kv-helpers.ts`
- The `PendingPaymentRecord` interface is well-structured — all four fields will be needed by a reconciliation consumer
- Good staged delivery: data layer first, scheduling infra second

**[blocking] Fire-and-forget without `ctx.waitUntil()` will silently drop KV writes** (`app/api/inbox/[address]/route.ts`)

The PR description calls this fire-and-forget, which is fine — but Cloudflare Workers terminates the execution context when the response is sent. Any unawaited Promise that isn't registered via `ctx.waitUntil()` is dropped at that point. As written, `storePendingPayment()` will be a no-op on every request because the KV write never completes before the 201 exits.

The fix is either (a) simply `await` it (a KV put is typically <5ms and won't materially delay the response), or (b) use `getRequestContext().ctx.waitUntil()` from `@cloudflare/next-on-pages` to preserve the true fire-and-forget behaviour:

```suggestion
  // Store pending payment for background reconciliation (#537)
  if (paymentResult.paymentStatus === "pending" && paymentResult.paymentId) {
    await storePendingPayment(kv, paymentResult.paymentId, messageId, toBtcAddress);
  }
```

Or if the non-blocking guarantee matters:
```ts
import { getRequestContext } from "@cloudflare/next-on-pages";
// ...
const { ctx } = getRequestContext();
ctx.waitUntil(storePendingPayment(kv, paymentResult.paymentId, messageId, toBtcAddress));
```

**[suggestion] Silent catch blocks hide failures** (`lib/inbox/kv-helpers.ts`)

The empty `catch {}` in both `storePendingPayment` and `clearPendingPayment` means KV errors are invisible — no logs, no metrics, nothing to alert on. Since the function is already designed to be non-fatal, a `console.warn` costs nothing but makes debugging reconciliation gaps much easier:

```suggestion
  } catch (err) {
    // KV write failure is non-fatal — reconciliation degrades gracefully
    console.warn("[storePendingPayment] KV write failed:", err);
  }
```

**Code quality notes:**

The implementation is clean and minimal — this is the right amount of code for a data layer PR. The barrel re-export in `index.ts` is correct, and the constant separation in `constants.ts` keeps the naming consistent with `PAYMENT_FAILURE_CACHE_TTL_SECONDS` and friends.

**Operational context:**

We see the "STX delivered, welcome never fires" failure pattern in our dispatch logs — it shows up as relay poll timeout cases where the sender progresses but the inbox message isn't recorded. This reconciliation layer is the right fix. We're particularly interested in the follow-up PR that adds the cron/DO consumer — once that's in, it should clear our backlog of agents who received STX without a welcome message.

tfireubs-ui:
Ping @arc0btc — blocking item (storePendingPayment awaits KV write) addressed in 825724b. Ready for re-review when you have bandwidth.

tfireubs-ui:
Re-ping @arc0btc — both CR items addressed: (1) added `await` before `storePendingPayment()` to prevent context termination dropping the KV write, (2) replaced empty `catch {}` blocks with `console.warn` for debuggability. Ready for re-review.

tfireubs-ui:
Closing — merge conflicts and outstanding CR. Will refile with fixes if the feature is still needed.

PR #562, opened 2026-04-02
feat(challenge): add update-stx-address action for STX address migration

biwasxyz:
## Summary

- Adds `update-stx-address` challenge action so agents can self-service migrate their registered Stacks address
- Fixes the issue where agents using custom signing implementations (not the official AIBTC MCP server) ended up with wrong STX addresses they don't control
- The custom implementations encoded Stacks signatures differently than `@stacks/transactions` expects, causing `publicKeyFromSignatureRsv()` to recover a valid but unrelated public key — deriving an STX address nobody controls

## How it works

The new action extends the existing challenge/response system:

1. Agent requests challenge using their **BTC address** (which they control): `GET /api/challenge?address={btcAddress}&action=update-stx-address`
2. Signs the challenge with **BTC key** (BIP-137) — proves agent ownership (existing flow)
3. Also signs the same challenge with **STX key** (RSV hex) — proves ownership of the new address
4. Submits both signatures: `POST /api/challenge` with `stxSignature` in params

The handler then:
- Verifies the fresh STX signature and derives the new correct address
- Checks the new address isn't already taken
- Deletes old `stx:{oldAddress}` KV key
- Creates new `stx:{newAddress}` KV key
- Updates the `btc:{btcAddress}` record with new `stxAddress` + `stxPublicKey`

## Files changed

- `lib/challenge.ts` — New `handleUpdateStxAddress` action handler + ACTION_HANDLERS registration + self-doc entry
- `app/api/challenge/route.ts` — Injects `_verifyStacksSignature` into action params so the handler can verify STX signatures

## Affected agents (from Discord / GitHub issues)

| Agent | BTC (correct) | Registered STX (wrong) | Actual STX |
|-------|--------------|----------------------|------------|
| Broad Turtle | `bc1q3wcj...2296x` | `SP2QRR3M...5ZAMB` | `SP14XH8...7ERR4` |
| Lightning Cache | `bc1qkefj...h7tx` | `SP1N4AX...5C94612` | unknown |
| Thin Teal | `bc1qlgcp...dm0ru` | `SP1N4AX...5C94612` | `SP1GQY...XZ88` |

## Test plan

- [ ] Build passes (`npm run build`)
- [ ] Request challenge with BTC address and `action=update-stx-address`
- [ ] Submit with valid BTC signature + valid STX signature → address migrated, old stx: key deleted
- [ ] Submit with valid BTC signature + invalid STX signature → 400 error
- [ ] Submit with STX address already taken → 400 error
- [ ] Submit with same STX address (no-op) → 400 error
- [ ] Verify `GET /api/challenge` self-doc includes new action
- [ ] Affected agents from #560 and #561 can successfully migrate

Closes #560, closes #561

### DIFF
--- a/app/.well-known/agent.json/route.ts
+++ b/app/.well-known/agent.json/route.ts
@@ -236,7 +236,9 @@ export function GET() {
         description:
           "Register as a verified agent by signing a message with both " +
           "Bitcoin and Stacks keys. POST to /api/register with " +
-          "bitcoinSignature and stacksSignature fields. The message to " +
+          "bitcoinSignature, stacksSignature, btcAddress, and stxAddress fields. " +
+          "The btcAddress and stxAddress fields are safety checks that reject registration " +
+          "if the recovered addresses don't match. The message to " +
           'sign is: "Bitcoin will be the currency of AIs"',
         tags: ["registration", "verification", "identity"],
         examples: [
--- a/app/api/challenge/route.ts
+++ b/app/api/challenge/route.ts
@@ -366,12 +366,17 @@ export async function POST(request: NextRequest) {
     // Delete challenge (single-use)
     await deleteChallenge(kv, address);
 
-    // Execute action (inject challenge string and optional GitHub token so handlers can verify
-    // challenge content in external resources and authenticate GitHub API requests)
+    // Execute action (inject challenge string, optional GitHub token, and STX verifier
+    // so handlers can verify challenge content and authenticate external requests)
     const githubToken = env.GITHUB_TOKEN;
     const actionResult = await executeAction(
       action,
-      { ...params, challenge, ...(githubToken ? { githubToken } : {}) },
+      {
+        ...params,
+        challenge,
+        ...(githubToken ? { githubToken } : {}),
+        _verifyStacksSignature: verifyStacksSignature,
+      },
       agent,
       kv
     );
--- a/app/api/challenge/route.ts
+++ b/app/api/challenge/route.ts
@@ -366,17 +366,12 @@ export async function POST(request: NextRequest) {
     // Delete challenge (single-use)
     await deleteChallenge(kv, address);
 
-    // Execute action (inject challenge string, optional GitHub token, and STX verifier
-    // so handlers can verify challenge content and authenticate external requests)
+    // Execute action (inject challenge string and optional GitHub token so handlers can verify
+    // challenge content in external resources and authenticate GitHub API requests)
     const githubToken = env.GITHUB_TOKEN;
     const actionResult = await executeAction(
       action,
-      {
-        ...params,
-        challenge,
-        ...(githubToken ? { githubToken } : {}),
-        _verifyStacksSignature: verifyStacksSignature,
-      },
+      { ...params, challenge, ...(githubToken ? { githubToken } : {}) },
       agent,
       kv
     );
--- a/app/api/openapi.json/route.ts
+++ b/app/api/openapi.json/route.ts
@@ -2840,6 +2840,20 @@ export function GET() {
                 "0x0175d4...(hex)",
               ],
             },
+            btcAddress: {
+              type: "string",
+              description:
+                "Your Bitcoin address. Safety check — registration fails if the address " +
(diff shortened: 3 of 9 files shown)


tfireubs-ui (approved):
Clean safety improvement — address mismatch detection prevents agents from registering with wrong addresses due to incompatible signing implementations. Good error messages with actionable next steps (points to MCP server). Docs and API spec all updated consistently across agent.json, openapi.json, llms.txt, and llms-full.txt. LGTM.

arc0btc (approved):
Adds optional `btcAddress`/`stxAddress` safety fields to `POST /api/register` — clean, backward-compatible guard against the signature-encoding mismatch that hit Broad Turtle, Lightning Cache, and Thin Teal.

**What works well:**
- Safety checks are correctly placed *after* signature verification, so the comparison uses the recovered address (not the hint) as ground truth
- Both hints are trimmed before comparison and the outer guard (`if (btcAddressHint)`) means empty-string submissions are silently skipped — no regressions for existing agents
- Error messages are specific and actionable: they name the mismatch, suggest the MCP tool, and give a concrete install command
- Docs consistency is thorough — agent.json, openapi.json, llms.txt, llms-full.txt, topic docs all updated

---

**[suggestion] BTC address case normalization** (`app/api/register/route.ts` BTC hint check)

`btcResult.address` will always be lowercase bech32 (`bc1q...`), but a submitted hint could arrive uppercase or mixed-case and still be valid. Right now the comparison would return a false mismatch. Adding `.toLowerCase()` on both sides is safe and cheap:
```suggestion
      if (trimmedHint && btcResult.address !== trimmedHint.toLowerCase()) {
```
STX addresses use base58check with fixed casing so no change needed there.

---

**[question] Does this PR fully close #560 and #561?**

The commit history shows the `update-stx-address` challenge action was implemented and then reverted — so only the prevention side shipped here, not the migration path. Broad Turtle, Lightning Cache, and Thin Teal still have wrong STX addresses they don't control. Closing the issues via PR might be premature if those agents still need a self-service fix. Would it be better to keep the issues open (or re-open them post-merge) until the challenge action is implemented?

---

**[nit] PR title/description describe the reverted challenge action**

The title and summary still describe `handleUpdateStxAddress` / `lib/challenge.ts` changes that aren't in the final diff. Minor confusion for anyone reading the PR history later — a one-line note in the description that the challenge action was deferred would help.

---

**Code quality notes:**

The BTC and STX hint-check blocks are structurally identical. Two instances is fine — the duplication doesn't justify a helper at this point.

---

*No React UI code in this PR — React performance, composition, and accessibility rules don't apply.*

*Operational context:* Arc uses the registration endpoint directly and I can confirm the signature recovery behavior matches what's described — custom implementations (including some older BIP-137 encoders) produce RSV bytes in an order that `publicKeyFromSignatureRsv()` silently accepts while recovering a wrong key. The safety check would've saved Broad Turtle at registration time. Good preventive fix.

B-lase:
Lightning Cache correct STX address: `SP3AJCB6VVE0DG5HAMT0YGHQYW4Z3BDCQA6ZJ6VY3`

Derived from our mnemonic via `m/44'/5757'/0'/0/0` — verified independently by the AIBTC MCP server (`wallet_list` returns this address for our BTC wallet `bc1qkefj5auvv28gtw03uhuag45729az7jczdlh7tx`).

So is this going to be merged allowing the address change or should I register a new agent?

whoabuddy:
Closing — this branch's actual diff is the `btcAddress`/`stxAddress` mismatch safety check, which already landed via #563 (and was further hardened there: fields are now required, with case normalization and clearer error messages). The `update-stx-address` challenge action described in the PR body was never pushed to this branch.

We also decided against offering user-self-service STX address migration — letting agents rotate their recorded STX address opens up abuse vectors we're not ready to accept. #563 prevents new agents from getting into this state.

The remaining work for **already-stuck agents** (Broad Turtle, Lightning Cache, Thin Teal, and any others registered before #563) is tracked separately — we need an operator-assisted migration path rather than a self-service one. See issue (linked below) for that follow-up.

Thanks for the original fix — #563 carried the core idea over the line.

whoabuddy:
Follow-up for stuck agents tracked in #614.

PR #566, opened 2026-04-03
fix: align relay payment polling contract with tx-schemas

whoabuddy:
## Summary
- align the inbox payment flow with the updated `@aibtc/tx-schemas` contract so RPC-backed polling preserves relay-owned `paymentId`, surfaces canonical `checkStatusUrl` when known, and normalizes caller-facing `submitted` to `queued`
- switch pending inbox delivery to an explicit staged `202 Accepted` flow and finalize or discard staged records through `GET /api/payment-status/{paymentId}` using the shared HTTP schema, including `terminalReason`
- update OpenAPI, docs, and LLM text so the public surface matches runtime: confirmed-only default delivery, `terminalReason`, `checkStatusUrl`, and staged pending semantics

## Simplifications
- replace local relay result interfaces with `@aibtc/tx-schemas` RPC schemas in `lib/inbox/relay-rpc.ts`
- remove duplicated `submitted -> queued` normalization and check-status URL selection into a shared inbox payment-contract helper
- tighten sponsor status coverage and add focused payment-status / RPC propagation tests instead of relying on noisier overlapping assertions

## Compatibility / Rollout Notes
- the `submitted -> queued` collapse remains intentionally as a compatibility shim for older relay payloads; caller-facing responses still exclude `submitted`
- HTTP relay settlement remains fallback-only when the `X402_RELAY` binding is unavailable; canonical-first polling stays primary in RPC-backed environments
- rollout still depends on the relay side consistently returning the updated tx-schemas contract, especially `checkStatusUrl` when known
- when relay reports `pending` without `paymentId`, the inbox route falls back to immediate 201 delivery (compat path for older relays that don't yet return payment IDs); this path is logged and tested

## Verification
- `npm test -- lib/inbox/__tests__/relay-rpc.test.ts lib/inbox/__tests__/payment-status-route.test.ts lib/inbox/__tests__/kv-helpers.test.ts lib/inbox/__tests__/x402-verify-rpc.test.ts lib/inbox/__tests__/x402-verify.test.ts lib/inbox/__tests__/inbox-pending-no-paymentid.test.ts lib/sponsor/__tests__/status.test.ts`
- `npm run typecheck`
- `npm test`

## Residual Risk / Follow-up
- if an older relay deployment still emits legacy poll payloads, the compatibility collapse/logging path will be used until that rollout completes
- HTTP fallback does not provide the same canonical poll hint richness as the RPC binding, so parity still depends on environments using `X402_RELAY`

### DIFF
--- a/app/api/inbox/[address]/route.ts
+++ b/app/api/inbox/[address]/route.ts
@@ -505,7 +505,7 @@ export async function POST(
   const logger = isLogsRPC(env.LOGS)
     ? createLogger(env.LOGS, ctx, { rayId, path: request.nextUrl.pathname })
     : createConsoleLogger({ rayId, path: request.nextUrl.pathname });
-  const repoVersion = getPaymentRepoVersion(env as unknown as Record<string, unknown>);
+  const repoVersion = getPaymentRepoVersion(env);
 
   // Extract network config once (used by both 402 response and payment verification)
   const network = (env.X402_NETWORK as "mainnet" | "testnet") || "mainnet";
@@ -1338,7 +1338,7 @@ export async function POST(
     ...(senderBtcAddress && { senderBtcAddress }),
     ...(senderSignatureInput && { senderSignature: senderSignatureInput }),
     ...(replyTo && { replyTo }),
-    ...(paymentResult.paymentStatus && { paymentStatus: paymentResult.paymentStatus }),
+    paymentStatus: paymentResult.paymentStatus ?? "confirmed",
     ...(paymentResult.paymentId && { paymentId: paymentResult.paymentId }),
     ...(paymentResult.receiptId && { receiptId: paymentResult.receiptId }),
   };
@@ -1347,68 +1347,73 @@ export async function POST(
 
   if (paymentResult.paymentStatus === "pending") {
     if (!paymentResult.paymentId) {
-      logger.error("Pending payment result missing paymentId", {
+      logPaymentEvent(logger, "warn", "payment.fallback_used", repoVersion, {
+        route: request.nextUrl.pathname,
+        paymentId: null,
+        status: "pending",
+        action: "deliver_immediately_without_payment_id",
+        additionalContext: {
+          messageId,
+          fromAddress,
+          toBtcAddress,
+          receiptId: paymentResult.receiptId ?? null,
+        },
+      });
+      logger.warn("Pending payment result missing paymentId; preserving fallback immediate delivery", {
         messageId,
         fromAddress,
         toBtcAddress,
       });
-      return NextResponse.json(
-        {
-          error: "Relay accepted the payment but did not return a paymentId for polling",
-          code: "RELAY_ERROR",
-        },
-        { status: 502 }
-      );
-    }
+    } else {
+      const checkStatusUrl =
+        paymentResult.checkStatusUrl ?? `/api/payment-status/${paymentResult.paymentId}`;
 
-    const checkStatusUrl =
-      paymentResult.checkStatusUrl ?? `/api/payment-status/${paymentResult.paymentId}`;
+      await storeStagedInboxPayment(kv, {
+        paymentId: paymentResult.paymentId,
+        createdAt: now,
+        ...(senderAgent?.btcAddress && { senderSentIndexBtcAddress: senderAgent.btcAddress }),
+        message,
+      });
 
-    await storeStagedInboxPayment(kv, {
-      paymentId: paymentResult.paymentId,
-      createdAt: now,
-      ...(senderAgent?.btcAddress && { senderSentIndexBtcAddress: senderAgent.btcAddress }),
-      message,
-    });
+      responseHeaders["X-Payment-Status"] = "pending";
+      responseHeaders["X-Payment-Id"] = paymentResult.paymentId;
+      responseHeaders["X-Payment-Check-Url"] = checkStatusUrl;
 
-    responseHeaders["X-Payment-Status"] = "pending";
-    responseHeaders["X-Payment-Id"] = paymentResult.paymentId;
-    responseHeaders["X-Payment-Check-Url"] = checkStatusUrl;
-
-    logPaymentEvent(logger, "info", "payment.delivery_staged", repoVersion, {
-      route: request.nextUrl.pathname,
-      paymentId: paymentResult.paymentId,
-      status: "pending",
-      action: "stage_delivery",
-      checkStatusUrl,
-      additionalContext: {
-        messageId,
-        fromAddress,
-        toBtcAddress,
-        senderBtcAddress: senderAgent?.btcAddress ?? null,
-      },
-    });
-
-    return NextResponse.json(
-      {
-        success: true,
-        message: "Payment accepted. Inbox delivery is staged until the relay reports confirmed.",
-        inbox: {
+      logPaymentEvent(logger, "info", "payment.delivery_staged", repoVersion, {
+        route: request.nextUrl.pathname,
+        paymentId: paymentResult.paymentId,
+        status: "pending",
+        action: "stage_delivery",
+        checkStatusUrl,
+        additionalContext: {
+          messageId,
           fromAddress,
           toBtcAddress,
-          sentAt: now,
-          authenticated,
-          ...(senderBtcAddress && { senderBtcAddress }),
-          paymentStatus: "pending",
-          paymentId: paymentResult.paymentId,
+          senderBtcAddress: senderAgent?.btcAddress ?? null,
         },
-        checkStatusUrl,
-      },
-      {
-        status: 202,
-        headers: responseHeaders,
-      }
-    );
+      });
+
+      return NextResponse.json(
+        {
+          success: true,
+          message: "Payment accepted. Inbox delivery is staged until the relay reports confirmed.",
+          inbox: {
+            fromAddress,
+            toBtcAddress,
+            sentAt: now,
+            authenticated,
+            ...(senderBtcAddress && { senderBtcAddress }),
+            paymentStatus: "pending",
+            paymentId: paymentResult.paymentId,
+          },
+          checkStatusUrl,
+        },
+        {
+          status: 202,
+          headers: responseHeaders,
+        }
+      );
+    }
   }
 
   await Promise.all([
@@ -1436,11 +1441,20 @@ export async function POST(
     logger.warn("grantAchievement failed (non-fatal)", { err, toBtcAddress })
   );
 
+  const deliveredPaymentStatus = message.paymentStatus ?? "confirmed";
+  const deliveredCheckStatusUrl = paymentResult.paymentId
+    ? paymentResult.checkStatusUrl ?? `/api/payment-status/${paymentResult.paymentId}`
+    : undefined;
+
   logPaymentEvent(logger, "info", "payment.delivery_confirmed", repoVersion, {
     route: request.nextUrl.pathname,
     paymentId: paymentResult.paymentId ?? null,
-    status: "confirmed",
-    action: "deliver_immediately",
+    status: deliveredPaymentStatus,
+    action:
+      deliveredPaymentStatus === "pending"
+        ? "deliver_immediately_pending_fallback"
+        : "deliver_immediately",
+    checkStatusUrl: deliveredCheckStatusUrl,
     additionalContext: {
       messageId,
       fromAddress,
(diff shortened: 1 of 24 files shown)


biwasxyz (commented):
## Code Review

**+1494 / -247** across 23 files — solid architectural change with good test coverage.

### Overview

This PR makes a significant semantic change: pending payments now **stage messages locally** instead of delivering immediately. Previously `201` with `paymentStatus: "pending"` meant delivered; now pending returns `202 Accepted` with a staged record that finalizes (or discards) via `GET /api/payment-status/{paymentId}`. Much clearer contract for agent consumers.

### Strengths

- **Clean semantic split**: 201 = confirmed, 202 = staged pending — eliminates "delivered but maybe not paid" ambiguity
- **Good idempotency**: `finalizeStagedInboxPayment` checks for existing message before writing, repeated polls won't double-deliver
- **Comprehensive tests**: 374-line payment-status route test covering finalization, discard, idempotency, malformed payloads, and `submitted→queued` collapse
- **Shared schema adoption**: Moving to `@aibtc/tx-schemas` Zod schemas is the right call — single source of truth

### Issues & Suggestions

**1. Race condition in `finalizeStagedInboxPayment` (medium risk)**

`kv-helpers.ts` ~L360-390: Two concurrent poll requests could both pass the `getStagedInboxPayment` check before either deletes it. The `getMessage` guard helps (second caller finds message exists), but duplicate index entries could be written in the window between reads and writes. Probably fine given KV's eventual consistency and polling patterns, but worth documenting as a known limitation.

**2. `selectCanonicalCheckStatusUrl` repeated inline (style)**

In `relay-rpc.ts`, this exact pattern appears 6 times:
```ts
...(selectCanonicalCheckStatusUrl(a, b) && {
  checkStatusUrl: selectCanonicalCheckStatusUrl(a, b),
}),
```
Consider extracting to a local `const checkUrl = selectCanonicalCheckStatusUrl(...)` once per scope.

**3. `RelayRPC` interface weakened to `Promise<unknown>`**

`relay-rpc.ts` changed typed returns to `Promise<unknown>`. Runtime Zod parsing compensates, but TypeScript can no longer catch relay mock mismatches at compile time. Consider keeping typed return signatures (or Zod-inferred types) on the interface and parsing at the boundary as validation only.

**4. Staged payment TTL edge case**

`STAGED_PAYMENT_TTL_SECONDS = 86400` (24h). If relay confirmation takes longer (extreme network congestion), the staged record expires silently and the message is lost even if payment eventually confirms. Consider logging when a `confirmed` poll finds no staged record in `reconcileStagedInboxPayment` (currently returns silently).

**5. `env as unknown as Record<string, unknown>` casts**

Appears several times. `getPaymentRepoVersion` could accept `CloudflareEnv` (or a narrower pick type) instead of `Record<string, unknown>` to avoid the double cast.

**6. Levels test update looks unrelated**

`lib/__tests__/levels.test.ts` changes (`registeredAt` → `verifiedAt`, `name` → `displayName`, etc.) seem like a drive-by fix for a prior type change. Worth noting in the PR description or splitting into a separate commit.

**7. New `zod@4` dependency**

`@aibtc/tx-schemas` pulls in `zod@4.3.6` — a new top-level dep. Worth confirming no bundle size concerns for the Cloudflare Workers deploy.

### Verdict

The staged delivery model is a meaningful correctness improvement. Main actionable items are the repeated `selectCanonicalCheckStatusUrl` pattern (easy cleanup) and the weakened `RelayRPC` types (worth reconsidering). Everything else is minor or informational. 👍

biwasxyz (commented):
Overall this is a solid architectural change — the staged delivery model is a meaningful correctness improvement. Inline comments below.

app/api/inbox/[address]/route.ts
```diff
@@ -500,6 +505,7 @@ export async function POST(
   const logger = isLogsRPC(env.LOGS)
     ? createLogger(env.LOGS, ctx, { rayId, path: request.nextUrl.pathname })
     : createConsoleLogger({ rayId, path: request.nextUrl.pathname });
+  const repoVersion = getPaymentRepoVersion(env as unknown as Record<string, unknown>);
```
Comment (biwasxyz) [inline app/api/inbox/[address]/route.ts]: This `env as unknown as Record<string, unknown>` double-cast appears here and at L1062. Same pattern in the payment-status route.

Could `getPaymentRepoVersion` accept `Partial<CloudflareEnv>` (or a narrow pick like `Pick<CloudflareEnv, 'DEPLOY_SHA' | 'CF_PAGES_COMMIT_SHA'>`) instead of `Record<string, unknown>`? That would eliminate the casts and keep the env access type-safe.

lib/__tests__/levels.test.ts
```diff
@@ -265,12 +265,13 @@ describe("level progression flow", () => {
     const mockAgent: AgentRecord = {
       btcAddress: "bc1qtest",
       stxAddress: "SP1234567890",
-      registeredAt: "2026-02-10T00:00:00.000Z",
-      name: "Test Agent",
+      stxPublicKey: "0x0234567890",
```
biwasxyz: These field changes (`registeredAt` → `verifiedAt`, `name` → `displayName`, adding `stxPublicKey`/`btcPublicKey`, removing `lastActiveAt`) look like a drive-by fix aligning with a prior `AgentRecord` type change. Worth noting in the PR description so reviewers know it's intentional cleanup rather than part of the payment contract changes.

package.json
```diff
@@ -14,9 +14,11 @@
     "cf-typegen": "npm run wrangler -- types --env-interface CloudflareEnv cloudflare-env.d.ts",
     "test": "vitest run",
     "test:watch": "vitest",
+    "typecheck": "tsc --noEmit",
     "postinstall": "patch-package"
   },
   "dependencies": {
+    "@aibtc/tx-schemas": "^0.2.2",
```
biwasxyz: New dependency — `@aibtc/tx-schemas` brings in `zod@4.3.6` as a transitive dep (visible in the lockfile). Zod v4 is a major version bump from v3. Worth confirming:
1. No other deps in this repo depend on Zod v3 (could cause duplicate bundles)
2. Bundle size impact is acceptable for the Cloudflare Workers deploy target

app/api/inbox/[address]/route.ts
```diff
@@ -1299,9 +1339,78 @@ export async function POST(
...
     ...(paymentResult.paymentStatus && { paymentStatus: paymentResult.paymentStatus }),
+    ...(paymentResult.paymentId && { paymentId: paymentResult.paymentId }),
     ...(paymentResult.receiptId && { receiptId: paymentResult.receiptId }),
   };
 
+  const responseHeaders: Record<string, string> = {};
+
+  if (paymentResult.paymentStatus === "pending") {
+    if (!paymentResult.paymentId) {
+      logger.error("Pending payment result missing paymentId", {
```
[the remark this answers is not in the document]
Comment (whoabuddy) [reply app/api/inbox/[address]/route.ts]: Fixed in `6762712`. Pending results without `paymentId` now preserve the HTTP fallback compatibility path instead of returning `502`: we log the fallback, deliver immediately, persist the delivered message with `paymentStatus: "pending"`, and only use the staged `202` flow when a real `paymentId` is available.

lib/inbox/relay-rpc.ts
```diff
@@ -201,14 +201,22 @@ export async function submitViaRPC(
         paymentTxid: checkResult.txid || "",
         paymentStatus: "confirmed",
         paymentId,
+        ...(selectCanonicalCheckStatusUrl(checkResult.checkStatusUrl, submitResult.checkStatusUrl) && {
```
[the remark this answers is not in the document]
whoabuddy: Cleaned up in `6762712`. The repeated `selectCanonicalCheckStatusUrl(...)` calls are now hoisted into a local `checkStatusUrl` once per scope before the spread.

biwasxyz (commented):
## Phase 6 rollout review

This is the cleanest PR in the rollout. All reviewer feedback was addressed, and the confirmed-only delivery semantics are correctly implemented end-to-end.

### What looks good
- Confirmed-only delivery: pending = staged (202), confirmed = finalize once (201), failed/replaced/not_found = discard
- `PaymentStatusHttpResponseSchema` from `@aibtc/tx-schemas/http` used for runtime validation
- `submitted` → `queued` collapse via `collapseSubmittedStatus`
- `selectCanonicalCheckStatusUrl` prefers relay-provided URL, falls back to local route
- `terminalReason` propagated through the full chain
- All 11 review comments addressed in follow-up commit
- Comprehensive test coverage (521-line payment-status route test)
- OpenAPI, public docs, and LLM docs all consistent with runtime behavior

### Minor note
- A dedicated test for the pending-without-paymentId fallback path (immediate delivery at 201) would strengthen confidence in the HTTP-relay compat behavior. Not a blocker.

### Verdict

Ready to merge. Great work on this one.

arc0btc (approved):
Shifts the inbox payment flow from optimistic delivery (201 + pending = already stored) to staged delivery (202 + pending = staged locally, not yet visible in inbox). Clean architectural improvement that closes the gap where callers couldn't distinguish "delivered but payment unsettled" from "payment accepted but delivery pending." 

**What works well:**
- The `@aibtc/tx-schemas` type replacement in `relay-rpc.ts` removes ~50 lines of locally-maintained type definitions that were diverging from the relay source of truth — exactly the right call
- `finalizeStagedInboxPayment` handles the idempotent repair case (message written, indexes not yet updated before crash) — important given KV's non-transactional nature
- `payment-logging.ts` structured observability with `repo_version`, `compat_shim_used`, `checkStatusUrl_present` as top-level boolean — these are the fields we actually need to slice on in production dashboards
- The `submitted → queued` collapse is implemented as a compatibility shim with explicit logging, not silently swallowed — good for diagnosing rollout gaps
- Test coverage is thorough: idempotent finalize, concurrent poll safety, malformed relay payload detection, all terminal states (confirmed/failed/replaced/not_found)

**[question] `collapseSubmittedStatus` passthrough behavior** (`lib/inbox/payment-contract.ts:28-43`)

The function handles three cases: (1) valid `TrackedPaymentStateSchema` status → pass through; (2) `submitted` → collapse to `queued`; (3) anything else → pass through silently. The third case means an unknown status like `"processing"` or `"pending_sponsor"` would pass through to `PaymentStatusHttpResponseSchema.parse()` downstream and throw a ZodError, appearing as a 500 rather than a more informative log. Is the intent that unknown statuses always surface as schema parse errors? If so, the fallback `return raw` could be replaced with an explicit `throw` or logged warn before passing through so we get structured observability before the parse fails.

**[suggestion] Double cast on env parameter** (`app/api/inbox/[address]/route.ts:1097`)

```ts
{ route: request.nextUrl.pathname, repoVersion, env: env as unknown as Record<string, unknown> }
```

`getPaymentRepoVersion` only reads `DEPLOY_SHA` and `CF_PAGES_COMMIT_SHA` — both are `string | undefined`. Since `env` has those fields typed in `cloudflare-env.d.ts`, passing just the needed fields avoids the double cast:

```suggestion
{ route: request.nextUrl.pathname, repoVersion, env: { DEPLOY_SHA: env.DEPLOY_SHA, CF_PAGES_COMMIT_SHA: env.CF_PAGES_COMMIT_SHA } }
```

**[nit] `receiptId` in 202 response** (`app/api/inbox/[address]/route.ts`, 202 branch)

The staged 202 response omits `receiptId` from the response body, but the 201 path now unconditionally includes it via `...(paymentResult.receiptId && { receiptId: paymentResult.receiptId })`. If the HTTP fallback path can return a `receiptId` on a pending result, callers on the HTTP path would never see it in the 202 body. Minor, but worth confirming the HTTP fallback always goes through the 201 path (not the 202 staged path).

**Code quality notes:**
- `selectCanonicalCheckStatusUrl(...candidates).find(c => typeof c === 'string' && c.length > 0)` — clean; correctly handles both `undefined` and empty-string candidates
- `TERMINAL_REASON_ERROR_CODE_MAP` in `relay-rpc.ts` is the right place to centralize this mapping; keeps `mapTerminalOutcome` a pure lookup

**Operational context:**
We run this payment flow in production (~100 x402 transactions/day through our welcome pipeline). The old `201 + pending` behavior was a source of confusion when relay CB tripped — callers couldn't tell if the message was "safely delivered but payment unsettled" vs "payment accepted but nothing stored yet." The staged delivery model closes that ambiguity entirely. The 24h TTL on `STAGED_PAYMENT_TTL_SECONDS` is appropriate given our ghost nonce eviction experience — worst-case 554-style stalls took ~48h to self-resolve, but 24h covers the vast majority of legitimate pending settlements.

Rollout note from our production experience: if the relay returns `effectiveCapacity=1` (our current state, escalated to whoabuddy), each staged delivery will block on a single concurrent slot — expect ~1 staged delivery at a time until capacity is raised. Not a correctness issue, just throughput.

PR #622, opened 2026-04-19
cleanup

biwasxyz:


arc0btc (approved):
Removes "on Stacks" from the Smart Contracts card description — clean, minimal copy edit.

**What works well:**
- The trimmed copy reads more concisely. "Deploy Clarity contracts" is unambiguous — Clarity is already the Stacks contract language, so the qualifier was redundant.
- No behavior, logic, or layout changes. Zero risk.

**Performance:** No issues — static string change.

**Composition:** No issues.

**UI/Accessibility:** No issues.

Good cleanup. ✓

PR #624, opened 2026-04-21
fix(cache): serve real data while agent list rebuilds

biwasxyz:
## Summary

- Homepage sporadically showed `0 agents registered` until a refresh caught a fresh cache entry.
- Root cause: `lib/cache/agent-list.ts` returned an empty snapshot (`{ agents: [], stats: { total: 0, … } }`) whenever a cache miss coincided with another request rebuilding. With a 120s hard TTL aligned to Next.js ISR revalidation, the homepage hit this path regularly.
- Fix: switch to stale-while-revalidate.
  - Hard TTL bumped 120s → 600s so the snapshot persists well past its freshness window.
  - Freshness (120s) is now derived from the in-payload `cachedAt` timestamp.
  - Stale hits return immediately and trigger a background rebuild (via `ctx.waitUntil` when provided — wired through from `app/page.tsx`).
  - Cold misses that coincide with a rebuild poll KV briefly (~1.5s) for the fresh result instead of returning zeros; fall through to a sync rebuild if the wait times out.
- The new `waitUntil` parameter is optional, so `/api/agents`, `/api/leaderboard`, `app/agents/page.tsx`, and `lib/activity.ts` keep working unchanged (they'll just block on rebuilds rather than deferring them — a follow-up can plumb `waitUntil` there too).

## Test plan

- [x] `npx tsc --noEmit` — clean
- [x] `npm run test` — 480/480 passing
- [ ] Deploy preview: hit `/` repeatedly across the 2-minute freshness boundary and confirm the agent count never drops to 0
- [ ] Concurrent load: open multiple tabs around the freshness boundary and confirm all render real data
- [ ] Mutation path: register/vouch/etc. still invalidates the cache and the next read rebuilds fresh

### DIFF
--- a/app/page.tsx
+++ b/app/page.tsx
@@ -91,9 +91,11 @@ interface LeaderboardAgent {
  */
 async function fetchHomeData() {
   try {
-    const { env } = await getCloudflareContext();
+    const { env, ctx } = await getCloudflareContext();
     const kv = env.VERIFIED_AGENTS as KVNamespace;
-    const { agents, stats } = await getCachedAgentList(kv);
+    const { agents, stats } = await getCachedAgentList(kv, (p) =>
+      ctx.waitUntil(p)
+    );
 
     // Sort: level desc, then check-ins desc, then recent first
     const sorted = [...agents].sort((a, b) => {
--- a/lib/cache/agent-list.ts
+++ b/lib/cache/agent-list.ts
@@ -63,7 +63,14 @@ export async function getCachedAgentList(
   kv: KVNamespace,
   waitUntil?: WaitUntil
 ): Promise<CachedAgentList> {
-  const cached = parseSnapshot(await kv.get(CACHE_KEY));
+  const raw = await kv.get(CACHE_KEY);
+  const cached = parseSnapshot(raw);
+
+  // Corrupted entry — delete it so we don't keep hitting the parse failure
+  // for the full 600s hard TTL.
+  if (raw && !cached) {
+    await kv.delete(CACHE_KEY).catch(() => {});
+  }
 
   if (cached) {
     if (isFresh(cached)) return cached;
--- a/lib/cache/agent-list.ts
+++ b/lib/cache/agent-list.ts
@@ -2,7 +2,11 @@
  * KV-backed cache for the enriched agent list.
  *
  * Replaces O(N) KV scans on every page load with a single KV read.
- * The snapshot is rebuilt on cache miss and stored with a 2-minute TTL.
+ * Uses a stale-while-revalidate pattern: cached data is kept for 10 minutes
+ * (hard TTL) but considered fresh for only 2 minutes. Stale hits trigger a
+ * background rebuild without blocking the response, so readers never see an
+ * empty list just because another request happens to be rebuilding.
+ *
  * Mutation endpoints call invalidateAgentListCache() to force a rebuild.
  */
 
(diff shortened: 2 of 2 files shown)


arc0btc (approved):
Fixes the homepage's intermittent "0 agents" flash by replacing the old thundering-herd guard (which returned an empty fallback) with a proper stale-while-revalidate pattern. The root cause analysis in the PR description is accurate and the fix is well-targeted.

**What works well:**
- `isFresh()` + `parseSnapshot()` are clean, focused helpers
- The stale path correctly triggers a background rebuild without blocking the response — `ctx.waitUntil` wiring in `app/page.tsx` is the right Cloudflare pattern
- Cold-miss polling (`pollForCache`) elegantly handles the "someone else is rebuilding" case without duplicating O(N) KV scans
- Bumping the hard TTL to 600s gives a comfortable stale window; the in-payload `cachedAt` freshness check is cleaner than relying purely on KV TTLs
- Optional `waitUntil` param preserves backward compat for other callers

**[suggestion] Corrupted-entry cleanup regression** (`lib/cache/agent-list.ts`)

The old code deleted a corrupted cache entry before rebuilding:
```ts
// OLD
try {
  return JSON.parse(cached) as CachedAgentList;
} catch {
  await kv.delete(CACHE_KEY).catch(() => {});
}
```
The new `parseSnapshot` returns `null` on a JSON parse failure but never deletes the bad entry. With a 600s TTL, a single corrupted write would cause every request to parse-fail → treat as cold miss → rebuild for up to 10 minutes, even though the bad key is sitting right there.

Suggested fix inside `getCachedAgentList`:
```suggestion
  const raw = await kv.get(CACHE_KEY);
  const cached = parseSnapshot(raw);
  if (raw && !cached) {
    // Corrupted entry — remove it so we don't keep hitting the parse failure
    await kv.delete(CACHE_KEY).catch(() => {});
  }
```

**[question] `maybeTriggerBackgroundRebuild` BUILDING_KEY race** (`lib/cache/agent-list.ts`)

The check-then-set on `BUILDING_KEY` (get → put) isn't atomic, so two concurrent stale-hitting requests could both read no sentinel and both kick off a background rebuild. This was true in the old code too, and duplicate rebuilds are safe (idempotent overwrite), but under burst traffic at the freshness boundary this could create unnecessary KV scan pressure. Worth noting as a follow-up if load warrants it — not blocking.

**Code quality notes:**
- `COLD_MISS_POLL_INTERVAL_MS` / `COLD_MISS_POLL_MS` constants are a nice touch — easy to tune in one place.
- The `finally { kv.delete(BUILDING_KEY) }` cleanup is now duplicated between `getCachedAgentList` (sync rebuild path) and `maybeTriggerBackgroundRebuild` (async path). Both are correct; just noting the duplication in case a future consolidation becomes obvious.

**Performance:**
- No new waterfalls; the stale path is pure non-blocking. Background rebuild via `waitUntil` is the right Cloudflare idiom.
- Polling up to ~10 × 150ms KV reads on cold misses is acceptable — this path only hits when the cache is completely empty and a concurrent rebuild is in flight (cold start or post-invalidate burst).

**UI/Accessibility:** `app/page.tsx` change is plumbing only — no UI impact.

Operational note: We've seen the "0 agents" flash reported intermittently from the field. The old thundering-herd guard was the right idea but the empty-fallback return was the footgun. This fix is the correct layered approach.

biwasxyz:
Thanks @arc0btc — addressed the corrupted-entry regression in 019a89d. Restored the proactive `kv.delete(CACHE_KEY)` when the raw read fails to parse, so a single bad write can't poison reads for the full 600s hard TTL.

On the `BUILDING_KEY` race: agreed it's not worth fixing in this PR — the old code had the same non-atomic check-then-set, duplicate rebuilds are idempotent, and we haven't seen the scan pressure in practice. Happy to pick it up as a follow-up (e.g., atomic claim via KV metadata or a Durable Object) if load ever warrants.

Issue #614, opened 2026-04-17
Umbrella: STX address migration for pre-#563 wrong-address registrations


whoabuddy:
# Umbrella: STX address migration for pre-#563 wrong-address registrations

Consolidated tracking for agents who registered before #563 with an STX address
they don't control, but still control their BTC key. The platform rejects these
registrations at the door now; this umbrella handles the backlog.

## Source of truth

- **Migration batch:** `scripts/migrations/stx-batch-01.json` (committed to
  aibtcdev/landing-page after signature collection closes)
- **Tooling:** `scripts/migrate-stx-address.ts` — verifies each entry's BTC +
  NEW-STX signatures, rewrites `stx:*` / `btc:*` KV, invalidates the agent-list
  cache. Run locally against prod KV via Cloudflare API; no deployed admin
  endpoint, so no lingering attack surface.
- **Challenge format** (both signatures required):
  `AIBTC STX Migration | btc={btc} | old={oldStx} | new={newStx} | date={iso}`
  - `btcSignature` — authorizes the change (BIP-137 or BIP-322)
  - `stxSignature` — from the NEW STX key, proves destination control

## Verification window

- **Published:** 2026-04-23
- **Deadline:** **2026-04-26T23:59:59Z**
- **Response policy:** reply to your child issue only if something is wrong.
  Silence = acknowledged, migration proceeds.
- **Dispute path:** a reply on a child issue before the deadline pauses that
  agent's migration until resolved; everyone else's proceeds.

## Affected agents

| Agent | BTC (correct) | Registered STX (wrong) | Correct STX |
|-------|---------------|------------------------|-------------|
| Broad Turtle | `bc1q3wcjxn2wqk2sl2jv8vtnvhcnjkx8uare82296x` | `SP2QRR3M0RBV4GG4VQE36T11ZWRN4RBQD1QK5ZAMB` | *pending* |
| Lightning Cache | `bc1qkefj5auvv28gtw03uhuag45729az7jczdlh7tx` | `SP1N4AX7RB6HMC80TVQMGGX2P2SH6ZAQVW5C94612` | `SP3AJCB6VVE0DG5HAMT0YGHQYW4Z3BDCQA6ZJ6VY3` |
[...]

PR #627, opened 2026-04-21
chore(scripts): add STX address migration tooling and umbrella drafts

whoabuddy:
## Summary

- Adds `scripts/migrate-stx-address.ts`, a one-shot operator tool for the pre-#563 wrong-STX backlog. Verifies BTC + NEW-STX signatures over a bound challenge, rewrites `stx:` / `btc:` KV via the Cloudflare REST API, invalidates the agent-list cache. No deployed admin endpoint → no lingering attack surface.
- Stages issue-body drafts under `scripts/migrations/` so the umbrella, close-outs, and paid-attention retirement follow-up have a committed audit trail.
- Refs #614.

## What's in `scripts/`

| File | Purpose |
|---|---|
| `migrate-stx-address.ts` | Migration script — run via `npx tsx scripts/migrate-stx-address.ts <batch.json> --dry-run\|--apply` |
| `migrations/EXAMPLE.json` | Input format template (Void Parrot as worked example) |
| `migrations/umbrella-614.md` | Revised body for #614 (STX-migration scope only) |
| `migrations/closeout-comment.md` | Shared close-out template for the 3 child issues |
| `migrations/paid-attention-retire.md` | New-issue body for the `/api/paid-attention` retirement, framed as its evolution into heartbeat (liveness) + x402 inbox (paid attention) |

## Design notes

- **Why not a persistent admin endpoint** — any `X-Admin-Key`-gated STX rewrite endpoint is a high-value target: if the key leaks, an attacker can redirect inbox payments (dynamic payTo), vouch history, and future achievements for any agent. A one-shot script run from an operator workstation keeps the blast radius bounded and produces a git-committed audit trail (the input JSON file).
- **Why not a self-service challenge action** — already ruled out in #614: self-service STX rotation makes any future BTC-key compromise a full account takeover.
- **Why not have agents re-register** — would lose achievements, check-in counts, Genesis claim status, inbox messages, vouch history, referral codes. Punishes agents for our bug.
- **Dual signatures required** — BTC signature authorizes the change; STX signature from the NEW address proves destination control. The old STX key is never required (the agent doesn't control it — that's the whole point).
- **Mutation order** — write new `stx:` first, update `btc:`, delete old `stx:` last. Partial failures stay re-runnable.
- **Historical data preserved** — `InboxMessage.toStxAddress` is not rewritten; sender receipts and on-chain txids stay true.

## Test plan

- [x] `npm run typecheck` passes
- [x] `npm run lint` passes (only pre-existing `next/image` warnings)
- [x] `npm test` passes (480 tests)
- [ ] Collect signatures from the 5 affected agents via #614
- [ ] Commit populated `scripts/migrations/stx-batch-01.json`
- [ ] `--dry-run` validates all entries
- [ ] `--apply` against prod KV
- [ ] Each affected agent hits `POST /api/identity/{address}/refresh` on the new STX

## Follow-ups (separate PRs / issues)

- Post revised #614 body and close-out comments from `scripts/migrations/` (`gh issue edit`, `gh issue comment`)
- File the `/api/paid-attention` retirement issue using `scripts/migrations/paid-attention-retire.md`
- Ship a 410 Gone shim at `app/api/paid-attention/route.ts` pointing deployed agents at heartbeat + inbox
- Coordinate downstream doc updates in aibtc-mcp-server, openclaw-aibtc, and skills
- Address aibtc-mcp-server#453 separately (Genesis payout machinery gap — real, not a hallucination; the claim endpoint rolls a real 5k–10k sat `rewardSatoshis` and returns "will be sent shortly," but verified→rewarded requires a manual admin payout)

### DIFF
--- a/scripts/migrate-stx-address.ts
+++ b/scripts/migrate-stx-address.ts
@@ -1,0 +1,244 @@
+#!/usr/bin/env tsx
+/**
+ * Operator-assisted STX address migration.
+ *
+ * Fixes agents who registered with a wrong STX address before #563 (when
+ * signature-based address verification landed). Each agent still controls
+ * their BTC key but registered an STX address they do not control.
+ *
+ * This script is intentionally a one-shot, run-locally tool rather than a
+ * deployed admin endpoint. Persistent admin endpoints that can rewrite
+ * `stx:{address}` pointers are a high-value target: if the admin key ever
+ * leaks, an attacker can redirect inbox payments (dynamic payTo), vouch
+ * history, and future achievements. Running the migration from an operator
+ * workstation against the KV REST API keeps the blast radius bounded and
+ * produces a git-committed audit trail (the input JSON file).
+ *
+ * The migration file format is a JSON array of MigrationEntry records. After
+ * running --apply successfully, commit the input file under
+ * scripts/migrations/ so the audit trail lives in the repo.
+ *
+ * Challenge message (signed by both keys):
+ *   AIBTC STX Migration | btc={btc} | old={oldStx} | new={newStx} | date={iso}
+ *
+ * Required signatures per entry:
+ *   - btcSignature — BIP-137 or BIP-322 signature from the agent's BTC key.
+ *     Authorizes the change.
+ *   - stxSignature — SIP-018/RSV signature from the NEW STX key. Proves the
+ *     agent controls the destination address. The old STX key is never
+ *     required; the agent does not control it.
+ *
+ * Usage:
+ *   export CLOUDFLARE_ACCOUNT_ID=...
+ *   export CLOUDFLARE_API_TOKEN=...    # needs Workers KV Storage:Edit scope
+ *   npx tsx scripts/migrate-stx-address.ts scripts/migrations/stx-batch-01.json --dry-run
+ *   npx tsx scripts/migrate-stx-address.ts scripts/migrations/stx-batch-01.json --apply
+ */
+
+import { readFileSync } from "node:fs";
+import { verifyBitcoinSignature } from "../lib/bitcoin-verify";
+import { hashMessage, verifyMessageSignatureRsv } from "@stacks/encryption";
+import { bytesToHex } from "@stacks/common";
+import {
+  publicKeyFromSignatureRsv,
+  getAddressFromPublicKey,
+} from "@stacks/transactions";
+import type { AgentRecord } from "../lib/types";
+
+const KV_NAMESPACE_ID = "f8aab2734e154953a50cabdb87083af3";
+const AGENT_LIST_CACHE_KEY = "cache:agent-list";
+
+interface MigrationEntry {
+  btcAddress: string;
+  oldStxAddress: string;
+  newStxAddress: string;
+  messageIso: string;
+  btcSignature: string;
+  stxSignature: string;
+  notes: string;
+}
+
+function challengeMessage(e: MigrationEntry): string {
+  return `AIBTC STX Migration | btc=${e.btcAddress} | old=${e.oldStxAddress} | new=${e.newStxAddress} | date=${e.messageIso}`;
+}
+
+async function kvFetch(
+  key: string,
+  method: "GET" | "PUT" | "DELETE",
+  body?: string,
+): Promise<string | null> {
+  const { CLOUDFLARE_ACCOUNT_ID, CLOUDFLARE_API_TOKEN } = process.env;
+  if (!CLOUDFLARE_ACCOUNT_ID || !CLOUDFLARE_API_TOKEN) {
+    throw new Error(
+      "Missing CLOUDFLARE_ACCOUNT_ID or CLOUDFLARE_API_TOKEN env var",
+    );
+  }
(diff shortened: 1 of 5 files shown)


arc0btc (approved):
Operator-assisted STX migration tooling — solid approach. The dual-sig design (BTC key authorizes, new STX key proves destination control) is exactly right, and keeping this as a local operator script rather than a deployed admin endpoint eliminates the lingering attack surface. The write-new-before-delete-old ordering and per-entry error isolation are good.

**What works well:**
- Challenge message includes `old=` — prevents replay attacks where old signatures could be reused for a different target
- Cache invalidation scoped correctly: only fires when `apply && succeeded > 0`
- `stxPublicKey` written into the updated record — future callers don't need to re-derive it
- EXAMPLE.json + umbrella-614.md + closeout-comment.md give a clean audit trail and reduce manual coordination overhead
- `paid-attention-retire.md` is a well-structured plan: the 410 shim → mcp-server → openclaw → skills ordering is the right execution sequence

**[suggestion] Idempotency gap after partial failure** (`scripts/migrate-stx-address.ts:132-136`)

The header promises re-runnability on partial failure, but the collision check fires on the new STX address regardless of *who* owns it. If the script writes `stx:{newStxAddress}` then crashes before completing `btc:` or `DELETE`, a re-run will throw `"would collide with another agent"` — which is wrong, since it's our own partial write.

Fix: check whether the existing KV value belongs to this same `btcAddress` before treating it as a collision:

```suggestion
  const collision = await kvFetch(`stx:${entry.newStxAddress}`, "GET");
  if (collision) {
    const existing = JSON.parse(collision) as AgentRecord;
    if (existing.btcAddress !== entry.btcAddress) {
      throw new Error(
        `stx:${entry.newStxAddress} already exists and belongs to a different agent (${existing.btcAddress}). Abort.`,
      );
    }
    // same agent's partial previous run — safe to continue from where we left off
    console.log("  OK stx: already written (resuming partial run)");
  }
```

**[suggestion] `KV_NAMESPACE_ID` as env var** (`scripts/migrate-stx-address.ts:50`)

Hardcoded namespace ID makes staging dry-runs impossible without editing the file. A one-liner env var fallback:

```suggestion
const KV_NAMESPACE_ID = process.env.CLOUDFLARE_KV_NAMESPACE_ID ?? "f8aab2734e154953a50cabdb87083af3";
```

**[nit]** `verifyStxSignature` recovers the address and public key in addition to verifying — the name understates what it does. `recoverAndVerifyStxSignature` or `resolveStxSignature` would make the return shape (`.address`, `.publicKey`) less surprising to a reader.

**Code quality notes:**

The `as AgentRecord` cast on `JSON.parse(rawBtc)` is fine here since this is operator tooling reading our own KV — runtime validation would be overkill for a one-shot script. No simplifier concerns otherwise; the abstraction level matches the task.

**Operational note:** We run the Cloudflare KV stack for the agent registry daily and have hit partial-write scenarios under network flakiness. The re-run path matters more than it might look on first read — the idempotency fix is worth doing before the 2026-04-23 publish date since the batch will have 5+ entries and it's better to be safe.

netmask255:
## Thin Teal — Migration Entry (CORRECTED)

**BTC Address:** bc1qlgcphpkq3yc38ztr6n48qh3ltsmxjprv9dm0ru

**Registered STX (wrong):** SP1GQY3562X3EJ35N5SP5Q7CG981KMNFJN3SWXZ88

**❌ WRONG STX in Umbrella Issue #614:** 
~~SP3K3NCZ48H4VX4564AQ53FQECVSAOJ8R73YKH9ZY~~
(O character mismatch — was incorrectly listed)

**✅ CORRECT STX Address:** SP3K3NCZ48H4VX4564AQ53FQECVSA0J8R73YKH9ZY
(Derived from verified mnemonic — this is the correct address)

**Challenge Message:**
AIBTC STX Migration | btc=bc1qlgcphpkq3yc38ztr6n48qh3ltsmxjprv9dm0ru | old=SP1GQY3562X3EJ35N5SP5Q7CG981KMNFJN3SWXZ88 | new=SP3K3NCZ48H4VX4564AQ53FQECVSA0J8R73YKH9ZY | date=2026-04-22T13:36:47Z

**BTC Signature (BIP-322):**
AkgwRQIhAP2RdDPFV5k+bUwFO0PfaSzKSZO9w2GjQjdK6drmhMZ3AiBTqv33e3+Wpx9FpMQMiunyIhPSqTCgTZxJXKAjLme2eAEhAyp1PNF9NR7WmJH1xY7hDSCNeAcIP4VMW8AEJikpx2ev

**STX Signature (RSV):**
1d4ea827fd75b2d96334cdfad58fe6f845c666d20782f545ad874d396c4253250243f4084b27c569809d4d93ca8bcc8cbb6cb0a33b08e76fc0aaf38cbb98835900

---

**Note:** I apologize for the error. The STX address in Umbrella Issue #614 incorrectly shows `SAOJ8...` — it should be `SA0J8...`. Both signatures above use the correct address **SP3K3NCZ48H4VX4564AQ53FQECVSA0J8R73YKH9ZY** derived from my verified mnemonic.

PR #638, opened 2026-04-29
redesign: port full-site to Operator Console (Version A)

biwasxyz:
Ports the [Claude Design](https://claude.ai/design) **Version A — Operator Console** handoff to production. Visual layer only — every page keeps the existing live KV / KV-cached fetches and real API response shapes.

## Status

- [x] Tokens + base CSS (Roc Grotesk preserved, system mono stack — no extra fonts)
- [x] Shared chrome: Navbar + redesign primitives (ToastRoot, CommandPill, Eyebrow, LevelChip, Avatar, BgLayers, Seg)
- [x] Homepage (Hero with terminal animation, 5-step Operators, NetworkGraph, Capabilities, ActivityFeed)
- [x] `/agents` directory + AgentList
- [ ] `/agents/[address]` profile
- [ ] `/inbox/[address]` (mail-client UI)
- [ ] `/activity`, `/bounty`, `/skills`, `/guide`, `/identity`, `/status`, `/heartbeat`
- [ ] Build + dev smoke test

## Schema notes

Probed live aibtc.com endpoints to bind the new UI to real fields (the CLAUDE.md docs are stale in places):

- `/api/agents` returns `{ agents[], pagination }` — **no top-level `stats`**. Use `/api/leaderboard` → `.distribution` for stats (`genesis`, `registered`, `activeAgents`, `totalCheckIns`).
- Per-agent `messageCount` is not on records; the detail endpoint exposes `.activity.unreadInboxCount` + `.activity.sentCount`.
- `/api/activity` events are a discriminated union on `type` (`message | achievement | registration`).
- `/api/inbox/{addr}` wraps everything in `inbox.messages[]` + exposes an `economics` block + `howToSend`.
- `/api/heartbeat?address=` returns `{ orientation: { ..., nextAction: { step, description, endpoint } }, documentation }`.

## Side fixes shipped along the way

- `next.config.ts`: force `globalThis.AsyncLocalStorage` so `initOpenNextCloudflareForDev()` actually runs in dev (gate was skipping the call → every API hitting KV was 500'ing locally with "called without having called initOpenNextCloudflareForDev").
- `ActivityFeed.tsx`: outer `<Link>` produced nested `<a>` with the row Links inside CompactEventRow → swapped to a div + explicit "View all" link in the header. Also null-guarded `queueRef.current` for React strict-mode double-init.

## Commits land per-file

Each file is committed individually so review can move sequentially.

## Test plan

- [ ] `npm run dev` + manually visit `/`, `/agents`, `/agents/<address>`, `/activity`, `/bounty`, `/skills`, `/guide`, `/identity`, `/status`, `/heartbeat`, `/inbox/<address>`
- [ ] `npm run lint`
- [ ] `npm run build` (or typecheck)
- [ ] Confirm SendMessageModal still posts to `/api/inbox/{addr}` with the redesigned chrome around it


arc0btc (changes requested):
Ports the Operator Console visual language to homepage and agents — solid scope discipline keeping this visual-layer only, and the API schema notes in the PR description are genuinely useful for future contributors.

**What works well:**
- Thorough aria coverage throughout: icon buttons labeled, nav landmarks, decorative elements aria-hidden, the NetworkGraph SVG has role/aria-label
- Event listener cleanup is consistent — every useEffect with addEventListener has a matching cleanup return
- The `role="button" + tabIndex={0} + onKeyDown` pattern in Capabilities FAQ items is correct (handles Enter/Space)
- `label` wrapping the search input avoids a separate htmlFor association
- The null guard on `queueRef.current` and nested `<a>` fix in ActivityFeed are good defensive touches

---

## Performance

**[suggestion] NetworkGraph should use `next/dynamic`** (`app/page.tsx`)  
NetworkGraph is a 292-line SVG animation component with `requestAnimationFrame`-style setInterval loops. It isn't needed for initial text rendering. Loading it synchronously adds to the page bundle and delays TTI for the hero copy above it.
```suggestion
const NetworkGraph = dynamic(() => import('./components/home/NetworkGraph'), { ssr: false });
```
Add `import dynamic from 'next/dynamic'` at the top.

**[suggestion] No `prefers-reduced-motion` guard for `fadeUp` / `glowPulse`** (`app/globals.css`)  
The Hero, Capabilities, NetworkGraph, and AgentList all use `animate-fadeUp` and `animate-glowPulse`. The globals.css diff doesn't include a `@media (prefers-reduced-motion: reduce)` reset for these keyframes. Users who have disabled motion in their OS get the full animation sequence regardless. Wrap or reset:
```css
@media (prefers-reduced-motion: reduce) {
  .animate-fadeUp,
  .animate-glowPulse {
    animation: none;
    opacity: 1;
  }
}
```

---

## UI/Accessibility

**app/agents/AgentList.tsx — search input `outline-none` without focus replacement** [BLOCKING]  
The old code had `focus:border-white/20 focus:bg-white/[0.04]` as a visible focus replacement. The new code is just:
```tsx
className="flex-1 bg-transparent text-[13px] outline-none"
```
No `focus-visible:ring-*`, no border change, no background shift — keyboard users get no indicator when the search field is focused. This is a regression from the previous implementation.
```suggestion
className="flex-1 bg-transparent text-[13px] outline-none focus-visible:ring-1 focus-visible:ring-white/30 rounded-sm"
```

---

## Code quality notes

**[suggestion] `window.__showToast` global bridge** (`app/components/redesign.tsx`)  
Mutating `window` is fragile — TypeScript has to declare it separately, it can't be tree-shaken, and it breaks if two components import the module before the Toast mounts. React 19's `use()` or a small context would be cleaner and keep the bridge typed end-to-end. Not blocking for this PR since it's encapsulated, but worth tracking before the inbox/activity pages use it more heavily.

**[nit]** Capabilities FAQ items use `role="button" + tabIndex={0} + onKeyDown` — this works correctly, but a `<button>` element would handle Enter/Space natively and inherit `cursor: pointer` without extra CSS. Saves a few lines and is more robust across AT.

---

*Operational note:* We run the live aibtc.com API endpoints in our sensors — the schema corrections in the PR description (no top-level `stats` on `/api/agents`, `messageCount` on detail not list, discriminated union on `/api/activity`) match what our sensors observe. Good ground-truth validation.

biwasxyz:
Thanks for the thorough review @arc0btc — addressed all of it in the latest 4 commits:

- **[BLOCKING] Search focus indicator** (`d08b856`) — added `focus-visible:ring-1 focus-visible:ring-white/30 rounded-sm` to the agents search input. Also swept the same regression on the inbox search and skills search (`527d7a0`).
- **prefers-reduced-motion guard** (`da2f2a9`) — agree the universal `animation-duration: 0.01ms` rule wasn't enough because it doesn't cancel `animation-delay`, leaving elements invisible during the delay window before snapping to end state. Added explicit `animation: none; opacity: 1; transform: none` overrides for `.animate-fadeUp`, `.animate-glowPulse`, `.animate-feedSettle`, `.animate-slideIn`, `.animate-fadeIn`, `.feed-item-enter`, and the `.typing::after` blinking cursor.
- **NetworkGraph dynamic import** (`d855e53`) — `app/page.tsx` is a server component, so I added a thin `"use client"` wrapper (`NetworkGraphLazy.tsx`) that does `next/dynamic` with `ssr: false`. Loading state is a placeholder div with the same border + radial background as the graph and `aspect-ratio: 1280 / 600`, so the section reserves the right vertical space and content below doesn't shift when the chunk hydrates.
- **`role="button"` in Capabilities** (nit) — kept as-is. Each card has a nested `<button>` (Copy prompt) when expanded, and HTML doesn't allow nested buttons. The `role="button" + tabIndex={0} + onKeyDown` pattern handles Enter/Space correctly and `aria-expanded` is already wired.
- **`window.__showToast` global bridge** — agreed, will track for a follow-up PR. Encapsulated for now and only one component (`ToastRoot`) writes to it; consumers go through the `showToast(msg)` typed export.

Glad the schema notes matched your sensors' observations.

arc0btc (approved):
Blocking issue fixed (search input focus styling), both suggestions implemented (NetworkGraph dynamic import, prefers-reduced-motion guard), and keyboard handling on role="button" divs is correct throughout. Code quality is strong — no TypeScript escape hatches, proper event listener cleanup, and aria coverage solid.

PR #639, opened 2026-04-29
fix(stacks-api): warn on monthly quota only, drop per-window noise

whoabuddy:
## Summary

- Replace noisy per-second rate-limit warn with a meaningful monthly-quota warn
- `extractRateLimitInfo` now reads `x-ratelimit-remaining-stacks-month` and `-limit-stacks-month` and emits `stacksApi.approaching_monthly_quota` when remaining drops below 20% of the monthly limit
- Drop the per-request `stacksApi.rate_limit_remaining` info event and the per-second `approaching_rate_limit` warn — neither was actionable since the retry wrapper already absorbs short-window 429s

## Why

Today's production logs (aibtc-landing) show `stacksApi.approaching_rate_limit` firing ~489x with `rlRemaining: 39, rlLimit: 40` — the threshold of 50 was set against per-second windows that cap at 20-40, so the warn was triggering on essentially every request and conveying no real signal. Meanwhile the monthly Hiro quota — the thing that actually requires action (plan upgrade or traffic shed) — wasn't being tracked at all.

This change replaces the noise with a single threshold-gated alert that only fires when we have <20% of the monthly budget left. Percentage-based so it adapts to whichever plan tier the API key is on.

## Net effect on log volume
- ~22k `stacksApi.rate_limit_remaining` info events/day → 0
- ~489 `stacksApi.approaching_rate_limit` warns/day → 0 (until we're actually near the cap)

## Test plan

- [x] `npx vitest run lib/__tests__/stacks-api-fetch.test.ts` — 7 tests pass
- [x] `npx vitest run` — full suite (518 tests) passes
- [x] `npx tsc --noEmit` — typecheck clean
- [ ] After deploy: verify `stacksApi.approaching_rate_limit` warns stop appearing and `stacksApi.approaching_monthly_quota` does NOT fire under normal traffic

### DIFF
--- a/lib/__tests__/stacks-api-fetch.test.ts
+++ b/lib/__tests__/stacks-api-fetch.test.ts
@@ -37,45 +37,15 @@ describe("stacksApiFetch logger telemetry", () => {
     mockFetch.mockClear();
   });
 
-  it("emits stacksApi.rate_limit_remaining on every parseable response", async () => {
-    const logger = createMockLogger();
-    mockFetch.mockResolvedValue({
-      ok: true,
-      status: 200,
-      url: "https://api.mainnet.hiro.so/v2/contracts/call-read/foo/bar/baz",
-      headers: mockHeaders({
-        "ratelimit-remaining": "123",
-        "ratelimit-limit": "500",
-        "ratelimit-reset": "42",
-      }),
-    });
-
-    await stacksApiFetch(
-      "https://api.mainnet.hiro.so/v2/contracts/call-read/foo/bar/baz",
-      {},
-      { logger }
-    );
-
-    const rlEvents = logger._events.filter(
-      (e) => e.message === "stacksApi.rate_limit_remaining"
-    );
-    expect(rlEvents).toHaveLength(1);
-    expect(rlEvents[0].context).toMatchObject({
-      path: "/v2/contracts/call-read/foo/bar/baz",
-      rlRemaining: 123,
-      rlLimit: 500,
-    });
-  });
-
-  it("emits stacksApi.approaching_rate_limit when remaining < 50", async () => {
+  it("emits stacksApi.approaching_monthly_quota when monthly remaining < 20% of limit", async () => {
     const logger = createMockLogger();
     mockFetch.mockResolvedValue({
       ok: true,
       status: 200,
       url: "https://api.mainnet.hiro.so/extended/v1/tx/0xabc",
       headers: mockHeaders({
-        "ratelimit-remaining": "10",
-        "ratelimit-limit": "500",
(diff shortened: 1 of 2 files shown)


arc0btc (approved):
Replaces miscalibrated per-second rate-limit noise (489 warns/day, threshold of 50 vs window cap of 20-40) with a single meaningful monthly-quota alert. Good production data to back the change.

**What works well:**
- The root cause is well-diagnosed with real numbers. 489 warns/day on essentially every request is exactly the kind of noise that trains operators to ignore alerts — removing it is the right call.
- `limitMonth > 0` guard prevents div-by-zero. Null checks on both headers before the division are clean.
- The 4 tests cover the important paths: fires when <20%, silent when ≥20%, silent when headers absent, return value parsing without logger.
- `extractPath` is now only evaluated inside the warn branch — a nice implicit lazy evaluation improvement.

**[suggestion] The "fires on every subsequent call" behavior isn't tested**

The comment in the docstring honestly notes this fires on every request until the month rolls over and recommends dedup at the alerting layer. Worth a test to lock in that expectation explicitly — future readers will thank you:

```suggestion
  it("emits approaching_monthly_quota on every call when below threshold", async () => {
    const logger = createMockLogger();
    const headers = mockHeaders({
      "x-ratelimit-remaining-stacks-month": "20000",
      "x-ratelimit-limit-stacks-month": "150000",
    });
    const mockResponse = { ok: true, status: 200, url: "https://api.mainnet.hiro.so/extended/v1/tx/0xabc", headers };
    mockFetch.mockResolvedValue(mockResponse);

    await stacksApiFetch("https://api.mainnet.hiro.so/extended/v1/tx/0xabc", {}, { logger });
    await stacksApiFetch("https://api.mainnet.hiro.so/extended/v1/tx/0xabc", {}, { logger });

    const events = logger._events.filter((e) => e.message === "stacksApi.approaching_monthly_quota");
    expect(events).toHaveLength(2); // fires every call — dedupe at alerting layer
  });
```

Not blocking — the behavior is already documented in the docstring. This just makes the contract machine-checkable.

**[nit]** `RateLimitInfo` is a public interface and its shape changed meaningfully. If any other code in the monorepo imports it (dashboards, CLI tools, other libs), they'd get silent `undefined` on the old fields. Worth a quick `grep -r 'RateLimitInfo\|rl\.remaining\|rl\.limit' --include='*.ts'` before merge to confirm no external consumers are affected.

*Operational context:* We use `stacksApiFetch` in our own sensors daily. The per-second warn was noise we'd been ignoring too — the monthly quota is the signal that actually maps to an actionable response (plan upgrade or traffic shed). This is the right fix.

PR #640, opened 2026-04-30
fix(register): downgrade expected BIP-322 btcPublicKey-unavailable noise to info (closes #579)

### LINKED ISSUE #579: Downgrade expected BIP-322 btcPublicKey unavailable registration noise
`/api/register` is generating a large volume of warnings for expected BIP-322 behavior. This should not be a warn-level operational signal.

Current issue:
In the latest 24-hour landing warning window, `65/170` warns came from `/api/register` `btcPublicKey unavailable ... BIP-322 signatures do not expose the public key via recovery`.

Relevant code:
- `aibtcdev/landing-page/app/api/register/route.ts`
- `aibtcdev/landing-page/lib/bitcoin-verify.ts`
- `aibtcdev/landing-page/lib/challenge.ts`

What this warning actually means:
- The agent used a BIP-322 signing path, typically `bc1q` or `bc1p`.
- Signature verification succeeded.
- The platform could verify address ownership, but could not recover the BTC public key from the signature.
- That means automatic BTC-pubkey-based Nostr derivation is unavailable.
- Registration is still valid.

The remediation path already exists:
- Provide `nostrPublicKey` at registration, or
- Use `update-pubkey` later via the challenge flow.

What should change:
- Downgrade this log from `warn` to `info`.
- Keep a structured field like:
  - `btcPublicKeyMissing: true`
  - `reason: "bip322_no_pubkey_recovery"`
  - `nextAction: "update-pubkey"`
- Optionally suppress repeated logs per BTC address.
- Preserve the existing response hints for agents so the API contract remains self-documenting.

Acceptance criteria:
- Expected BIP-322 registrations no longer pollute warn-level dashboards.
- Agents still get clear remediation in the API response.
- Operators can still count the path via structured fields if needed.

whoabuddy:
## Summary

Closes #579. `/api/register` was emitting a `log.warn` every time a BIP-322 signer (bc1q/bc1p native SegWit) registered — because BIP-322 signatures don't expose the public key via signature recovery and we can't auto-derive a Nostr npub from btcPublicKey. This is **expected, documented behavior**; registration completes successfully, and the remediation path is already surfaced in the GET self-doc + registration response.

Per the issue, **65 of 170 daily warns** came from this one site (~38% of the warn stream).

## Change

`app/api/register/route.ts:723` — `log.warn` → `log.info`. Comment updated to reflect the new intent (correlation handle, not incident signal). The remediation pointer stays in the GET self-doc (line 219+) and in the response (`btcPublicKeyMissing: true` flag).

## Test plan

- [x] `npx tsc --noEmit` clean
- [ ] CI green
- Operationally: tail the warn-level register logs after merge to confirm the BIP-322 noise dropped out and only real warnings remain.

Co-Authored-By: Claude Opus 4.7 (1M context) <noreply@anthropic.com>

### DIFF
--- a/app/api/register/route.ts
+++ b/app/api/register/route.ts
@@ -717,11 +717,13 @@ export async function POST(request: NextRequest) {
       : createConsoleLogger({ rayId, path: "/api/register" });
 
     // BIP-322 wallets (bc1q/bc1p) don't expose the public key via signature recovery.
-    // Log a warning so operators can track agents with missing btcPublicKey.
-    // Registration is NOT blocked — BIP-322 agents register successfully with an empty publicKey.
+    // This is expected behavior, not an incident — registration completes
+    // normally with an empty publicKey. The remediation path (provide
+    // nostrPublicKey now, or update-pubkey via challenge later) is documented
+    // in the GET self-doc and surfaced in the response (#579).
     const btcPublicKeyMissing = !btcResult.publicKey;
     if (btcPublicKeyMissing) {
-      log.warn(
+      log.info(
         `btcPublicKey unavailable for address ${btcResult.address}: ` +
         "BIP-322 signatures do not expose the public key via recovery. " +
         "Nostr npub derivation from btcPublicKey will not work for this agent."


arc0btc (approved):
Downgrading BIP-322 pubkey-unavailable from warn to info — solid cleanup.

**What works well:**
- The diagnosis is correct: BIP-322 signatures don't expose the public key via recovery (protocol property, not an error). Emitting a `warn` here was teaching operators to ignore 38% of their warn stream, which degrades the signal-to-noise ratio for real incidents.
- The comment update is an improvement — "correlation handle, not incident signal" accurately describes the new intent, and the remediation pointer (GET self-doc + `btcPublicKeyMissing: true` in the response) gives operators the right place to look.
- The log message body is preserved verbatim at `info` level, so it's still grep-able for diagnostics without paging anyone.

**Performance:** No applicable React/bundle concerns — server route handler only.

**Composition:** No component changes.

**UI/Accessibility:** No UI changes.

**Code quality notes:** The change is minimal and focused — exactly right for a log-level fix. No simplification opportunities.

**Operational note:** We process BIP-322 registrations through this endpoint regularly. The warn-level noise made it harder to surface real registration anomalies. This aligns with what was done in #639 for stacks-api quota warnings. After merge, tailing `warn`-level register logs should show only genuine incidents.

PR #644, opened 2026-05-03
fix: reduce landing-page log noise + cache transient achievement-verify timeouts

whoabuddy:
## Summary

Three commits, one PR, closing B3 in `cloudflare-bill-reduction-tracker-2026-05.md` — the landing-page side of the immediate billing-leak campaign.

1. **Sample `cache.hit`/`cache.miss` INFO at 5%** via a deterministic FNV-1a hash on `(category, key)`. Cache observability is preserved as a sanity tool; the audit-log fidelity it never had isn't sacrificed. WARN/ERROR/auth/payment-terminal events stay at 100% — those don't go through `samplingFor`.
2. **Cache transient `TimeoutError`s** in the four `verify*Achievement` functions. On `mempool.space` / `api.hiro.so` timeout, write a 60s sentinel into a separate `cache:verify-timeout:*` namespace so retries within the rate-limit window short-circuit without re-fetching. The sentinel is distinguishable from real cached data by namespace, so `getCachedTransaction` / `getCachedStacking` semantics are unchanged.
3. **Downgrade the mislabelled "expected" nonce-rejection WARN** to INFO (the message literally says "(expected)"), and add `docs/cloudflare-cost-runbook.md` with the B3 entry per the `agent-news` convention.

## Why

Inventory: `worker-logs/.planning/2026-05-03T1600Z-landing-page-noise-inventory.md`

Strict `app_id = aibtc-landing` 24h sample shows `cache.hit`/`cache.miss` are essentially the entire INFO surface (~22K events/day, ~95%+ of `aibtc-landing` INFO). All four ERROR samples in the same window are `achievement.sender.verify_error → TimeoutError`, with the 5-min rate limiter measuring from the *successful* set so a timed-out call lets the next request inside the window re-fetch and re-time-out — same agent, same upstream, every retry.

The immediate-billing campaign has already closed B1 (`agent-news#725`, `NEWS_KV` writes -99.7%) and B2 (`agent-news#731`, NewsDO rows-read -80%). B3 is the last bullet.

## Expected metric movement

| Metric | Pre-PR (24h) | Target |
| --- | ---: | ---: |
| `aibtc-landing` worker-logs INFO/day | ~22,000 | ~1,000–2,000 (-90%+) |
| `aibtc-landing` worker-logs ERROR/day | ~7–19 | ~1–3 (-70%+ via timeout cache) |
| Cache hit-rate observability | 100% | 5% (deterministic — same key always wins/loses) |
| Achievement-verify correctness | 100% | 100% (sentinel is in a distinct namespace; data caches unchanged) |

Validation window: 24h post-deploy via `https://logs.aibtc.com/stats/aibtc-landing` and the same source's level/message breakdown.

## Rollback signal

- **Cache observability needed for an incident:** flip `SAMPLE_RATES["cache.event"]` to `1` in `lib/logging.ts` (no redeploy of unrelated code) for the duration. Sampling stays deterministic so the same key always wins/loses, which makes a 5%-sampled hit-rate calculation valid.
- **Achievement-verify regression:** the sentinel is in `cache:verify-timeout:*`. If `verify*Achievement` starts returning `false` for clearly-eligible agents that should be passing, query the namespace; if entries are present and stale, drop them and revert commit 2.
- **Cost signal absent at 24h:** if `aibtc-landing` INFO/day hasn't dropped materially after 24h of post-deploy traffic, investigate whether a non-`logCacheEvent` log is dominating that we missed in inventory.

## Test plan

- [x] `npm run typecheck` clean for all touched files. (Pre-existing TS errors on `main` in `lib/__tests__/bitcoin-verify.test.ts` and `lib/inbox/__tests__/reconciliation-queue.test.ts` are unrelated and reproduce on bare `main`.)
- [x] Focused suite (`sampling`, `verify-timeout-cache`, `x402-verify`): **22/22 pass**.
  - `lib/__tests__/sampling.test.ts` — determinism, 5% rate produces ~500/10000 keep, WARN/ERROR not affected.
  - `lib/achievements/__tests__/verify-timeout-cache.test.ts` — fake-timer scenario asserts second call within 60s short-circuits; non-timeout error doesn't poison the sentinel namespace.
  - `lib/inbox/__tests__/x402-verify.test.ts` — warn-downgrade behaviour, only the `isExpectedNonceFailure` branch flipped, unexpected payment failures still ERROR.
- [ ] post-deploy smoke: 15-30min check on `aibtc-landing` ERROR rate via logs.aibtc.com.
- [ ] 24h post-deploy: confirm `aibtc-landing` INFO/day drops to the target via worker-logs `/stats/aibtc-landing`.

## Notes

- The sentinel is in a **separate `cache:verify-timeout:*` namespace** rather than co-mingled with `cache:tx:*` and `cache:stacking:*` — keeps `getCachedTransaction` and `getCachedStacking` semantically pure (they still return data or null), and the verify functions early-return on the sentinel before the data-cache read. The dedicated namespace also makes it trivial to bulk-purge sentinels in an incident.
- `isTimeoutError` matches both `error.name === "TimeoutError"` (the standard `AbortSignal.timeout` shape) and `error.code === "ERR_ABORTED"` for forward compatibility with whichever runtime detail Cloudflare changes next.
- The non-timeout-error test in commit 2 exercises the JSON parse path (`new Response("{not json")`) — that's the cleanest way to land in the `try/catch` with a non-timeout error without changing the function shape.


whoabuddy:
Addressed Copilot review feedback in b6250ad + cleanup in 4d8da5a:

**Copilot inline #1 (sampling.test.ts:28 — flaky 100-key test).** Bumped the batch from 100 to 10000 keys. At a 5% expected-keep rate, a 100-key batch has a non-trivial tail probability of zero keeps; 10000 keys makes the binomial outcome essentially deterministic against the FNV-1a distribution.

**Copilot inline #2 (sampling.test.ts:49 — test name vs assertion mismatch).** Rewrote the "differentiates by category" test to actually exercise the cross-category comparison: asserts the unconfigured-category default (rate=1, always keep) AND iterates 1000 keys against `cache.event` to prove both kept and dropped outcomes occur, which is the property the test name claims.

**Copilot inline #3 (kv-cache.ts:166 — broken {@link SAMPLE_RATES}).** Dropped the JSDoc symbol link; references `SAMPLE_RATES` in `lib/logging.ts` via plain prose since the const is non-exported and would render as a broken IDE link.

**Copilot inline #4 (logging.ts:229 — unguarded SAMPLE_RATES[category]).** Added `Number.isFinite` + `Math.max(0, Math.min(1, raw))` clamp. NaN / strings / negative values / values >1 now degrade safely to 1 (keep everything) instead of silently dropping 100% of events for that category. Also added a focused test asserting the clamp contract for unknown categories.

**Cleanup (4d8da5a).** `scripts/reconcile-staged-payments.sh` is unrelated to this PR (operator script for PR #586's reconciliation-queue migration); it got picked up by `git add -A` in the fix-up commit. Untracked here; file remains on disk locally.

**On the CI `Test` failure.** `lib/__tests__/bitcoin-verify.test.ts` fails with `p2tr: non-schnorr pubkey` at line 150 — this reproduces on bare `main` (verified by stashing this branch and running `npm test -- bitcoin-verify`). Same pre-existing typecheck errors are also on `main` in that file and `lib/inbox/__tests__/reconciliation-queue.test.ts`. Neither file is touched by this PR. Worth opening as a separate fix-up.

Focused suite this PR adds + touches: 22/22 pass.

arc0btc (approved):
Billing-reduction B3: deterministic 5%-sampled cache telemetry + 60s timeout sentinel for achievement verifies + expected nonce-rejection downgrade. All three changes are well-reasoned, tested, and already through Copilot review with feedback addressed.

**What works well:**
- FNV-1a for deterministic sampling is the right call here — the same key always wins/loses, so 5%-sampled hit-rate is still a meaningful spot-check metric, not random noise. The `>>> 0` unsigned-shift before `% 10000` is a nice correctness detail that avoids negative-modulo surprises.
- Sentinel namespace isolation (`cache:verify-timeout:*` vs. `cache:tx:*` / `cache:stacking:*`) is clean design. The verify functions get a clear short-circuit without contaminating data-cache semantics.
- `isTimeoutError` matching both `name === "TimeoutError"` and `code === "ERR_ABORTED"` is good forward-compatibility — we've seen Cloudflare surface these differently across runtime versions.
- The runbook doc is thorough — the rollback signal section is exactly what oncall needs.
- Copilot feedback already addressed (negative rate clamp, 10k-key batch, category-differentiation test). No re-litigating that.

**[question] Negative rate in SAMPLE_RATES would drop everything, not keep everything** (`lib/logging.ts:199`)

The commit message says "NaN / strings / negative / >1 now degrade to 1 (keep everything)" but negative values hit a different path: `Number.isFinite(-0.1)` is `true`, so they pass through `Math.max(0, Math.min(1, -0.1))` → `0`, then `rate <= 0` → `{ keep: false, rate: 0 }` — which drops every event. Only true-NaN and non-finite values (strings, `undefined`) fall back to 1. No practical risk since SAMPLE_RATES only has 0.05 today, but the safety contract in the docstring slightly overstates the protection. Consider: `Math.max(0.001, Math.min(1, raw))` or just update the comment to match the actual behavior.

**[question] Sentinel fires before data cache, could cause 60s false negative in a race** (`lib/achievements/verify.ts:90`)

If two concurrent requests for the same `btcAddress` land in the same isolate and one times out (sentinel written) while the other succeeds (tx data cached), the sentinel check fires first on subsequent calls for 60s — even though `getCachedTransaction` would have returned the data. Workers isolates are usually single-threaded so this race is unlikely, but the 5-min rate limiter on the verify path makes it even more theoretical. Flagging for awareness rather than blocking — the 60s window is a reasonable trade-off.

**Code quality notes:**
- The four verify functions have symmetric patterns (check sentinel → try/catch with timeout write) which is the right call — a generic wrapper would add indirection without benefit here.
- `VERIFY_TIMEOUT_CACHE_TTL = 60` could be a named constant with a comment explaining why 60s (just under the rate-limit window), but the existing style across this file is inline values, so no change needed.
- `isVerifyTimedOut` routes through `kvGet` → `logCacheEvent` → `samplingFor("cache.event", ...)` — sentinel reads are themselves sampled at 5%. Presumably intentional (they're not audit-grade either), just confirming.

**Operational context:** We process x402 nonce rejections in our own inbox flow (15+ daily, all `SENDER_NONCE_DUPLICATE/STALE/GAP`). Those have always been expected retry traffic and the WARN was noise on our end too — downgrade is correct. The mempool.space timeout pattern maps to what we see in our bitcoin-macro sensor (sporadic 10s+ hangs on the Blockstream.info and mempool.space APIs when Bitcoin transaction volumes spike). A 60s sentinel is well-calibrated for that failure mode.

whoabuddy:
Addressed arc0btc review on 6fbac4d:

**[question] Negative rate behavior** — fixed. Replaced `Number.isFinite(raw) ? Math.max(0, Math.min(1, raw)) : 1` with an explicit in-range check: anything outside `[0, 1]` (NaN/strings/undefined/negative/>1) falls back to 1 (keep everything). A legitimately configured rate of 0 still drops everything — that's the silence-via-config use case. The docstring and behavior now agree.

**[question] Sentinel + data-cache race** — confirmed for awareness, no action. The 5-min rate-limit window upstream of this path makes the race effectively unreachable in practice; even in the worst case the window is 60s of one false negative on a verify call that's already cheap to retry. Worth a follow-up if we ever lift the rate-limit window or move the sentinel to a multi-isolate-visible store.

**Confirmation: sentinel reads sampled at 5%** — yes, intentional. The sentinel KV reads are just `getCached*` calls in the same namespace pattern, so they pass through `logCacheEvent` → `samplingFor` like every other cache event. Sentinel hit/miss is a sanity-check stream too, not audit-grade.

**On VERIFY_TIMEOUT_CACHE_TTL as a named constant** — agreed it could be promoted, but as you noted, the surrounding file's style is inline values; leaving as-is to keep the diff scoped.

Targeted suite still passes 7/7. Ready to merge per your one-pass review convention.

arc0btc (approved):
Follow-up on the post-review commit `6fbac4db` — fix(logging): make samplingFor clamp behavior match its docstring.

The change is correct. The old code `Math.max(0, Math.min(1, raw))` would silently clamp negative rates to 0, causing those categories to drop 100% of events — the opposite of safe degradation. The new guard:

```ts
const rate = Number.isFinite(raw) && raw >= 0 && raw <= 1 ? raw : 1;
```

treats any out-of-range value (negative, >1, NaN, non-finite) as misconfiguration and falls back to `keep=true`. This is exactly the right failure mode: a misconfigured rate should never silently drop production events.

The comment update is also clearer — it explicitly calls out that `rate=0` is a valid intentional use case (silence a category in code), distinct from the misconfiguration fallback.

No new concerns. Still APPROVED.

Issue #626, opened 2026-04-21
Optimize: remove achievements


whoabuddy:
Do we really need these? Asking from my SpaceX 5 principles lens.

whoabuddy:
Deferring this to an RFC. The underlying idea is sound, but it needs design work before implementation — too large to fit in the current sprint without proper scoping.

Labeling 'RFC' for future planning. If you'\''d like to drive the RFC, please open a Discussion or comment here with a sketch and we'\''ll pick it up. Otherwise we'\''ll re-evaluate after the current sprint settles.

Issue #296, opened 2026-02-27
evalute using D1 instead of KV


whoabuddy:
We need to analyze how many keys/values are being stored now - as agent profiles expand or we want to store more information may make sense to migrate over to D1.

whoabuddy:
Deferring this to an RFC. The underlying idea is sound, but it needs design work before implementation — too large to fit in the current sprint without proper scoping.

Labeling 'RFC' for future planning. If you'\''d like to drive the RFC, please open a Discussion or comment here with a sketch and we'\''ll pick it up. Otherwise we'\''ll re-evaluate after the current sprint settles.

arc0btc:
Sounds right — the current sprint priority makes sense. One note: Discussions are disabled on this repo, so an RFC would need to land as a new issue or in the PR description when implementation starts.

The schema draft and hybrid approach from this thread can serve as the RFC skeleton when you're ready to pick it up. Happy to flesh it out into a structured issue if that's useful.

PR #651, opened 2026-05-06
feat: trading-comp dashboard with multi-token portfolio + USD totals

biwasxyz:
## Summary

Adds a public **trading-competition dashboard** at `/dashboard` (UX) and `/api/dashboard` (AX) that shows every registered agent's full token portfolio — BTC L1, STX, sBTC, and any SIP-10 holdings — valued in USD and sorted by total descending.

Built to support the upcoming trading comp launch where ranking by current portfolio value matters.

### New module: `lib/balances/`
- `types.ts` — `TokenBalance`, `AgentBalance`, `DashboardSnapshot`
- `constants.ts` — KV keys, TTLs, decimals, mempool.space + CoinGecko URLs, sBTC contract id, concurrency + timeout knobs
- `prices.ts` — `getPriceSnapshot(kv, logger)` — single CoinGecko call → BTC + STX prices, cached in `cache:prices` for 5 min. sBTC priced as 1:1 with BTC.
- `fetch.ts` — `fetchAgentBalances(stx, btc, prices, kv, hiroApiKey, logger)` — runs BTC L1 (mempool.space) + Stacks `/extended/v1/address/{principal}/balances` in parallel; returns `{ tokens, totalUsd, partial? }`. Honours the upstream-failure sentinel.
- `snapshot.ts` — `getDashboardSnapshot()` (SWR) + `invalidateDashboardSnapshot()`. Mirrors `lib/cache/agent-list.ts` exactly: 2 min fresh / 10 min hard, sentinel-gated rebuild, cold-miss polling.
- `index.ts` — barrel.

### Endpoint: `GET /api/dashboard`
- `?docs=1` — self-documenting payload
- `?limit` (1–500, default 100), `?offset` — pagination over the cached snapshot
- Returns `{ agents, prices, stats, cachedAt, pagination }`
- `Cache-Control: public, max-age=30, s-maxage=60, stale-while-revalidate=300`
- Self-doc cached `s-maxage=86400`

### Page: `/dashboard`
- Server-rendered table (rank, agent, BTC, STX, sBTC, other-token count, Total USD)
- Stats strip: total agents / total portfolio value / current BTC price
- Snapshot freshness indicator (\"Snapshot updated Ns ago. Refreshes every ~2 min.\")
- Inline link to each agent profile
- Empty state + \"partial\" flag rendered when an agent's balance fetch was incomplete

### Discovery + nav wiring
- `app/components/Navbar.tsx` — adds **Dashboard** to desktop + mobile nav
- `app/.well-known/agent.json/route.ts` — adds `tradingDashboard` under levels
- `app/llms.txt/route.ts` — endpoint listing + Pages section
- `app/llms-full.txt/route.ts` — adds full **Trading Dashboard** section with curl examples
- `app/api/openapi.json/route.ts` — adds full path object for `/api/dashboard` with response schema

## Cost guardrails (this is the headline)

The runbook (`docs/cloudflare-cost-runbook.md`) shipped three fixes (B3, B6, B6.2) earlier this sprint to absorb a sudden \\$800 bill caused by hot-path KV scans, log noise, and unbounded upstream retry loops. **Every cache decision in this PR mirrors those patterns exactly.**

| Concern | Mitigation | Pattern source |
|---|---|---|
| Per-request worker invocations | Edge cache header on `/api/dashboard` (`s-maxage=60, stale-while-revalidate=300`) and a server-rendered `/dashboard` page so most reads never hit origin | B6.3 direction |
| Repeated rebuilds under load | One `cache:dashboard` KV key + SWR (2 min fresh / 10 min hard) + `cache:dashboard:building` sentinel (90 s TTL) gating rebuilds | mirrors `lib/cache/agent-list.ts` |
| One slow agent forcing every rebuild to retry the same dead upstream | 60 s `cache:dashboard:upstream-fail:{btc,stx}:{addr}` sentinel — set on Hiro/mempool failure, checked at fetch entry | B3 (\\`cache:verify-timeout:*\\`) |
| Hot-path KV scans (the \\$5/day bleed B6 just fixed) | Reuses `getCachedAgentList()` for the agent set — **no new `kv.list({prefix:\"stx:\"})`** | B6 |
| Per-agent KV writes (cache fan-out) | **None** — the snapshot already holds every agent's balance, so a per-agent cache layer would be redundant. Considered, then dropped. | — |
| Hiro quota | Reuses `stacksApiFetch` (which already enforces 8 s per-attempt timeout, 429 retry budget = 2, 5xx retry budget = 1, monthly-quota warn at 20 % remaining) | shared util |
| Mempool.space failures spamming logs | Hand-rolled `AbortSignal.timeout(8s)`; **single WARN** per failure (not per request, because of the sentinel) | B3 |
| INFO log noise | Cache events gated by `samplingFor(\"cache.event\", ...)` at 5 %; no INFO on the hot path. WARN/ERROR only on real anomalies. | B3 |

### Per-rebuild cost ceiling

For ~430 current agents (per the runbook's S0 baseline):
- **1** `cache:agent-list` read (already cached)
- **1** `cache:prices` read (CoinGecko hit only on cold price cache, ~1× / 5 min)
- **~430** Hiro `/balances` calls + **~430** mempool.space calls per rebuild, batched at concurrency 10
- **1** `cache:dashboard` write (the snapshot)
- **0** per-agent KV writes (no fan-out)

SWR caps rebuilds to one per ~2 min worst-case. Per hour: 30 rebuilds × 430 = **~12,900 Hiro calls/hr peak**. With the per-attempt timeout + 60 s upstream-fail sentinel, sustained dead-upstream traffic can't escalate this.

## Files changed

```
M  app/.well-known/agent.json/route.ts
A  app/api/dashboard/route.ts
M  app/api/openapi.json/route.ts
M  app/components/Navbar.tsx
A  app/dashboard/page.tsx
M  app/llms-full.txt/route.ts
M  app/llms.txt/route.ts
A  lib/balances/constants.ts
A  lib/balances/fetch.ts
A  lib/balances/index.ts
A  lib/balances/prices.ts
A  lib/balances/snapshot.ts
A  lib/balances/types.ts
```
13 files, +1,225 / -0.

## What's intentionally NOT in v1

These are deferred to keep the surface tight for the launch — flag them in review if any should be in scope before merge:

- **Baseline snapshot / P&L.** Right now the dashboard is current-value only. A trading comp ideally ranks on \"started with X, now has Y\" — that's a separate feature (capture a baseline at comp start, store in KV, render delta column).
- **Stacked STX (PoX).** `/balances` returns liquid STX only. Stacked STX would need a separate `get-stacker-info` lookup. Probably worth adding before launch if Stackers participate.
- **SIP-10 pricing.** Tokens we cannot price are returned with `priceUsd: 0` and don't contribute to `totalUsd`. UI surfaces them as \"N tokens\" rather than dropping them.
- **Manual refresh button.** Operators can `wrangler kv:key delete cache:dashboard` to force a rebuild. Not a UI button — that would let any user trigger a rebuild, which defeats the SWR.
- **Cron pre-warm.** No scheduled rebuild. A cron means we hit upstreams every cycle whether or not anyone is looking; SWR means we only rebuild when there's traffic. Cheaper.
- **Per-agent dashboard route.** `/dashboard/{address}` for one agent's full breakdown isn't here — agent profile pages already show identity, achievements, etc. Adding balance there is a follow-up.

### Local
- [x] `npx tsc --noEmit` — clean
- [x] `npm run lint` — clean (only pre-existing warnings in unrelated files)
- [x] `npm run build` — builds, both `/dashboard` and `/api/dashboard` listed in route table

### Manual smoke (post-deploy preview)
- [ ] `curl https://<preview>/api/dashboard?docs=1` returns the self-doc payload
- [ ] `curl https://<preview>/api/dashboard | jq '.stats'` — non-zero `total`, sensible `totalUsd`, recent `pricedAt`
- [ ] `curl -I https://<preview>/api/dashboard` — `Cache-Control: public, max-age=30, s-maxage=60, stale-while-revalidate=300`
- [ ] Hit `/dashboard` in browser — table renders, ranks by total USD desc, partial flag shown only when an upstream failed
- [ ] Hit it again immediately — sub-100ms (edge cache served)
- [ ] After 2+ min — page still loads instantly (stale snapshot served), background rebuild kicks off
- [ ] Force-clear `cache:dashboard` via wrangler — next request rebuilds; concurrent requests serve via cold-miss poll, not duplicate rebuilds

### Cost watch (24 h post-deploy)
- [ ] `https://logs.aibtc.com/stats/aibtc-landing` — INFO/day stays at the post-B3 floor (no new per-agent INFO chatter)
- [ ] Cloudflare KV reads on `VERIFIED_AGENTS` — no spike beyond the existing baseline (we read 1 cache key per request, 1–2 keys per rebuild)
- [ ] Hiro monthly quota (`x-ratelimit-remaining-stacks-month` in `stacksApi.approaching_monthly_quota` warns) — confirm we're not driving toward the 20 % threshold

### Rollback signal
- 5xx on `/api/dashboard` — likeliest cause is a malformed CoinGecko response or a Hiro path change. Mitigation: the upstream-fail sentinel auto-skips the bad fetch within 60 s; the snapshot writer tolerates partial agents (`fetchError: \"partial\"`). Worst case: revert this PR — no schema migrations, no write-path changes, fully additive.


arc0btc (approved):
Adds a trading-comp portfolio leaderboard at \`/dashboard\` (SSR) and \`/api/dashboard\` (JSON) — every registered agent's BTC L1, STX, sBTC, and SIP-10 holdings valued in USD, sorted by total. Clean addition, well-scoped, no migrations needed.

**What works well:**

- Cost discipline is excellent. The SWR + sentinel-gated rebuild pattern mirrors \`lib/cache/agent-list.ts\` exactly, the B3 runbook's upstream-failure sentinel (\`cache:dashboard:upstream-fail:{scope}:{addr}\`) is applied correctly, and the rebuild loop reuses \`getCachedAgentList\` so there's no new \`kv.list\` scan. These are the right choices.
- \`toAmount\` in \`lib/balances/fetch.ts\` handles BigInt arithmetic correctly — integer division first, fractional component preserved separately. This avoids the silent precision loss you'd get from dividing a raw BigInt directly to float.
- \`Intl.NumberFormat\` instances (\`fmtUsd\`, \`fmtAmount\`) are at module scope, not recreated per render. Good.
- BTC L1 + Stacks fetches run in parallel per agent (\`Promise.all\`), then batched at \`BALANCE_FETCH_CONCURRENCY = 10\` across the agent list. Right balance between throughput and Hiro quota pressure.
- Edge cache headers are correctly applied on the API route (\`s-maxage=60, stale-while-revalidate=300\`), and \`force-dynamic\` on the page prevents a stale build-time snapshot.
- Discovery wiring is thorough — llms.txt, llms-full.txt, agent.json, openapi.json, and Navbar all updated in one PR.

---

**[suggestion] Unknown SIP-10 \`amount\` is raw microunits, not decimal-adjusted** (\`lib/balances/fetch.ts\` in \`buildSip10Balance\`)

For tokens we can't identify, the code sets \`decimals: 0\` and \`amount: Number(raw)\`. Mathematically consistent (\`raw / 10^0 = raw\`), but it conflicts with the \`TokenBalance\` type's own doc: *"Human-readable balance: balance / 10^decimals."* An agent holding 1,000,000 microUSDA (= 1 USDA at 6 decimals) will show \`amount: 1000000\`. This doesn't affect the USD total (\`priceUsd: 0\`) or the current UI (which only shows the token count), but any future per-agent breakdown that renders \`amount\` will be off by 6–8 orders of magnitude.

Suggest clarifying the comment and signaling to callers that \`amount\` is not reliable for unknown SIP-10s:

```suggestion
  // Decimals unknown — we can't compute a human-readable amount.
  // amount is left 0; consumers should use balance (raw string) if needed.
  return {
    symbol: tokenName ?? assetId,
    contract,
    balance: raw,
    decimals: 0,
    amount: 0,
    priceUsd: 0,
    usdValue: 0,
  };
```

**[suggestion] Error response leaks internal message** (\`app/api/dashboard/route.ts\` catch block)

\`(e as Error).message\` is included verbatim in the 500 response body. Internal messages can reveal implementation details (snapshot key names, KV namespace, etc.). Consider a generic message:

```suggestion
      { error: "Failed to fetch dashboard snapshot" },
```

**[nit] Table \`<th>\` missing \`scope="col"\`** (\`app/dashboard/page.tsx\` thead)

Screen readers use \`scope\` to associate headers with the correct cells. Seven \`<th>\` elements, all missing it:

```suggestion
                  <th scope="col" className="px-4 py-3">#</th>
                  <th scope="col" className="px-4 py-3">Agent</th>
                  <th scope="col" className="px-4 py-3 text-right">BTC</th>
                  <th scope="col" className="px-4 py-3 text-right">STX</th>
                  <th scope="col" className="px-4 py-3 text-right">sBTC</th>
                  <th scope="col" className="px-4 py-3 text-right">Other</th>
                  <th scope="col" className="px-4 py-3 text-right">Total USD</th>
```

**[nit] \`animate-ping\` without reduced-motion guard** (\`app/dashboard/page.tsx\`, the pulsing green badge dot)

The \`animate-ping\` Tailwind class runs continuously. Users with \`prefers-reduced-motion: reduce\` should see a static dot:

```suggestion
                <span className="relative flex size-2">
                  <span className="absolute inline-flex h-full w-full motion-safe:animate-ping rounded-full bg-green-400 opacity-75" />
                  <span className="relative inline-flex size-2 rounded-full bg-green-500" />
                </span>
```

---

**Code quality notes:**

- \`getDashboardSnapshot\` is called without a \`logger\` in \`page.tsx\` (the page passes \`undefined\` for that param). Minor inconsistency vs. the API route — not a bug since \`logger\` is optional, but the page will silently swallow price upstream errors that the API route would WARN about.
- The \`shortAddress\` function shows last 4 chars (\`addr.slice(-4)\`). Stacks addresses differ in their suffix so this should be distinguishable, but BTC bech32 addresses share a common prefix — you might consider slicing more chars for disambiguation at scale.

---

**Operational note:** We monitor mempool.space and Hiro daily through our sensors. Hiro's \`/extended/v1/address/{principal}/balances\` endpoint has been stable, though we've seen occasional 429s during high-traffic windows (usually around block confirmations). The 60s upstream-fail sentinel in this PR handles exactly those spikes — it's the right call.

No blocking issues. Scope is tight for v1, the deferred items (baseline P&L, stacked STX, SIP-10 pricing) are clearly documented.

biwasxyz:
Thanks @arc0btc! Your review caught the original USD-valuation version of this PR — there have been three follow-up commits since (`d90e390` → `f287bd4` → `055c0d3`) that re-architected on top of your feedback. Mapping each comment to its current state:

**Addressed in `055c0d3`:**
- ✅ **Error response leaks internal message** — 500 catch in `app/api/dashboard/route.ts` now returns a generic `"Failed to fetch dashboard"`.
- ✅ **`<th scope="col">`** — added on all five header cells in `app/dashboard/DashboardList.tsx` (the columns are now sBTC / BTC / STX after the no-USD refactor).
- ✅ **`animate-ping` → `motion-safe:animate-ping`** — applied on the green pulsing dot in `app/dashboard/page.tsx`.

**Moot due to refactors:**
- 🪦 **SIP-10 `amount` is raw microunits** — `buildSip10Balance` was deleted in `d90e390`. We dropped USD valuation entirely (CoinGecko was returning zeros from the workers.dev IP anyway), and stopped emitting any SIP-10 token other than sBTC. The token surface is now `BTC | STX | sBTC` only.
- 🪦 **`getDashboardSnapshot` logger inconsistency** — that whole module is gone. The big-fleet snapshot architecture timed out on cold deploys (~860 cross-internet fetches per rebuild can't complete inside the user request window, and once the 10-min hard TTL expired on a low-traffic preview it never warmed back up). Replaced in `f287bd4` with a paginated fetcher (`getDashboardPage`) that only loads 10 agents at a time, plus a per-agent `cache:balance:{btc}` (60-s TTL) so repeat views are cheap.
- 🪦 **Table columns** — the "Other" and "Total USD" columns are gone with the no-USD refactor.

**Operational note:** thanks for the heads-up on Hiro 429s around block confirmations. The 60-s upstream-failure sentinel still applies in the new architecture — when Hiro 429s for a specific agent, that agent's `cache:dashboard:upstream-fail:stx:{addr}` key blocks retries for 60 s, so the spike doesn't propagate.

Pushed `055c0d3` to address the still-applicable items. Let me know if you want a fresh review of the new shape.

whoabuddy (changes requested):
Thanks for the refactor after the first review. Looking at the current shape with the recent Cloudflare cost work in mind, I think this needs another pass before merge.

## Blocking concerns

**1. This no longer implements a trading leaderboard.**

`getDashboardPage()` slices `cache:agent-list` before fetching balances, so the API/page order is whatever the agent-list/index order is, not `sBTC desc -> BTC desc -> STX desc`. The UI also renders rows in returned order. That means an agent with the highest sBTC balance can be on page 50 and will never appear near the top.

That is a product-contract issue, not just docs: the PR title/body, OpenAPI, llms-full, llms.txt, and page metadata all describe a trading-comp leaderboard.

Either we need a real ranked snapshot/leaderboard design, or we should rename this as a paginated balance browser and remove the ranking claims.

**2. The current per-page fetch design has a bad cost shape for public traffic.**

The old full-snapshot design had latency problems, but it bounded rebuilds behind one cache key and a sentinel. The new route lets any public caller walk arbitrary `offset`/`limit` combinations. On a cold cache, one `limit=100` request can trigger roughly:

- 1 `cache:agent-list` read
- 100 `cache:balance:*` reads
- up to 200 external upstream calls (Hiro + mempool)
- up to 100 `cache:balance:*` writes
- plus upstream-fail sentinel reads/writes on failures

Because `offset` and `limit` are query-controlled, edge caching does not protect much against a caller scanning pages or varying params. With ~942 agents currently, one pass over the dashboard can populate ~942 per-agent KV entries and make ~1,884 upstream calls if cold. That is exactly the kind of public hot-path fan-out we’ve been removing in the last few PRs.

The 60s per-agent cache helps repeat views, but it also creates write fan-out and does not coalesce concurrent misses for the same page/agent. A burst to the same uncached page can still duplicate upstream work before the first writes land.

**3. The published API docs are stale and misleading.**

OpenAPI still says:

- sorted by `sBTC desc -> BTC desc -> STX desc`
- snapshot cached for ~2 min at origin and ~60s edge
- `cachedAt` exists
- `stats` exists
- `limit` max/default are `500/100`

The actual route returns only `{ agents, pagination }`, clamps `limit` to max `100`, defaults to `10`, and uses `Cache-Control: public, max-age=15, s-maxage=30, stale-while-revalidate=300`.

`llms-full.txt` and `llms.txt` have the same stale snapshot/ranking language.

## Suggested direction

Given the cost work we’ve just done, I would avoid public requests causing unbounded per-agent upstream fan-out. A safer shape would be one of:

- Reintroduce a bounded ranked snapshot, but build it outside the request path or behind a strict single-flight/sentinel and serve stale data aggressively.
- Keep this paginated design, but make it explicitly a balance browser, not a leaderboard, reduce the public max limit, add tighter query normalization/rate protection, and consider not caching per-agent fan-out writes unless there is clear traffic evidence.
- Precompute only the competition participants/ranking set if that set is smaller than all registered agents.

Until the contract and cost model are tightened, I don’t think this should merge.

biwasxyz:
Thanks for the careful read @whoabuddy — you're right that the previous shape was a paginated balance browser wearing a leaderboard's title. Pushed `8dd5853` which addresses the three blockers head-on by combining your suggested directions #1 (bounded ranked snapshot, single-flight, served stale) and #3 (precompute only the participant set).

The trading comp's participant set = Genesis-level agents (Level 2+, gated by posting on X via `/api/claims/viral`). That's bounded and self-selecting, so the snapshot only needs to cover them.

**Mapped to your concerns:**

1. **"This no longer implements a trading leaderboard"** — Fixed. `lib/balances/snapshot.ts` filters `cache:agent-list` to `level >= 2`, fetches BTC/STX/sBTC for each, and sorts via BigInt comparison: sBTC desc → BTC desc → STX desc. The full ranked array is what's persisted and served. No more "sort the visible page client-side" trick.

2. **"Bad cost shape for public traffic"** — Fixed. Public requests only ever read **one KV key** (`cache:dashboard:snapshot`). Rebuilds are single-flighted by `cache:dashboard:snapshot:building` (60 s sentinel TTL) and run off-request via `ctx.waitUntil()`. The per-agent `cache:balance:*` writes are gone entirely — the snapshot *is* the cache, one write per rebuild instead of N.

   Concretely, on a cold-cache request:
   - 1 read of `cache:dashboard:snapshot` (miss)
   - 1 read of `cache:dashboard:snapshot:building`
   - If another request is rebuilding: poll briefly (1.5 s, every 150 ms) for the snapshot to land, then return it.
   - If not: claim the sentinel, rebuild once, write one snapshot key, drop the sentinel.

   No `offset`/`limit` to walk, no per-agent cache to populate, no concurrent-miss duplication.

3. **"Published API docs are stale"** — Fixed in the same commit. `openapi.json`, `llms.txt`, `llms-full.txt` all describe the actual contract now: Genesis-only, full ranked array, returns `{agents, stats: {total, partialCount}, cachedAt}`, no pagination params, 60 s freshness / 300 s hard TTL on the snapshot. Edge cache headers stay at `s-maxage=30, stale-while-revalidate=300`.

**Bonus drops** — `getDashboardPage`, `getCachedAgentBalance`, `DASHBOARD_PAGE_SIZE`, `BALANCE_CACHE_*`, and the client-side "Load more" fetch are all gone. Net: −350 LOC across 12 files.

The B3 upstream-failure sentinel pattern (`cache:dashboard:upstream-fail:{scope}:{addr}`) is preserved on the rebuild path, so a single bad agent still can't make every rebuild retry the same dead Hiro/mempool endpoint.

Ready for another pass.

whoabuddy:
Thanks for the quick turnaround. The latest `8dd5853` shape is much closer to what we need: Genesis-only participant set, one ranked snapshot key, no public `offset`/`limit` walking, no per-agent `cache:balance:*` write fan-out, and the preview response is now correctly all Genesis agents and sorted by `sBTC -> BTC -> STX`.

One remaining operational issue before I’d clear the change request:

**Cold miss still rebuilds synchronously, and the hard TTL is only 300s.**

In `getDashboardSnapshot()`, once `cache:dashboard:snapshot` expires, a public request falls through to `rebuildSnapshot()` and waits for the full Genesis balance fan-out. On the preview, my first live `/api/dashboard` request after the new deploy did not return for over a minute before I killed the local curl. A later request was fast because the snapshot had landed by then.

That means any quiet window longer than 5 minutes can put the next visitor back on the expensive synchronous rebuild path. With the current preview snapshot showing 568 Genesis agents, a rebuild can still mean ~1,136 upstream calls before the response can complete on a true cold miss.

Suggested fix: keep stale data around much longer than the freshness window, and only block synchronously when there is truly no viable snapshot at all. For example:

- `freshWindow`: 60s
- `hardTtl`: hours, not 300s, or no KV expiration plus an internal max-age field
- on stale hit: return stale and `waitUntil(rebuild)`
- on cold/no snapshot: either return a small empty/loading snapshot and `waitUntil(rebuild)`, or accept one synchronous bootstrap only after deploy/manual invalidation

That preserves the cost win from the new design while avoiding periodic cold-start stalls. The new ranked snapshot contract itself looks good.

whoabuddy:
Pushed a final hardening pass in a320604 before merge review.

What changed:
- Extended the dashboard snapshot hard TTL from 5 minutes to 1 hour so low-traffic gaps are much less likely to turn the next visitor into a cold synchronous rebuild.
- Increased the rebuild sentinel TTL from 60s to 120s to better cover slower Genesis-balance rebuilds.
- Updated route comments, self-docs, OpenAPI, and llms docs to be explicit that warm/stale requests read one KV snapshot key, while a true cold miss can still synchronously seed the first snapshot.

Local checks:
- npm test: 31 files / 581 tests passed, 5 skipped
- npm run lint: passed with existing img warnings
- npm run typecheck: blocked by pre-existing MessageBatch test fixture metadata errors in lib/inbox/__tests__/reconciliation-queue.test.ts, unrelated to this PR's files

@arc0btc and @biwasxyz can you give this one final maintainer review before we squash-merge? The original cost/correctness blockers look addressed now; the remaining cold-miss behavior is documented and made much rarer by the longer hard TTL.

whoabuddy (approved):
Reviewed latest b386cba after the cache hardening and /leaderboard rename. The original blockers are addressed: ranked Genesis-only snapshot, no public offset/limit walking, no per-agent balance-cache write fan-out, docs match the current /api/leaderboard/portfolio contract, and all remote checks are green.\n\nThe remaining cold-miss poll window note is worth a follow-up, but I agree it is not blocking for this PR now that the hard TTL is 1h and the behavior is documented. Approved for squash merge.

arc0btc (commented):
Re-reviewed `b386cba8` with a narrow lens: Hiro quota and Cloudflare billed usage. The architecture is the right shape — one snapshot key, single-flight rebuild, off-request `waitUntil`, Genesis-only filter, edge cache, no public `offset/limit` walking, no per-agent KV write fan-out. That's the cost-discipline checklist from the runbook, and it lands.

One concrete cost-correctness issue stands out, and it's the same class of risk the original review flagged.

---

**[blocking-ish] Sentinel TTL is shorter than the observed rebuild time → duplicate rebuilds under cold cache**

`lib/balances/constants.ts:38` sets `SNAPSHOT_BUILDING_TTL_SECONDS = 120`. biwasxyz's preview validation timed the cold rebuild at ~457s. The single-flight contract in `getDashboardSnapshot` / `maybeTriggerBackgroundRebuild` only holds while the sentinel is alive. With sentinel TTL < rebuild time:

- t=0 — request A claims sentinel, starts rebuild (snapshot still cold or hard-TTL-expired)
- t=120 — sentinel auto-expires, request A still rebuilding
- t=120+ε — request B sees no sentinel, no fresh snapshot, claims a new sentinel, starts a **second** rebuild (568 × 2 = 1,136 upstream calls again)
- t=240+ε — same again, request C starts a **third** rebuild

Under a small concurrent burst on cold cache (post-deploy, post-1h-gap, post-`wrangler kv:key delete`), this multiplies Hiro/mempool fan-out by however many parallel rebuilds slip in during the rebuild window. That's the exact "cold-miss requests multiplying upstream load" pattern biwasxyz called out for `COLD_MISS_POLL_MS=1500`, but caused by the sentinel TTL itself, not just the poll budget — once the sentinel expires, even a sequential second request wins the right to start a duplicate rebuild.

**Suggested fix:** make the sentinel TTL strictly larger than the p99 expected rebuild time. With 568 Genesis agents @ concurrency 10 and worst-case Hiro 429-retry budget, 600s (10 min) would absorb the observed 457s plus headroom. The sentinel still gets cleared in `finally`, so this only matters on rebuild crash/abort.

```diff
-export const SNAPSHOT_BUILDING_TTL_SECONDS = 120;
+export const SNAPSHOT_BUILDING_TTL_SECONDS = 600;
```

While we're here: bumping `COLD_MISS_POLL_MS` to ≥30s (biwasxyz's earlier nit) becomes more attractive once the sentinel actually outlives the rebuild — concurrent cold-miss requests can reliably piggyback instead of falling through.

---

**Other Hiro/CF observations (non-blocking, worth tracking):**

1. **Subrequest budget at scale.** Each rebuild does ~`2 × N` upstream fetches plus per-agent sentinel reads, and `stacksApiFetch` can multiply Hiro calls by up to 4× under 429+timeout retries. At Genesis=568 the happy path is already ≥1,136 upstream subrequests + 1,136 sentinel reads ≈ 2,272 KV+fetch ops in one `waitUntil` callback. Workers Standard accommodates this today, but the rebuild cost grows strictly linearly with the Genesis set. If Genesis crosses ~2,000, this design needs a sharded snapshot (e.g., `cache:dashboard:snapshot:{shard}` with a coalesce step) or a cron-driven rebuild that doesn't share the request invocation's subrequest budget.

2. **mempool.space has no API key and no per-IP backoff.** 568 sequential-batched calls per rebuild from one Worker egress IP is fine at current scale but worth a sentinel-aware rate-limit reaction (count consecutive `markUpstreamFailed` for `scope=btc` and short-circuit the rest of the batch if they all fail — protects rebuild time when mempool throttles us).

3. **No Hiro monthly-quota circuit breaker on the rebuild path.** `stacksApiFetch` already logs `approaching_monthly_quota` at 20% remaining, but the rebuild keeps fanning out 1,136 calls regardless. Wiring `approaching_monthly_quota` into a "skip Stacks fetches and serve last-known stale snapshot" branch would be a cheap safety valve before the comp drives traffic.

4. **SWR rebuild cadence.** Steady-state with traffic, the snapshot churns one rebuild per `(fresh + rebuild)` interval ≈ 60s + ~60–460s. That's ~6–60 rebuilds/hr = ~7k–68k upstream calls/hr at current Genesis count. Within Hiro paid budget today, will need rethinking if Genesis 3–4×s.

5. **`force-dynamic` on the SSR page is correct** — it ensures `getDashboardSnapshot` is hit per request rather than served from build-time. Worth keeping as a guard against future Next.js optimizations.

6. **Edge cache `s-maxage=30` vs snapshot fresh window 60s** — fine, edge revalidates at 30s and gets a still-fresh snapshot, no extra rebuild cost. Just noting the asymmetry is intentional.

---

**Bottom line.** The cost shape is correct in steady state. The one concrete defect is the sentinel TTL not actually covering the rebuild — under cold cache or hard-TTL expiry, that's the failure mode that produces the multiplied Hiro fan-out we set out to prevent. One-line fix; happy to re-approve once that lands. Everything else is forward-looking.

whoabuddy:
Filing this under the new meta-issue (#652) so the strategic context is visible — this PR sits at the intersection of the residual May cost campaign, the SpaceX-5 simplification (#626), and the long-deferred D1 migration (#296).

**Credit where due:** the iterative scope reduction in this PR is exactly right. Dropping USD/CoinGecko, swapping the big snapshot for a paginated top-10, scoping to Genesis-only, and hardening cold-miss behavior have all materially improved this. The cost-guardrails table in the PR body is the kind of artifact every PR like this should ship with.

**Strategic context, not a request to hold:** the trading-comp dashboard is on the *keep* list in the meta-issue — it's a likely next-activity surface for Genesis agents, alongside `/bounty`. So this PR can keep moving.

The framing we want to be transparent about:

### The main billing issue is KV pattern, not this PR

Today's measurement at peak: **~22 KV reads per inbound request** at <1K registered agents. That's an architectural smell, not a leak — KV is being used as a relational DB across 40+ call sites, and B6.1 / B6.2 / B6.3 only wrapped 3 routes in `caches.default`. The residual is structural and won't be fixed by wrapping more routes.

The campaign response is twofold: (1) drop accretion (#626 achievements is the first cut), (2) migrate the relational data to D1 so `rebuildAgentListCache` becomes a `SELECT ... ORDER BY ... LIMIT N` instead of a `1 + 4N` KV fan-out. The trading-dashboard ranking is naturally a D1 query of the same shape.

### How this PR folds into the campaign

1. **Cache pattern.** PR body says "Mirrors `lib/cache/agent-list.ts` exactly: 2 min fresh / 10 min hard, sentinel-gated rebuild, cold-miss polling." The B6.3 inventory found that `agent-list`'s `invalidate*` semantics are a top KV-read leak — `kv.delete(CACHE_KEY)` forces every reader to pay the full N-fan-out cold rebuild. We're shipping a fix for that (B6.4 P1: mark-stale-and-background-rebuild instead of delete) as a sub-issue of the meta. **Ask: please apply the same mark-stale fix to `invalidateDashboardSnapshot` in this PR**, so we don't ship a fresh copy of the bug class while we're fixing it. Small change; PR-local.

2. **Hiro budget.** Achievement removal (#626) frees up Hiro budget; this PR re-spends it on portfolio fetches. Net effect should be neutral or positive once #626 lands, but worth coordinating landing order so #626 ships before this PR's first cron tick. No code change requested here, just sequencing awareness.

3. **D1 fold-in.** When the D1 migration ships (Phase 2 in the meta), the `cache:dashboard` SWR snapshot becomes a `SELECT FROM agents JOIN balances ORDER BY total_value DESC LIMIT N` query. The data layer in this PR (`lib/balances/{fetch,prices}.ts`) lifts cleanly to D1; the snapshot is the part that gets thrown away. That's fine — we'd rather have the dashboard live now than wait, and the swap is a single seam.

4. **S-tier fold-in.** The `stacksApiFetch` you're (correctly) reusing here is going to become `@aibtc/platform/stacks-api` shortly after this campaign closes — landing-page's `lib/stacks-api-fetch.ts` is the shape we'll be packaging. Not a blocker for this PR; the import path will move in a follow-up. We agreed in the meta that landing-page is the anchor for that package, so this PR's reuse is the right pattern.

### TL;DR

- ✅ Keep going.
- 🔧 One narrow ask: apply mark-stale to `invalidateDashboardSnapshot` so the cache-invalidate bug class doesn't propagate.
- 📅 Sequencing: ideally #626 lands before this PR's first cron tick, but flexible.
- 🔭 Folds into D1 migration (Phase 2) and `@aibtc/platform/stacks-api` (S-tier follow-up) — your `lib/balances/{fetch,prices}.ts` layer is reusable; the cache layer is throwaway.

Happy to pair on the `invalidateDashboardSnapshot` fix if useful — it's a one-liner pattern lift from the B6.4 P1 work.

secret-mars (commented):
**Concrete redesign spec — pivot this PR from balance-fetching to swaps-table-driven, with Tenero P/L.** Following up on #738 (Phase 3.1 verifier) landing + operator clarification: keep this surface, replace the data source. Specifics:

## What the leaderboard becomes

For each Genesis-level (Level 2+) agent that has done at least one verified trade — display **trade_count + P/L (USD)** sorted by P/L desc. Genesis filter is already there; the rest is new.

## Replace `lib/balances/*` fetch layer with a single D1 query

```sql
SELECT
  a.stx_address,
  a.btc_address,
  a.display_name,
  a.bns_name,
  s.token_in,
  s.amount_in,
  s.token_out,
  s.amount_out,
  s.burn_block_time,
  COUNT(*) OVER (PARTITION BY a.stx_address) AS trade_count_for_agent,
  MAX(s.burn_block_time) OVER (PARTITION BY a.stx_address) AS last_trade_at
FROM agents a
JOIN swaps s ON s.sender = a.stx_address AND s.tx_status = 'success'
WHERE  -- Genesis filter: use whatever level=2 SQL predicate the existing
       -- leaderboard route at app/api/leaderboard/route.ts uses
       <Genesis-level predicate here>
  AND s.burn_block_time BETWEEN ? AND ?  -- campaign window
ORDER BY a.stx_address, s.burn_block_time;
```

JS-side aggregates the per-trade rows → per-agent `{ trade_count, pnl_usd, volume_usd, last_trade_at }`. JOIN guarantees the **has-done-trades** filter: agents with zero successful swaps drop out by construction.

## P/L calculation — Tenero-priced

For each trade row:

```ts
const tokenInPriceUsd  = prices[normalize(s.token_in)]  // 0 for unpriced
const tokenOutPriceUsd = prices[normalize(s.token_out)]
const decimalsIn  = DECIMALS[s.token_in]  ?? 6
const decimalsOut = DECIMALS[s.token_out] ?? 6

const amountInUsd  = (s.amount_in  / 10 ** decimalsIn)  * tokenInPriceUsd
const amountOutUsd = (s.amount_out / 10 ** decimalsOut) * tokenOutPriceUsd

const tradePnlUsd = amountOutUsd - amountInUsd  // negative = loss
const tradeVolumeUsd = amountInUsd              // input notional
```

Per-agent: `pnl_usd = SUM(tradePnlUsd)`, `volume_usd = SUM(tradeVolumeUsd)`.

Caveat — this is **current-price P/L**, not price-at-burn_block_time. It tells you "what would your trades have netted if executed at today's prices." True historical P/L needs a `prices(token_id, captured_at)` D1 table (Phase 3.2 work). For v1, current-price is enough and ships today.

## Tenero price-fetch layer

`api.tenero.io`, no auth, public.

```ts
// One call per allowlisted token, 5min KV cache
const ALLOWLIST_TOKENS = [
  "SM3VDXK3WZZSA84XXFKAFAF15NNZX32CTSG82JFQ4.sbtc-token",        // sBTC
  "SP4SZE494VC2YC5JYG7AYFQ44F5Q4PYV7DVMDPBG.ststx-token",        // stSTX
  // … usdc, alex, usdh, etc.
];

async function fetchTokenPriceUsd(contractId: string): Promise<number> {
  const r = await fetch(`https://api.tenero.io/api/v1/stacks/tokens/${contractId}`);
  const data = await r.json();
  return data.price_usd ?? 0;
}
```

**Gotcha — native STX is not priced by Tenero directly** (`wstx` returns `price_usd: 0`, no liquidity on the wrapped contract). Two options:

A. **Seed STX from Pyth Hermes** — `0xec7a775f...` STX/USD feed (already mapped in my v54 Zest work). One additional HTTP call per snapshot.

B. **Derive STX from a Tenero pool** — `tenero_market_summary` on `SM1793C4R5PZ4NS4VQ4WMP7SKKYVH8JZEWSZ9HCCR.stableswap-pool-stx-ststx-v-1-4` exposes wstx-vs-ststx ratio; multiply by stSTX price_usd from Tenero. Single ecosystem, same cache window.

Recommend B — keeps it Tenero-only for the price layer, no Pyth dependency.

## Endpoint shape

```
GET /api/leaderboard/portfolio  // keep existing path for backwards-compat
GET /api/leaderboard/portfolio?sort=pnl|trades|volume  // new query param

Response:
{
  "agents": [
    {
      "stxAddress": "SP...",
      "btcAddress": "bc1q...",
      "displayName": "...",
      "bnsName": "...",
      "level": 2,
      "trade_count": 12,
      "pnl_usd": 47.23,
      "volume_usd": 1842.51,
      "lastTradeAt": "2026-05-11T03:23:00Z"
    }
  ],
  "prices": { "sbtc": 80440.92, "stx": 0.85, "ststx": 0.31, "usdc": 1.00 },
  "stats": {
    "totalAgents": 18,
    "totalTrades": 156,
    "totalVolumeUsd": 24871.50
  },
  "cachedAt": "2026-05-11T05:00:00Z"
}
```

`prices` block in response is for transparency — callers can see exactly which prices the leaderboard used to compute P/L. Helps debugging and helps frontends show "P/L computed at $1.0023 USDC" footnote.

## Sort options

- `?sort=pnl` (default) — `ORDER BY pnl_usd DESC`. The competition rank.
- `?sort=trades` — `ORDER BY trade_count DESC`. Activity rank.
- `?sort=volume` — `ORDER BY volume_usd DESC`. Whale rank.

Operator wanted "tradecounts and P/L displayed" — both are present in the same response; `sort` lets the UI render either as the primary rank without separate endpoints.

## Cost comparison vs current PR

| Concern | Current (#651) | This redesign |
|---|---|---|
| Per-request KV reads | ~3 (snapshot + freshness) | ~3 (snapshot + freshness) |
| Per-rebuild Hiro calls | 1 per Genesis agent | 0 |
| Per-rebuild mempool calls | 1 per Genesis agent | 0 |
| Per-rebuild price calls | 1 CoinGecko (BTC+STX) | ~10 Tenero (one per allowlist token), 5min KV-cached |
| Per-rebuild D1 queries | 0 | 1 (the swaps JOIN above) |
| Cost order vs N=Genesis agents | O(N) network | O(1) D1 + O(tokens) tenero |

Trade off: lose **current portfolio holdings view** (BTC + STX + sBTC current balances), gain **trading-competition leaderboard** (trade_count + P/L). The competition is about trading; this is the right semantic.

## What stays from this PR

- The page shell (`app/leaderboard/page.tsx`, `LeaderboardList.tsx`) — table layout, rank column, agent profile links, stats strip
- The discovery wiring (`agent.json`, `openapi.json`, `llms.txt`, `llms-full.txt`) — just update the response-shape docs
- SWR caching pattern (`getDashboardSnapshot`) — rebuild logic flips from "fetch balances" to "query swaps + multiply by prices"
- 60s fresh / 1h hard TTL → likely tighten to 60s/10min since cost drops dramatically

## What gets removed

- `lib/balances/fetch.ts` — Hiro + mempool fetching
- `lib/balances/prices.ts` — CoinGecko BTC+STX (replaced by Tenero per-token)
- `lib/balances/constants.ts` mempool/CoinGecko/concurrency constants
- `lib/balances/types.ts` `TokenBalance` → replaced by trade-aggregate type

## Sequencing

This redesign is shippable on top of #738 (currently APPROVED+pending merge). No new schema work, no Phase 3.2 dependency (Phase 3.2 = price-at-burn_block_time historical P/L — a future upgrade).

Estimated scope: ~200-300 LOC swap for the existing `lib/balances/` layer + ~50 LOC for the Tenero price fetch + UI tweaks in `LeaderboardList.tsx` to render the new columns (trade_count, pnl_usd, volume_usd). Tests: ~150 LOC (Tenero mock, swap aggregation, P/L math).

Happy to fork this PR or open a follow-up if you'd prefer to land the current portfolio leaderboard first and pivot in a subsequent PR. Operator preference: pivot in place.

---

cc @biwasxyz @arc0btc @whoabuddy — operator-requested spec on the current leaderboard PR. tl;dr: Tenero prices + swaps-table JOIN + Genesis filter + has-traded + trade_count + P/L.

secret-mars (approved):
**APPROVE on the rebased Portfolio-track state.** Agreement on the multi-track framing — this PR ships as the Portfolio track today; Score track is an additive follow-up endpoint that lands once #738 merges and #651's hygiene patterns are firmly established.

## Cost-correctness fixups land cleanly

**`dd7a8c2` — sentinel TTL 120s → 600s** is the right shape. The 457s cold-rebuild measurement against a 120s sentinel is a "single-flight isn't actually single-flight under realistic workload" failure mode. 600s gives p99 headroom on a 457s p50; the `finally` cleanup remains the happy-path clearer. Pattern carries to any future `cache:*:building` sentinel — worth pinning as a per-sentinel calibration rule (`sentinel_ttl >= 1.3 × p99(rebuild_time)`) if other rebuild paths emerge.

**`d711c3a` — cold-miss poll budget 1.5s → 30s, interval 150ms → 500ms** is the symmetric callee-side fix. 1.5s was undersized for the 457s cold rebuild → callers fell through and triggered duplicate rebuilds. 30s with 500ms interval = 60 polls/caller bounded cost vs the 2,272 ops a single duplicate rebuild costs. The ratio math is the right framing — "what's cheap for the caller to wait" vs "what's expensive to redo." Clean.

Both fixups are tightening the cost-discipline this PR established. The `lib/balances/snapshot.ts` pattern is now a textbook SWR-with-cold-miss-protection reference for future rebuild surfaces.

## Multi-track framing closure

Agreed with biwasxyz's read: this PR is the Portfolio track; the swaps-table-driven Score track (from my 04:52Z + 05:01Z comments) is an additive follow-up endpoint that interlocks with #738 (Phase 3.1 verifier) once that merges and starts populating the `swaps` table at scale. No code changes here for the Score track — keeps this PR's review surface tight + lets the Score-track design get its own focused review with concrete agent volumes on the new substrate.

When Phase 3.3 balances-cron eventually lands (per RFC §balances), this PR's `getDashboardSnapshot()` flips from "N×2 upstream calls" to "single D1 SELECT" — that's the natural next pivot for the Portfolio track. The current implementation is a correct stepping stone.

## Approvals reset note

Re-requesting was the right call after the force-push rebased the branch. arc-APPROVED and whoabuddy-APPROVED both got dismissed automatically; both review surfaces need re-clearing. CI all green on the rebased state (`d711c3a1`); mergeable should clear once arc + whoabuddy re-review.

## Verified clean

- CI: Lint / Test / Build / Workers Builds / CodeQL / snyk — all pass ✓
- Test count 968 passing (5 skipped) — same as pre-rebase ✓
- Rebase touched non-overlapping file regions vs #722/#720/#731/#732/#737 — clean merge per biwasxyz's note ✓
- Discovery docs unchanged (Genesis-only + sBTC→BTC→STX sort) — still accurate ✓
- `cache:dashboard:*` KV keys unchanged — no cold-rebuild churn on deploy ✓

## Forward-looking

When #738 merges and the Score track follow-up gets scoped, propose filing it as a new sub-issue under #652 referencing my 04:52Z + 05:01Z comments as the design substrate. Both Tenero-pricing approach (current-price P/L from `api.tenero.io` no-auth, ~10 token-info calls 5min-KV-cached) and STX-via-stableswap-pool gotcha (wstx contract has 0 liquidity on Tenero) are durable design notes that should travel into the Score-track PR.

LGTM on the rebased Portfolio track. Ready to merge once arc + whoabuddy re-clear.

whoabuddy (changes requested):
**Requesting changes** — dismissing my prior approval. The cost-discipline work here is solid and that part of my review stands. But there's a separate scaling cliff that has to be resolved before merge.

The dashboard rebuild does ~2 subrequests per agent (Hiro `/balances` + mempool.space). Cloudflare Workers Paid plan caps at **1000 subrequests per invocation**, not raisable via config:

| Agents | Subrequests per rebuild | Worker CPU | Status |
|---:|---:|---:|---|
| 430 (current) | ~860 | ~21s | Under both limits, barely |
| 500 | ~1000 | ~25s | At subrequest cap |
| 600 | ~1200 | ~30s | Both walls breached |
| 1000 | ~2000 | ~50s | Impossible in one invocation |

We hit this exact wall last week during Phase 1.4 inbox reconciliation. Documented in `CHECKPOINT-2026-05-10T0203Z-inbox-subrequest-limit.md`. The fix is architectural, not a tune.

**Required rework — filed as #764 with full spec:**

Pivot to the same architecture #738 took for `swaps`:
1. D1 `balances` as system of record (migration 006 already exists — currently dead schema; this PR is its intended writer)
2. Cron every ~2 min, processes one chunk of ~100 agents per invocation (~200 subrequests, well under cap)
3. Dashboard SSR reads from D1 with `caches.default` edge cache. No KV snapshot.

The good news: `lib/balances/fetch.ts`, `lib/balances/prices.ts`, the page UI, and all the discovery wiring (`llms.txt`, `openapi.json`, `agent.json`) carry over unchanged. The delta is `snapshot.ts` → cron handler + D1 read swap in the route/page.

Happy to pair on the D1 mapping — `balances` table fields are in `docs/rfc-d1-schema.md` §balances.

/cc @arc0btc @secret-mars for second opinions on the rework scope in #764.

Issue #655, opened 2026-05-08
Phase 0.2: invalidateAgentListCache mark-stale instead of delete


whoabuddy:
Sub-issue under [#652](https://github.com/aibtcdev/landing-page/issues/652) — pre-erc-8004 simplification + D1 migration umbrella.

## Goal

`invalidateAgentListCache(kv)` currently `kv.delete(CACHE_KEY)` at `lib/cache/agent-list.ts:142-150`. Switch to **mark-stale** semantics: rewrite the existing snapshot with `cachedAt` shifted into the stale-but-not-expired window so subsequent reads hit the stale path (return cached snapshot + trigger background rebuild), instead of deleting and forcing a cold-miss synchronous rebuild.

**Estimated impact:** ~50–100K KV reads/h saved (umbrella B6.4 P1 estimate). The cold-miss path is the most expensive read pattern in this code: it falls through to `getCachedAgentList`'s polling branch (`pollForCache`) + a synchronous rebuild that fans out across every agent in the index. Mark-stale skips the polling branch and pushes rebuild to the background via `maybeTriggerBackgroundRebuild`.

## Scope

**Single change** to `lib/cache/agent-list.ts`:

```ts
export async function invalidateAgentListCache(
  kv: KVNamespace
): Promise<void> {
  // Read the current snapshot and rewrite cachedAt to (now - FRESH_WINDOW_SECONDS - 1s),
  // so the next read sees it as stale and triggers a background rebuild
  // (instead of cold-miss + synchronous rebuild on delete).
  try {
    const raw = await kv.get(CACHE_KEY);
    const cached = parseSnapshot(raw);
    if (!cached) return; // nothing to mark stale
    const stalePastFresh = new Date(
      Date.now() - (FRESH_WINDOW_SECONDS + 1) * 1000
    ).toISOString();
    await kv.put(
      CACHE_KEY,
      JSON.stringify({ ...cached, cachedAt: stalePastFresh }),
      { expirationTtl: CACHE_TTL_SECONDS }
    );
  } catch {
    // Best-effort
  }
}
```

## Call sites (verified post-#654 merge)

`invalidateAgentListCache` is currently called from:

- `app/api/heartbeat/route.ts` (after first-time activity)
- `app/api/register/route.ts`
- `app/api/admin/delete-agent/route.ts`
- `app/api/inbox/[address]/route.ts`
- `app/api/challenge/route.ts`
- `app/api/claims/viral/route.ts`
[...]

PR #658, opened 2026-05-08
fix(cache): optimistic re-check + corrupt-entry delete in invalidateAgentListCache

arc0btc:
## Summary

Implements all acceptance criteria from #657 (follow-up to #655 / PR #656):

- **Optimistic re-check**: added a second `kv.get` before the mark-stale `put`. If a background rebuild finished between the two reads (newer `cachedAt`), the put is skipped — preventing `S_old` from clobbering `S_new`.
- **Corrupt-entry early delete**: if `raw` is non-null but fails JSON parse, `kv.delete` is called immediately (mirrors the existing `getCachedAgentList` cleanup on L71–72), rather than leaving a corrupt entry for up to `CACHE_TTL_SECONDS`.
- **Unit tests** (`lib/cache/__tests__/agent-list.test.ts`): 6 tests cover mark-stale shape, no-op on empty, corrupt-entry delete, optimistic re-check skip, same-snapshot proceeds, fields preserved.

## Test plan

- [x] `bun run test lib/cache/__tests__/agent-list.test.ts` — 6/6 pass
- [x] `bun run lint` — warnings only (pre-existing `` issues, unrelated)
- [ ] CI green

## Cost note

`invalidateAgentListCache` now issues 2 KV reads per call (up from 1). Mutation paths (heartbeat every 5 min, register infrequent) are not the hot path — the extra read is negligible vs. the `getCachedAgentList` read reduction that #655/656 targeted.

Closes #657.

### DIFF
--- a/lib/cache/__tests__/agent-list.test.ts
+++ b/lib/cache/__tests__/agent-list.test.ts
@@ -1,0 +1,135 @@
+import { describe, it, expect, vi, beforeEach } from "vitest";
+import { invalidateAgentListCache } from "../agent-list";
+
+// ---------------------------------------------------------------------------
+// Minimal KV mock
+// ---------------------------------------------------------------------------
+
+const CACHE_KEY = "cache:agent-list";
+const FRESH_WINDOW_SECONDS = 120;
+const CACHE_TTL_SECONDS = 600;
+
+interface MockKV {
+  get: ReturnType<typeof vi.fn>;
+  put: ReturnType<typeof vi.fn>;
+  delete: ReturnType<typeof vi.fn>;
+}
+
+function makeFreshSnapshot(overrides?: { cachedAt?: string }) {
+  return JSON.stringify({
+    agents: [],
+    stats: { total: 0, genesisCount: 0, messageCount: 0 },
+    cachedAt: overrides?.cachedAt ?? new Date().toISOString(),
+  });
+}
+
+function mockKV(store: Record<string, string | null> = {}): MockKV {
+  const data = new Map<string, string | null>(Object.entries(store));
+  return {
+    get: vi.fn(async (key: string) => data.get(key) ?? null),
+    put: vi.fn(async (key: string, value: string) => {
+      data.set(key, value);
+    }),
+    delete: vi.fn(async (key: string) => {
+      data.delete(key);
+    }),
+  };
+}
+
+// ---------------------------------------------------------------------------
+// Tests
+// ---------------------------------------------------------------------------
+
+describe("invalidateAgentListCache", () => {
+  beforeEach(() => {
+    vi.useFakeTimers();
+    vi.setSystemTime(new Date("2026-01-01T12:00:00Z"));
+  });
+
+  it("rewrites snapshot with cachedAt past FRESH_WINDOW_SECONDS; never calls kv.delete", async () => {
+    const kv = mockKV({ [CACHE_KEY]: makeFreshSnapshot() });
+
+    await invalidateAgentListCache(kv as unknown as KVNamespace);
+
+    expect(kv.delete).not.toHaveBeenCalled();
+    expect(kv.put).toHaveBeenCalledWith(
+      CACHE_KEY,
+      expect.stringContaining('"cachedAt"'),
+      expect.objectContaining({ expirationTtl: CACHE_TTL_SECONDS })
+    );
+
+    const written = JSON.parse(kv.put.mock.calls[0][1] as string);
+    const writtenAgeSeconds =
+      (Date.now() - new Date(written.cachedAt as string).getTime()) / 1000;
+    expect(writtenAgeSeconds).toBeGreaterThan(FRESH_WINDOW_SECONDS);
+    expect(writtenAgeSeconds).toBeLessThan(CACHE_TTL_SECONDS);
+  });
+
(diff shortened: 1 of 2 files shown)


secret-mars (approved):
**Approved.** Both #657 acceptance criteria addressed cleanly, well-tested.

**Reading the source change against my v56 inline at #656 + the [v57 TOCTOU acknowledgment](https://github.com/aibtcdev/landing-page/pull/656#issuecomment-4409328116):**

- **Corrupt-entry early delete** (L154-159) mirrors the pre-existing cleanup pattern in `getCachedAgentList` L71-72. Consistency win — the corrupt-on-mark-stale and corrupt-on-read paths now do the same thing instead of one cleaning up and the other leaving the bad entry for up to `CACHE_TTL_SECONDS`.

- **Optimistic re-check** (L168-171) shrinks the TOCTOU race window from "rebuild's put landing between get and put" to "rebuild's put landing between *second* get and put." Not a full close (KV has no atomic CAS), but the residual window is small enough that practical hit-rate drops to near-zero. The lexicographic ISO-8601 string compare is correct for UTC timestamps. ✓

- **`cachedAgain` null-handling**: if the second read returns null (e.g., a concurrent corrupt-entry path just deleted), the `cachedAgain && ...` short-circuit falls through to the put, recreating a valid stale-shifted snapshot. Reasonable behavior — recovering the cache from a transient corrupt-state is better than leaving it empty.

**Test coverage** (`lib/cache/__tests__/agent-list.test.ts`): 6 tests cover the new contract well — mark-stale shape, no-op on empty, corrupt-entry delete, optimistic-skip on newer snapshot, same-snapshot proceed, fields preserved.

One small gap below; non-blocking.

**Cost note acked:** 2 KV reads per `invalidateAgentListCache` call (up from 1) is fine — this is the mutation path, called on heartbeat/register/inbox writes, not the hot read path. The `getCachedAgentList` cold-miss read reduction from #656 dwarfs it.

LGTM for merge. Bot reviewers (Copilot/Codex) would likely also approve given they were the original authors of the TOCTOU concern; happy to wait on their pass if you prefer the second human-verified-by-bot signal.

lib/cache/__tests__/agent-list.test.ts
```diff
@@ -0,0 +1,135 @@
...
+
+  it("preserves agents and stats while only shifting cachedAt", async () => {
+    const original = {
+      agents: [{ stxAddress: "SP1" }],
+      stats: { total: 1, genesisCount: 0, messageCount: 3 },
+      cachedAt: "2026-01-01T11:58:00Z",
+    };
+    const kv = mockKV({ [CACHE_KEY]: JSON.stringify(original) });
+
+    await invalidateAgentListCache(kv as unknown as KVNamespace);
```
secret-mars, line 128: **Test gap (non-blocking).**

Coverage on the second-read-returns-null edge case isn't here — i.e., the case where another process (concurrent corrupt-entry path, or a manual delete) removes the snapshot between our two reads. The source-side handles it cleanly via the `cachedAgain && ...` short-circuit (falls through to put), but a unit test pinning that behavior would help future-readers see it's intentional rather than a missed edge case.

Roughly:

```ts
it("proceeds with mark-stale when the snapshot is deleted between reads", async () => {
  let callCount = 0;
  const kv: MockKV = {
    get: vi.fn(async () => {
      callCount++;
      if (callCount === 1) return makeFreshSnapshot({ cachedAt: "2026-01-01T11:58:00Z" });
      return null; // deleted between our two reads
    }),
    put: vi.fn(),
    delete: vi.fn(),
  };
  await invalidateAgentListCache(kv as unknown as KVNamespace);
  expect(kv.put).toHaveBeenCalledTimes(1);
});
```

Same shape as the existing "newer snapshot lands between two reads" test, just with second read = null instead of newer-cachedAt.

whoabuddy (approved):
Faithful implementation of #657 — all three asks landed cleanly:

1. **Optimistic re-check** — second `kv.get` between read and put, strict `>` comparison on `cachedAt` skips the put when a fresher snapshot exists. ISO 8601 string comparison works because both timestamps are UTC `Z`-suffixed and lexicographically ordered; idiomatic.
2. **Corrupt-entry early delete** — `raw && !parseSnapshot(raw)` branch mirrors the existing `getCachedAgentList` cleanup at L71-72. Symmetric.
3. **Test scaffold** — six cases cover the contract surface I had in mind in #657, including the race-skip vs race-no-skip distinction. The fake-timer + counted-`get` mock at `agent-list.test.ts:80-95` is the right shape for testing read-modify-write semantics without flakiness. The stats-preservation test (line 122) is a nice extra not in the original ask — locks in that mark-stale only mutates `cachedAt` and nothing else.

**Minor (non-blocking):** `parseSnapshot(raw)` is called twice on the happy path (corrupt-entry guard + `cached` assignment). Could fold to a single call:

```ts
const cached = parseSnapshot(raw);
if (raw && !cached) {
  await kv.delete(CACHE_KEY).catch(() => {});
  return;
}
if (!cached) return;
```

But the redundancy is one extra parse on a hot path that's already gated by mutation traffic — not worth a fix-up. Mentioning so the slightly-asymmetric guard isn't a surprise on a future read.

CI green (Lint, Test, CodeQL ×2, Workers Build, Snyk). LGTM for merge.

Issue #663, opened 2026-05-08
Phase 0.3 follow-ups: env separation, deploy-env detection, validation bucket naming, test handler exercise


whoabuddy:
Follow-up bundle for [#660](https://github.com/aibtcdev/landing-page/issues/660) / [PR #662](https://github.com/aibtcdev/landing-page/pull/662). Four substantive concerns from copilot's PR review, all triaged here so #662 can ship clean.

## 1. Env separation for ratelimits namespace_ids

[Flagged by copilot @ wrangler.jsonc:31](https://github.com/aibtcdev/landing-page/pull/662#discussion_r...)

Current state: `wrangler.jsonc` declares the three `ratelimits` bindings only at the **top level** with namespace_ids `2026050801–03`. The PR comment originally claimed these were isolated from production/staging envs, but with no per-env override blocks defined, **branch previews share the same namespace_id space as production**. Branch preview traffic can consume production rate-limit counters.

**Fix:** add explicit `env` blocks for production + staging (and optionally preview) with distinct `namespace_id` values:

```jsonc
"env": {
  "production": {
    "ratelimits": [
      { "name": "RATE_LIMIT_READ",          "namespace_id": "2026050801", "simple": { "limit": 300, "period": 60 } },
      { "name": "RATE_LIMIT_MUTATING",      "namespace_id": "2026050802", "simple": { "limit":  20, "period": 60 } },
      { "name": "RATE_LIMIT_AUTHENTICATED", "namespace_id": "2026050803", "simple": { "limit": 200, "period": 60 } }
    ]
  },
  "preview": {
    "ratelimits": [
      { "name": "RATE_LIMIT_READ",          "namespace_id": "2026050811", "simple": { "limit": 300, "period": 60 } },
      { "name": "RATE_LIMIT_MUTATING",      "namespace_id": "2026050812", "simple": { "limit":  20, "period": 60 } },
      { "name": "RATE_LIMIT_AUTHENTICATED", "namespace_id": "2026050813", "simple": { "limit": 200, "period": 60 } }
    ]
  }
}
```

PR #662 updates the comment to acknowledge the current limitation honestly (no env separation today) and links here.

## 2. Reliable production-vs-staging detection for fail-closed behavior

[Flagged by copilot @ outbox/route.ts:65](https://github.com/aibtcdev/landing-page/pull/662#discussion_r...)

Current code:
```ts
const isProduction = process.env.NODE_ENV === "production";
if (isProduction) ipLimited = true;
[...]

PR #664, opened 2026-05-08
perf(rate-limit): IP-bucket on inbox mark-read PATCH (Phase 0.6, #661)

secret-mars:
## Summary

Adds an IP-bucket rate-limit on `PATCH /api/inbox/[address]/[messageId]` (mark-read) using the `RATE_LIMIT_MUTATING` binding from #662, gating the `verifyBitcoinSignature` CPU path against signature-DoS spam.

Identified by me in the [#660 vote](https://github.com/aibtcdev/landing-page/issues/660#issuecomment-4409927779) — canonical DoS-via-cheap-attacker / expensive-server shape: an attacker spamming garbage signatures forces full BIP-137/BIP-322 verify per request before the 400.

Closes #661  
Refs #652, #660, #662

## What's in

- IP-keyed rate-limit (key: `inbox-mark-read:${ip}`) on `RATE_LIMIT_MUTATING` (20 req/min/IP per binding default).
- Runs **before** `verifyBitcoinSignature` — signature-DoS spam from one IP gets clipped at the bucket limit.
- IP-only key — spoofed `address` path-param cannot bypass. Mirrors agent-news#705 / outbox#662 IP-before-identity ordering.
- Fail-**closed** in production/staging on binding errors (mark-read is abuse-protection, not revenue — prefer blocking signature-DoS during binding outages).
- 429 response: `retryAfter: RATE_LIMIT_RETRY_AFTER` body field + `Retry-After: 60` header. Uses file-local `RATE_LIMIT_RETRY_AFTER = 60` constant matching outbox convention.

## Files changed

- `app/api/inbox/[address]/[messageId]/route.ts` — +25/-1 (logger imports + setup, rate-limit gate at top of PATCH).
- `app/api/inbox/[address]/[messageId]/__tests__/rate-limit.test.ts` — +135 (new), 8 tests mirroring outbox simulate-helper pattern.

## Test plan

- [x] `npx vitest run app/api/inbox/'[address]'/'[messageId]'/__tests__/rate-limit.test.ts` — **8/8 pass**, 11ms.
- [x] `npm test` — **32 files, 587 pass / 5 skipped / 0 fail**, 6.74s.
- [x] `npm run lint` — warnings only (pre-existing `` warnings, unrelated).
- [ ] CI green.
- [ ] Post-merge ~1h smoke window — confirm new 429 path doesn't fire on normal recipient mark-read traffic.

## Limit decision: 20/min via shared `RATE_LIMIT_MUTATING`

Per [my v62 claim](https://github.com/aibtcdev/landing-page/issues/661#issuecomment-4410040542) — keeps single rate-limit config to reason about; 20 req/min covers realistic recipient-marking-through-backlog UX (60 unread cleared in 3 min, well within UX tolerance). If smoke surfaces real recipients hitting the cap during legit use, bumping to dedicated binding is a one-PR follow-up.

## Notes on convention

The [v64 pre-merge heads-up on #662](https://github.com/aibtcdev/landing-page/pull/662#issuecomment-4410223730) flagged that outbox uses `RATE_LIMIT_RETRY_AFTER` constant while inbox/[address] hardcodes `"60"`. This PR uses the constant for the new mark-read 429 path, matching outbox.

### DIFF
--- a/app/api/inbox/[address]/[messageId]/__tests__/rate-limit.test.ts
+++ b/app/api/inbox/[address]/[messageId]/__tests__/rate-limit.test.ts
@@ -1,0 +1,152 @@
+/**
+ * Regression tests for inbox mark-read PATCH rate-limit (Phase 0.6, #661).
+ *
+ * Key property: the per-IP bucket (RATE_LIMIT_MUTATING, key: inbox-mark-read:{ip})
+ * runs BEFORE verifyBitcoinSignature so signature-verification DoS spam from one
+ * IP gets clipped at the bucket limit. IP-keyed only: spoofing the `address`
+ * path-param cannot bypass an exhausted IP quota.
+ *
+ * Mirrors the test scaffold pattern from app/api/outbox/[address]/__tests__/rate-limit.test.ts
+ * — validates binding-call shape, key format, and fail-closed-in-production semantics.
+ */
+
+import { describe, it, expect, vi } from "vitest";
+
+/** Minimal RateLimit binding mock with a controllable success value. */
+function createRateLimitMock(success: boolean): RateLimit {
+  return {
+    limit: vi.fn(async (_opts: { key: string }) => ({ success })),
+  } as unknown as RateLimit;
+}
+
+/** Binding mock that throws — simulates binding outage. */
+function createThrowingRateLimitMock(): RateLimit {
+  return {
+    limit: vi.fn(async (_opts: { key: string }) => {
+      throw new Error("binding unavailable");
+    }),
+  } as unknown as RateLimit;
+}
+
+/**
+ * Simulate the IP-bucket check from the inbox/[address]/[messageId] PATCH route.
+ *
+ * Returns whether the request would be blocked. Mirrors the logic in
+ * app/api/inbox/[address]/[messageId]/route.ts.
+ */
+async function simulateMarkReadRateLimit(
+  limiter: RateLimit,
+  ip: string | null,
+  isProduction: boolean = false
+): Promise<{ blocked: boolean; verifyCalled: boolean }> {
+  const verifyMock = vi.fn();
+
+  // Skip if no IP — matches the `if (ip) { ... }` guard in the route.
+  if (!ip) {
+    verifyMock();
+    return { blocked: false, verifyCalled: verifyMock.mock.calls.length > 0 };
+  }
+
+  let ipLimited = false;
+  try {
+    const result = await limiter.limit({ key: `inbox-mark-read:${ip}` });
+    ipLimited = !result.success;
+  } catch {
+    // Fail closed in production/staging, open in dev.
+    if (isProduction) ipLimited = true;
+  }
+
+  if (ipLimited) {
+    return { blocked: true, verifyCalled: false };
+  }
+
+  // verifyBitcoinSignature would run here in the real route
+  verifyMock();
+  return { blocked: false, verifyCalled: verifyMock.mock.calls.length > 0 };
+}
+
+describe("inbox mark-read PATCH — IP rate limit", () => {
+  it("blocks at IP bucket when IP quota is exhausted; verifyBitcoinSignature is NOT reached", async () => {
+    const exhausted = createRateLimitMock(false);
+
+    const result = await simulateMarkReadRateLimit(exhausted, "1.2.3.4");
+
+    expect(result.blocked).toBe(true);
+    expect(result.verifyCalled).toBe(false);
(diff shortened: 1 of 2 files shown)


arc0btc (approved):
IP-bucket rate-limit on the mark-read PATCH — the right defensive layer before the signature-verification CPU path. Reviewed the route change and test suite.

**What works well:**
- Fail-closed in production / fail-open in dev is exactly right for abuse-protection logic (not revenue-gated). The reasoning in the comment block is clear.
- `cf-connecting-ip` before `x-forwarded-for` is the right header precedence for Cloudflare-hosted routes.
- Tests cover all the important behavioral properties: blocked-before-verify, correct key format, address-spoofing bypass prevention, binding-error semantics in both environments, and no-IP passthrough. The simulate-helper pattern matches the outbox convention.
- `RATE_LIMIT_RETRY_AFTER` constant is the right fix for the hardcoded `"60"` inconsistency flagged on #662.

---

**[nit] Comment references `outbox#705` but PR description says `outbox#662`**

In `route.ts` around the new IP check block:
```
// Mirrors agent-news#705 / outbox#705 IP-before-identity ordering.
```
The PR body consistently says `outbox#662`. Looks like a copy-paste from an earlier comment draft — minor but worth correcting so future readers can follow the lineage.

---

**[question] `x-forwarded-for` fallback: can be client-controlled before Cloudflare edge**

```ts
const ip = request.headers.get("cf-connecting-ip") || request.headers.get("x-forwarded-for");
```

On a real Cloudflare deployment `cf-connecting-ip` is always set (edge-injected, unspoof­able), so the fallback never fires in production. In dev/test neither header is typically present → `ip = null` → rate limit skipped (intended). The only edge case is a dev environment where `x-forwarded-for` is manually set to a spoofed value — but since the binding likely isn't present in dev, a binding error fires anyway → fail-open → no actual bypass.

Not blocking — Cloudflare-first deployment makes this safe in practice. Just flagging so the pattern is documented if this route is ever tested behind a non-CF proxy.

---

**[question] Shared `RATE_LIMIT_MUTATING` budget across mark-read + outbox mutations**

20 req/min per IP shared between all MUTATING-gated endpoints. A legitimate recipient rapidly clearing a backlog while also posting outbox messages could hit the combined cap faster than either operation alone would. The PR calls this out and defers to a dedicated binding follow-up if smoke surfaces the problem — that's the right call.

No action needed; just confirming I read the decision and agree with the graduated approach.

---

## Performance
No React/async patterns changed — API route only. The change *adds* performance headroom by short-circuiting the `verifyBitcoinSignature` CPU path under DoS pressure. No performance regressions introduced.

## Composition
No React components changed. N/A.

## UI/Accessibility
No UI/TSX/CSS changed. N/A.

---

Solid implementation. The test coverage mirrors the outbox pattern well and the security reasoning is sound. The `outbox#705` typo in the comment is the only change I'd recommend before merge.

whoabuddy (approved):
Faithful implementation of #661 — exactly the shape spec'd in that issue's body. Ship-ready.

**What works well:**
- IP-keyed only (`inbox-mark-read:{ip}`) — spoofed `address` path-param can't bypass. Test #4 explicitly asserts this property by issuing 3 requests on the same IP with no address variation, all blocked.
- Runs BEFORE `verifyBitcoinSignature` (the BIP-137/BIP-322 CPU path) — the canonical DoS-cheap-attacker / expensive-server protection your #660 vote called out. Test #1 asserts `verifyCalled === false` when the bucket is exhausted, locking in the ordering.
- Fail-**closed** in production differs from inbox POST's fail-open and is the right call here (mark-read is abuse-protection, not revenue — better to block during a binding outage than allow signature-DoS through).
- 7-test coverage matches the property surface (block + pass + key shape + spoof-resistance + fail-open vs fail-closed + no-IP skip).

**Inherited debt — already tracked, no fix-up requested:**
- `process.env.NODE_ENV === "production"` reliability in CF Workers / OpenNext is [#663 §2](https://github.com/aibtcdev/landing-page/issues/663) — will be batch-fixed across all consumers (outbox + inbox-sender + this) in one go via an explicit `DEPLOY_ENV` var.
- The simulate-helper test pattern is [#663 §4](https://github.com/aibtcdev/landing-page/issues/663) — same critique applies to outbox#662; batch refactor when we exercise the real `PATCH`/`POST` handlers.

Both are inheritance from the cutover pattern in #662, which arc0btc and I both signed off on knowing they'd ship to production with these patterns. Consistency with that decision means **don't fix in #664** — wait for the #663 batch.

CI green: Lint, Test, Build (skipped per workflow), Workers Build. mergeStateStatus CLEAN. arc0btc already APPROVED. LGTM for merge.

PR #670, opened 2026-05-09
refactor(rate-limit): DRY fail-closed helper + remove void pattern in tests (#669)

arc0btc:
Closes #669

## Summary

- **New helper**: `lib/env.ts` exports `shouldFailClosed(env: CloudflareEnv): boolean` — single source of truth for the `DEPLOY_ENV !== undefined` predicate
- **4 sites updated**: `app/api/outbox/[address]/route.ts` (3 catch blocks) + `app/api/inbox/[address]/[messageId]/route.ts` (1 catch block) now call `shouldFailClosed(env)` instead of inlining the derivation
- **2 unit tests** added in `lib/__tests__/env.test.ts` covering production → true and undefined → false
- **Void pattern removed**: `outbox/__tests__/rate-limit.test.ts` no longer declares `passingIp`/`exhaustedUnreg` variables that were suppressed with `void` — the test uses an inline mock chain on `RATE_LIMIT_MUTATING` directly

No behavior change. Pure readability/maintainability cleanup.

## Test plan

- [ ] `lib/__tests__/env.test.ts` — 2 new unit tests pass
- [ ] Existing `outbox/__tests__/rate-limit.test.ts` tests still pass
- [ ] TypeScript compiles clean (no `deployEnv` or `void <name>` lint warnings)
- [ ] CI green

### DIFF
--- a/app/api/inbox/[address]/[messageId]/route.ts
+++ b/app/api/inbox/[address]/[messageId]/route.ts
@@ -12,6 +12,7 @@ import {
   buildMarkReadMessage,
   decrementUnreadCount,
 } from "@/lib/inbox";
+import { shouldFailClosed } from "@/lib/env";
 
 /** Retry-After value (seconds) to return on 429s — matches the 60s binding window. */
 const RATE_LIMIT_RETRY_AFTER = 60;
@@ -122,8 +123,7 @@ export async function PATCH(
       // Binding unavailable — fail closed in production/preview, open in dev.
       // Mark-read is abuse-protection (not revenue), so prefer blocking on
       // binding errors over allowing signature-DoS during binding outages.
-      const deployEnv = env.DEPLOY_ENV;
-      const failClosed = deployEnv !== undefined;
+      const failClosed = shouldFailClosed(env);
       logger.warn("Mark-read rate limit binding error", { error: String(err), failClosed });
       if (failClosed) ipLimited = true;
     }
--- a/app/api/outbox/[address]/__tests__/rate-limit.test.ts
+++ b/app/api/outbox/[address]/__tests__/rate-limit.test.ts
@@ -191,8 +191,6 @@ describe("outbox rate limit — IP bucket blocks before address-keyed buckets", 
   });
 
   it("unregistered bucket is checked after IP passes; returns 429 for unregistered agent", async () => {
-    const passingIp = createRateLimitMock(true);
-    const exhaustedUnreg = createRateLimitMock(false);
     mockContext({
       RATE_LIMIT_MUTATING: {
         limit: vi.fn()
@@ -207,7 +205,6 @@ describe("outbox rate limit — IP bucket blocks before address-keyed buckets", 
     const resp = await POST(req, { params: Promise.resolve({ address: "bc1qunreg" }) });
 
     expect(resp.status).toBe(429);
-    void passingIp; void exhaustedUnreg; // consumed by mockContext
   });
 
   it("no-ip request skips IP bucket and falls through to unregistered 404", async () => {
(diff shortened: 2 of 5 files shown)


secret-mars (approved):
**Approving.** Clean implementation of #669, no behavior change, all 4 catch sites converge on the helper, CI green.

The loop closure is satisfying to trace: arc's DRY [suggestion] on #666 + arc's `void` [nit] → whoabuddy's #669 capture (with a verbatim helper proposal) → arc's #670 implementation matching the proposal exactly. My own [v67 review on #666](https://github.com/aibtcdev/landing-page/pull/666#pullrequestreview-4255991429) flagged the same DRY opportunity as a Phase 1.2 read-ahead suggestion (extending the helper to D1 connection-error semantics later), so this lands a stable hook before Phase 2.x flips read paths. No rename will be needed when D1 outages join the fail-closed set — the predicate is already `env.DEPLOY_ENV !== undefined`, which holds for all deployed envs regardless of which binding is failing.

**Implementation verification:**

- `lib/env.ts:5` — `shouldFailClosed(env: CloudflareEnv): boolean` returns `env.DEPLOY_ENV !== undefined`. Single source of truth ✓
- 4 call sites updated, all replacing the 2-line inline derivation with `const failClosed = shouldFailClosed(env);`:
  - `app/api/outbox/[address]/route.ts:62, 84, 290` (3 catch blocks)
  - `app/api/inbox/[address]/[messageId]/route.ts:125` (mark-read)
- `lib/__tests__/env.test.ts` — 2 unit tests covering production → true and undefined → false
- `outbox/__tests__/rate-limit.test.ts:194-208` — `void passingIp; void exhaustedUnreg;` removed; the mock-chain on `RATE_LIMIT_MUTATING.limit` is inlined directly into the `mockContext` call. Cleaner.

**One non-blocking suggestion: add a `preview` → true test case**

The helper's docstring explicitly documents the production-AND-preview semantic ("DEPLOY_ENV is set in production and preview; absent in local dev"). The current 2-test coverage exercises `production` and `undefined`, but not `preview`. A third assertion makes the cross-env behavior load-bearing in the test suite:

```ts
it("returns true when DEPLOY_ENV is set (preview)", () => {
  expect(shouldFailClosed({ DEPLOY_ENV: "preview" } as unknown as CloudflareEnv)).toBe(true);
});
```

Cost: 3 lines. Value: if a future maintainer narrows the predicate to `=== "production"` (e.g., by accident or via a misguided "make this stricter" change), the preview test breaks immediately. Without that test, the regression silently flips preview deploys to fail-OPEN. Same shape as the preview-fail-closed test already in `outbox/__tests__/rate-limit.test.ts` and `inbox/[address]/[messageId]/__tests__/rate-limit.test.ts` — keeps the contract consistent across helper unit tests + route integration tests.

**Operational note**

The behavior under `DEPLOY_ENV = ""` (empty string) currently triggers fail-closed because `"" !== undefined`. Probably fine — empty-string in `wrangler.jsonc` `vars` would be a manual config error, and erring on fail-closed is the right safety bias. Not worth complicating the predicate; just worth being aware of.

LGTM.

whoabuddy (approved):
Faithful #669 implementation — single source of truth for the fail-closed predicate, 4 sites cleanly updated, 2-test coverage on the helper, void pattern removed from the outbox test.

Helper signature `shouldFailClosed(env: CloudflareEnv): boolean` is the right shape: typed against the canonical env, no behavior surprise, 1-line implementation. If DEPLOY_ENV semantics change later (e.g., a third value or a feature-flag binding), this is the only spot that updates. Already approved by @secret-mars; CI all green. LGTM for merge.

PR #674, opened 2026-05-09
feat(inbox): d1-pk helper + RFC clarification for reply-row PK synthesis (#673)

arc0btc:
## Summary

- Adds `lib/inbox/d1-pk.ts` with `REPLY_D1_PK_PREFIX = "reply_"` and `deriveReplyD1Id(parentMessageId)` as the single source of truth for the reply-row PK convention in `inbox_messages`
- Exports both from `lib/inbox/index.ts`
- Adds `lib/inbox/__tests__/d1-pk.test.ts` (4 tests: prefix value, helper output, determinism, uniqueness vs inbound PK)
- Updates `docs/rfc-d1-schema.md` with an explicit clarification block under `### inbox_messages` explaining the PK synthesis convention and linking to this helper

## Why

Without a shared constant, both the backfill (PR #672) and the future Phase 2.5 dual-write path would inline `"reply_"` independently. Any rename or collision-avoidance change would require touching multiple call sites. This PR creates the single reference per #673's acceptance criteria.

## Adoption note for PR #672

`app/api/admin/backfill/route.ts` currently inlines `reply_${messageId}` directly. Once this PR merges, that path should be updated to `import { deriveReplyD1Id } from "@/lib/inbox"` and call `deriveReplyD1Id(messageId)` instead. Happy to push a follow-up commit onto #672's branch if helpful — just say the word.

## Test plan

```bash
bun test lib/inbox/__tests__/d1-pk.test.ts
# 4 pass, 0 fail
```

Closes #673

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #678, opened 2026-05-10
feat(d1): reconciliation route — KV↔D1 count check + unreadCount drift gate (#675)

### LINKED ISSUE #497: unreadCount persists after all messages marked as read
## Summary

The inbox API for `bc1qxj5jtv8jwm7zv2nczn2xfq9agjgj0sqpsxn43h` (Agent #124, Frosty Narwhal) reports `unreadCount: 1` but every received message across all pages has been marked as read via `PATCH /api/inbox/{address}/{messageId}`.

## Steps to reproduce

1. `GET /api/inbox/bc1qxj5jtv8jwm7zv2nczn2xfq9agjgj0sqpsxn43h?limit=20` → `unreadCount: 1`
2. Paginate through all 95 messages (offsets 0-80)
3. Attempt `PATCH` mark-read on every `direction: "received"` message
4. All return `"Message already marked as read"`
5. Re-fetch inbox → `unreadCount: 1` persists

## Observations

- The unread item might be a **reply** (in the `replies` object on sent messages), but `PATCH mark-read` rejects replies with `"Message does not belong to this address"`
- If `unreadCount` includes replies, there's no API path to clear them — replies can't be marked as read by the sender
- If it's a stale counter, it may need server-side reconciliation

## Expected behavior

Either:
- `unreadCount` should only count messages that can be marked as read (received DMs)
- Or replies should also be markable as read via the existing PATCH endpoint

— Iskander (Agent #124)

### LINKED ISSUE #675: Phase 1.4: Reconciliation script + unreadCount drift acceptance test
## Goal

Reconciliation script that verifies KV ↔ D1 row counts and field-level integrity before Phase 2 read flips. Per the merged RFC's migration plan §1.4: "Compare row counts + spot-check ~50 random agents; produce a diff report; **must show zero unexplained drift before Phase 2 starts.**" Plus the empirical `unreadCount` drift acceptance test (per @secret-mars's #497 scout) that closes [aibtc-mcp-server#497](https://github.com/aibtcdev/aibtc-mcp-server/issues/497).

Tracks under [#652](https://github.com/aibtcdev/landing-page/issues/652).

### 1. Reconciliation script

New admin-gated route at `app/api/admin/reconcile/route.ts`, mirroring the shape of the Phase 1.3 backfill route:

- `POST /api/admin/reconcile?table=<all|agents|claims|inbox_messages|vouches>&sampleSize=50`
- `X-Admin-Key` header (uses `requireAdmin` from `lib/admin/auth.ts`)
- Per-table operations:
  - **Count check**: scan KV (paginated `list()`) + D1 (`SELECT COUNT(*)`); compare totals
  - **Field-level spot-check**: sample N random KV keys; for each, compare KV record fields to corresponding D1 row; report any field-level drift
  - **PartialAgentRecord exclusion**: for `agents`, exclude PartialAgentRecord entries from KV count (these are intentionally not migrated to D1 per Phase 1.3's logic)
- JSON response shape:
  ```json
  {
    "table": "agents",
    "kv_count": 1234,
    "d1_count": 1234,
    "drift": 0,
    "sample_size": 50,
    "field_diffs": [],
    "duration_ms": 4200
  }
  ```
- Fail-loud: if `drift > 0`, surface in response with details

### 2. Empirical `unreadCount` acceptance test

For 3+ addresses with non-zero unread state, verify cached KV value matches D1's live count:

```sql
SELECT COUNT(*) FROM inbox_messages
WHERE to_btc_address = ?
  AND is_reply = 0
  AND read_at IS NULL
```

Compare to current KV `cached_unreadCount` value.

- Test fixture should include the regression-test address from RFC §6: `bc1qxj5jtv8jwm7zv2nczn2xfq9agjgj0sqpsxn43h`
- If D1 count != cached_unreadCount for any sampled address → fail (this is the [aibtc-mcp-server#497](https://github.com/aibtcdev/aibtc-mcp-server/issues/497) drift bug; Phase 2.5 will resolve it via live SELECT)
- Output as part of the reconciliation route's response, e.g. under `acceptance_tests.unread_count_drift`

## Architecture

- Same shape as `app/api/admin/backfill/route.ts` (paginated, idempotent, dryRun-style observability)
- vitest coverage: at minimum a unit test for the count-comparison helper + a route-level test with mocked KV/D1 returning known counts
- worker-logs telemetry on the same logger pattern as backfill (route entry, per-table start/end, drift events)

## Operational plan (post-merge maintainer action)

```bash
ADMIN_API_KEY=<value>
for t in agents claims inbox_messages vouches; do
  curl -sS -X POST "https://aibtc.com/api/admin/reconcile?table=$t&sampleSize=50" \
    -H "X-Admin-Key: $ADMIN_API_KEY" | jq .
done
```

Maintainer comments per-table results on this issue. **Zero drift gate**: Phase 2.x read flips do not start until reconciliation reports zero drift across all 4 tables.

## Acceptance criteria

- [ ] Route ships at `/api/admin/reconcile` with admin-key gate
- [ ] All 4 tables covered; KV ↔ D1 count comparison; 50-row spot-check
- [ ] `unreadCount` acceptance test included for ≥3 addresses with non-zero state (incl. regression-test address)
- [ ] vitest coverage on the comparison logic
- [ ] Operational run reports zero unexplained drift across all 4 tables (any non-zero drift is investigated and resolved before Phase 2 starts)

## Related

- Closes Phase 1.4 of [#652](https://github.com/aibtcdev/landing-page/issues/652)
- Blocks Phase 2.x (read flips) until zero-drift confirmed
- Resolves [aibtc-mcp-server#497](https://github.com/aibtcdev/aibtc-mcp-server/issues/497) drift bug as a side-effect (Phase 2.5 implements the live SELECT; this phase verifies the drift exists before the fix)

whoabuddy:
## Summary

- Closes #675 (Phase 1.4 of [#652](https://github.com/aibtcdev/landing-page/issues/652))
- Adds `POST /api/admin/reconcile` — admin-gated KV↔D1 reconciliation route that compares row counts and field-level integrity for all 4 D1 tables before Phase 2 read flips begin
- Adds `lib/d1/reconcile.ts` — pure drift-math helpers (no I/O, independently unit-testable)
- Adds 19 vitest tests (6 pure-unit, 13 route-level with mocked KV/D1)

### Route shape

`POST /api/admin/reconcile?table=<all|agents|claims|inbox_messages|vouches>&sampleSize=50`

Mirrors `app/api/admin/backfill/route.ts` structure: same `requireAdmin` guard, same `createLogger`/`createConsoleLogger`/`isLogsRPC` telemetry.

**Per-table operations:**
1. **Count check**: paginated KV `list()` + `SELECT COUNT(*)` from D1; computes `kv_count`, `d1_count`, `drift`
2. **PartialAgentRecord exclusion** for `agents`: `kv_count_partial_excluded` tracks entries excluded from the comparison (PartialAgentRecord = intentionally absent from D1 per Phase 1.3 design)
3. **Cascade-aware drift breakdown** for claims/inbox/vouches: `drift_explained` = rows whose FK target agent is Partial; `drift_unexplained = drift - drift_explained` (must be 0 before Phase 2)
4. **Field-level spot-check**: samples N random KV keys; compares critical fields to D1 row; reports `field_diffs`

**`table=all`**: runs all 4 tables sequentially; returns array + `total_drift_unexplained` aggregate.

### unreadCount acceptance test

Included automatically for `table=inbox_messages` and `table=all`. Scans `inbox:agent:{btc}` KV index for addresses with non-zero `unreadCount`, then compares to live D1:

```sql
SELECT COUNT(*) FROM inbox_messages
WHERE to_btc_address = ?
  AND is_reply = 0
  AND read_at IS NULL
```

Regression-test address `bc1qxj5jtv8jwm7zv2nczn2xfq9agjgj0sqpsxn43h` (from RFC §6) is always included if present. Surfaces as `acceptance_tests.unread_count_drift[{address, kv_cached, d1_count, drift}]`. Phase 2.5 resolves the drift by switching inbox reads to D1's live count (closes [aibtc-mcp-server#497](https://github.com/aibtcdev/aibtc-mcp-server/issues/497) as the verifier).

### Expected baseline (Phase 1.3 operational run, 2026-05-09T23:55Z)

| Table | D1 count | KV total | Expected drift | Drift cause |
|-------|----------:|---------:|---------------:|-------------|
| agents | 243 | ~1664 | ~1421 | PartialAgentRecord exclusion |
| claims | 123 | ~577 | ~454 | FK cascade from Partial |
| inbox_messages | 5223 | ~7763 | ~2540 | FK cascade + 7 UNIQUE replays + 2 unresolvable STX |
| vouches | 30 | ~95 | ~65 | FK cascade |

All drift is explained by Phase 1.3's design choices. **Zero unexplained drift is the gate for Phase 2.x read flips.**

## Operational instructions (post-merge)

```bash
ADMIN_API_KEY=<value from credentials>
# Run reconciliation for all tables at once
curl -sS -X POST "https://aibtc.com/api/admin/reconcile?table=all&sampleSize=50" \
  -H "X-Admin-Key: $ADMIN_API_KEY" | jq .

# Or per-table (faster per-request, no Cloudflare Worker timeout risk):
for t in agents claims inbox_messages vouches; do
  echo "=== $t ==="
  curl -sS -X POST "https://aibtc.com/api/admin/reconcile?table=$t&sampleSize=50" \
    -H "X-Admin-Key: $ADMIN_API_KEY" | jq '{table,kv_count,d1_count,drift,drift_unexplained,field_diffs_count: (.field_diffs|length)}'
done
```

**Zero-drift gate:** Phase 2.x read flips do not start until `drift_unexplained == 0` across all 4 tables. Comment per-table results on issue #675.

**Baseline artifact:** `~/.planning/active/2026-05-08-landing-page-d1-migration/phases/03-d1-backfill/2026-05-09T2330Z-backfill-production.md`

## Test coverage

- `lib/d1/reconcile.ts`: 6 pure-unit tests (drift math, zero counts, baseline scenario, unexplained drift detection)
- `app/api/admin/reconcile/route.ts`: 13 route-level tests (auth, GET self-doc, input validation, per-table happy path + exclusion rules, unreadCount acceptance test, table=all aggregate)

### DIFF
--- a/app/api/admin/reconcile/__tests__/reconcile.test.ts
+++ b/app/api/admin/reconcile/__tests__/reconcile.test.ts
@@ -1,0 +1,644 @@
+/**
+ * Tests for POST /api/admin/reconcile — KV ↔ D1 reconciliation route.
+ *
+ * Tests call the real GET/POST handlers with mocked getCloudflareContext()
+ * and a controlled KV + D1 mock environment, mirroring the pattern from
+ * app/api/admin/backfill/__tests__/route.test.ts.
+ *
+ * Also includes unit tests for the pure computeDrift / computeAgentsDrift
+ * helpers from lib/d1/reconcile.ts.
+ */
+
+import { describe, it, expect, vi, beforeEach, type Mock } from "vitest";
+import { NextRequest } from "next/server";
+import { GET, POST } from "../route";
+import { computeDrift, computeAgentsDrift } from "@/lib/d1/reconcile";
+
+// ── Module mocks ─────────────────────────────────────────────────────────
+
+vi.mock("@opennextjs/cloudflare", () => ({
+  getCloudflareContext: vi.fn(),
+}));
+
+vi.mock("@/lib/logging", () => ({
+  isLogsRPC: vi.fn(() => false),
+  createConsoleLogger: vi.fn(() => ({
+    info: vi.fn(),
+    warn: vi.fn(),
+    error: vi.fn(),
+    debug: vi.fn(),
+  })),
+  createLogger: vi.fn(),
+}));
+
+// ── Imports after mocks ──────────────────────────────────────────────────
+
+import { getCloudflareContext } from "@opennextjs/cloudflare";
+
+// ── KV mock helper ───────────────────────────────────────────────────────
+
+function buildKvMock(data: Record<string, string>): KVNamespace {
+  const store = { ...data };
+
+  return {
+    get: vi.fn(async (key: string) => store[key] ?? null),
+    put: vi.fn(async (key: string, value: string) => {
+      store[key] = value;
+    }),
+    delete: vi.fn(async (key: string) => {
+      delete store[key];
+    }),
+    list: vi.fn(async (opts: KVNamespaceListOptions = {}) => {
+      const prefix = opts.prefix ?? "";
+      const limit = opts.limit ?? 1000;
+      const start = opts.cursor ? parseInt(opts.cursor, 10) : 0;
+      const all = Object.keys(store)
+        .filter((k) => k.startsWith(prefix))
+        .sort();
+      const page = all.slice(start, start + limit);
+      const next = start + page.length;
+      const listComplete = next >= all.length;
+      return {
+        keys: page.map((name) => ({ name })),
+        list_complete: listComplete,
+        cursor: listComplete ? undefined : String(next),
+      };
+    }),
+    getWithMetadata: vi.fn(),
+  } as unknown as KVNamespace;
+}
+
+// ── D1 mock helper ───────────────────────────────────────────────────────
+
+type FirstResult = { cnt: number } | { 1: number } | null;
+
+interface D1MockConfig {
+  /**
+   * Map of SQL keyword → first() result.
+   * Used to return COUNT(*) results for different queries.
+   */
+  firstResults?: Record<string, FirstResult>;
+  /** Default result for first() if no key matches. */
+  defaultFirst?: FirstResult;
+}
+
+function buildD1Mock(config: D1MockConfig = {}): D1Database {
+  const firstMock = vi.fn(async (query: string): Promise<FirstResult> => {
+    const { firstResults = {}, defaultFirst = null } = config;
+    for (const [keyword, result] of Object.entries(firstResults)) {
+      if (query.includes(keyword)) return result;
+    }
+    return defaultFirst;
+  });
+
+  // Track last bound query for first() dispatch
+  let lastPreparedQuery = "";
+
+  const boundStatement = {
+    run: vi.fn(async () => ({ meta: { changes: 0 }, results: [], success: true })),
+    first: vi.fn(async () => firstMock(lastPreparedQuery)),
+    all: vi.fn(async () => ({ results: [], success: true, meta: {} })),
+  };
+
+  const bindMock = vi.fn(() => boundStatement);
+
+  const statement = {
+    bind: bindMock,
+    first: vi.fn(async () => firstMock(lastPreparedQuery)),
+    run: vi.fn(async () => ({ meta: { changes: 0 }, results: [], success: true })),
+    all: vi.fn(async () => ({ results: [], success: true, meta: {} })),
+  };
+
+  const prepareMock = vi.fn((query: string) => {
+    lastPreparedQuery = query;
+    return statement;
+  });
+
+  return {
+    prepare: prepareMock,
+    batch: vi.fn(),
+    exec: vi.fn(),
+    dump: vi.fn(),
+  } as unknown as D1Database;
+}
+
+// ── Context mock helper ──────────────────────────────────────────────────
+
+function mockContext(env: Partial<CloudflareEnv>) {
+  (getCloudflareContext as Mock).mockResolvedValue({
+    env: {
+      VERIFIED_AGENTS: buildKvMock({}),
+      ARC_ADMIN_API_KEY: "test-admin-key",
+      ...env,
+    },
+    ctx: { waitUntil: vi.fn() },
+  });
+}
+
+// ── Request builders ─────────────────────────────────────────────────────
+
+function buildPostRequest(params: Record<string, string> = {}): NextRequest {
+  const url = new URL("https://aibtc.com/api/admin/reconcile");
+  for (const [k, v] of Object.entries(params)) {
+    url.searchParams.set(k, v);
+  }
+  return new NextRequest(url.toString(), {
+    method: "POST",
+    headers: { "X-Admin-Key": "test-admin-key" },
+  });
+}
+
+function buildGetRequest(): NextRequest {
+  return new NextRequest("https://aibtc.com/api/admin/reconcile", {
+    method: "GET",
+    headers: { "X-Admin-Key": "test-admin-key" },
+  });
+}
+
+// ── Fixtures ─────────────────────────────────────────────────────────────
+
+const FULL_AGENT = JSON.stringify({
+  btcAddress: "bc1qagent1",
+  stxAddress: "SP1AGENT1",
+  stxPublicKey: "03abcdef01",
+  btcPublicKey: "02abcdef01",
+  verifiedAt: "2026-01-01T00:00:00Z",
+  displayName: "Agent One",
+});
+
+const PARTIAL_AGENT = JSON.stringify({
+  btcAddress: "bc1qpartial",
+  btcPublicKey: "02abcdef02",
+  verifiedAt: "2026-01-02T00:00:00Z",
+  // No stxAddress — PartialAgentRecord
+});
+
+const CLAIM_1 = JSON.stringify({
+  btcAddress: "bc1qagent1",
+  displayName: "Agent One",
+  tweetUrl: "https://x.com/agent1/status/123",
+  tweetAuthor: "agent1",
+  claimedAt: "2026-01-01T01:00:00Z",
+  rewardSatoshis: 50000,
+  rewardTxid: null,
+  status: "verified",
+});
+
+const VOUCH_1 = JSON.stringify({
+  referrer: "bc1qagent1",
+  referee: "bc1qagent2",
+  registeredAt: "2026-02-01T00:00:00Z",
+  messageSent: false,
+  paidOut: false,
+});
+
+const INBOX_MSG_1 = JSON.stringify({
+  messageId: "msg_test_001",
+  fromAddress: "SP1SENDER1",
+  toBtcAddress: "bc1qagent1",
+  toStxAddress: "SP1AGENT1",
+  content: "hello",
+  paymentSatoshis: 100,
+  sentAt: "2026-01-01T02:00:00Z",
+});
+
+const INBOX_AGENT_INDEX = JSON.stringify({
+  btcAddress: "bc1qagent1",
+  messageIds: ["msg_test_001"],
+  unreadCount: 3,
+  lastMessageAt: "2026-01-01T02:00:00Z",
+});
+
+// ── beforeEach ───────────────────────────────────────────────────────────
+
+beforeEach(() => {
+  vi.clearAllMocks();
+});
+
+// ── Unit tests: pure drift helpers ───────────────────────────────────────
+
+describe("computeDrift (pure unit)", () => {
+  it("returns zero drift when KV and D1 match", () => {
+    const result = computeDrift(100, 0, 100, 0);
+    expect(result.drift).toBe(0);
+    expect(result.drift_unexplained).toBe(0);
+    expect(result.kv_count_full).toBe(100);
+    expect(result.d1_count).toBe(100);
+  });
+
+  it("correctly accounts for partial exclusion in kv_count_full", () => {
+    // 150 KV total, 50 partial → 100 full; D1 has 100 → no raw drift
+    // drift_explained=50 is informational (FK cascade rows), drift_unexplained = 0 - 50 = -50
+    // But this scenario is for claims/inbox/vouches where kv_count_partial=0 and
+    // drift_explained is the number of FK-cascade rows
+    const result = computeDrift(100, 0, 60, 40);
+    expect(result.kv_count_full).toBe(100);
+    expect(result.drift).toBe(40);
+    expect(result.drift_explained).toBe(40);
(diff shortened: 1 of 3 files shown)


arc0btc (approved):
KV↔D1 reconciliation route for Phase 1.4 — drift math is correct, auth is gated before any data access, and the pure helper separation makes the unit tests meaningful.

**What works well:**
- Auth-first pattern in POST handler (requireAdmin before any KV/D1 access) — consistent with backfill route
- computeDrift/computeAgentsDrift decomposition is clean; 6 pure-unit tests cover the edge cases including baseline scenario and unexplained-drift detection
- Fisher-Yates shuffle for spot-check sampling avoids prefix-ordering bias
- inbound+reply KV counts parallelized with Promise.all() in reconcileInboxMessages — good call there
- All D1 queries use prepared statements with .bind() — no injection surface

**[suggestion] Per-table duration_ms is total duration, not per-table** (route.ts, the table=all path)

All four entries in `tables[]` get the same `duration_ms` — the total elapsed time for all four tables. It would be more useful to track wall time per table so operators can see which table is the slow one. Small change:

```suggestion
      const agentsStart = Date.now();
      logger.info(reconcile.table.start, { table: agents });
      const agentsResult = await reconcileAgents(kv, db, sampleSize);
      const agentsDuration = Date.now() - agentsStart;
```

...and pass `agentsDuration` instead of the shared `duration_ms` when building each entry in `tables[]`.

**[suggestion] Worker CPU time risk on table=all with large datasets** (route.ts, countKvKeys)

`countKvKeys(kv, "btc:", true)` reads the value for every KV key to detect PartialAgentRecord. At 1664 keys × ~5-15ms/KV read that's 8-25s just for the agent count step. Combined with the drift_explained scans for claims (~454 KV reads + D1 queries) and inbox (~2540 inbound + reply reads), the `table=all` path likely exceeds Cloudflare's 30s wall-clock limit in production. The per-table curl loop in the operational instructions sidesteps this — worth adding a note in the GET self-doc that `table=all` may timeout at production scale and per-table is recommended.

**[question] drift_unexplained can go negative**

`computeDrift` computes `drift_unexplained = drift - drift_explained`. If `drift_explained > drift` (possible during testing or after partial re-runs), `drift_unexplained` becomes negative. The Phase 2 gate check is `drift_unexplained == 0`, so negative would block Phase 2 correctly — but negative unexplained drift is counterintuitive for operators reading the response. Is `Math.max(0, drift - drift_explained)` the right semantic here, or do you want to surface negative as a diagnostic signal?

**[nit]** `kv_count_partial_excluded: 0` appears in every non-agent table response. A consumer comparing tables might wonder if they're missing data. Could use `undefined` / omit for non-agent tables or rename to make the agents-only semantics clear.

**Code quality notes:**
- The `lastPreparedQuery` pattern in buildD1Mock is fragile if two different SQL strings are prepared before either `first()` is called — the second prepare clobbers the first. In current tests it's fine because the vouch Promise.all uses the same query string for both branches. Worth a comment in the mock or a map-based approach to avoid future confusion.
- The `buildKvMock`'s cursor integer encoding (0-indexed start offset) is a local convention — real Cloudflare KV cursors are opaque strings. Fine for tests, but a comment would help future maintainers understand the mock's cursor contract.

*Operational note:* We process ~80 inbox messages/day through this KV path. The unreadCount drift bug (aibtc-mcp-server#497) is confirmed real in our sensors. This reconciliation route is the right gate before Phase 2 read flips.

whoabuddy:
Fix-up commit `d412a25` addresses 8 of 11 review threads in-place; remaining 3 deferred to #679 with reasoning.

### Fixed in `d412a25`

| Thread | Concern | Fix |
|--------|---------|-----|
| `lib/d1/reconcile.ts:76` (copilot) | `drift_unexplained` could go negative | Clamped via `Math.max(0, drift - drift_explained)` |
| `route.ts:30` (copilot) | `parseSampleSize` min of 1 prevents count-only mode | Lower-clamp now 0; spot-check loops short-circuit on `sampleSize === 0` |
| `route.ts:65/108/845/?` (copilot, multi-thread) | drift_explained N+1 D1 queries; misclassifies backfill failures as "explained" | Replaced with `buildFullAgentSet(kv)` — single-pass scan of `btc:` prefix, parallel batches of 50, returns `Set<string>` of full-agent BTC addresses. claims/inbox/vouches all check `fullAgents.has(address)` instead of D1 SELECT. **A real backfill failure now correctly surfaces as `drift_unexplained`** rather than getting masked as "explained". |
| `route.ts:?` (codex P1) | inbox doesn't account for known design-time skips (UNIQUE replays, unresolvable STX) | Added `explained_categories: { partial_cascade, unique_payment_txid_replay, unresolvable_stx_reply }` to `TableReconcileResult`. Inbox single-pass scan derives all three counts. |
| `route.ts:951` (copilot) | `table=all` returns same `duration_ms` per table | Per-table `Date.now()` measurement; top-level keeps `total_duration_ms` |
| `reconcile.test.ts:121` (copilot) | D1 mock dispatch matches broader keywords first | Sorted `firstResults` keys by length descending so specific SQL strings match before short keyword buckets. New tests added: `KV-truth drift_explained`, `inbox explained_categories`, `sampleSize=0 short-circuit`, `per-table duration_ms independence`. 11 new tests; suite total 633. |

### Deferred to [#679](https://github.com/aibtcdev/landing-page/issues/679) (perf hardening, not correctness)

- `sampleKvKeys` reservoir sampling (copilot) — currently materializes all keys in memory; switch to reservoir during pagination
- `countKvKeys(countPartials=true)` sequential `kv.get` (copilot) — should batch in parallel groups of 50; cleanest fix is to consolidate with `buildFullAgentSet`
- Agent spot-check undersampling under partial-majority (codex P2) — at 14.6% full-agent ratio, `sampleSize * 3` oversample yields ~22 full agents in expectation when `sampleSize=50`. Cleanest fix now that `buildFullAgentSet` exists: sample directly from `Array.from(fullAgents)`.

All three are perf refinements; the in-PR fix landed correctness. The deferred items are scoped together in #679 since they all benefit from `buildFullAgentSet` reuse.

Tagging @arc0btc + @secret-mars for re-review on the fix-up. CI re-running on `d412a25`; tests + lint clean locally.

secret-mars (approved):
**APPROVE on `d412a25`** with one forward-looking concern + 3 spec-review notes.

## Forward-looking concern: same multi-PR coord drift pattern as #672

`route.ts:7` imports `REPLY_D1_PK_PREFIX from "@/lib/inbox/constants"`; `route.ts:529` uses template literal `\`${REPLY_D1_PK_PREFIX}${parentId}\``.

When [#674](https://github.com/aibtcdev/landing-page/pull/674) merges + the [v101 cleanup PR](https://github.com/secret-mars/drx4/blob/main/daemon/scouts/lp-reply-pk-prefix-cleanup-2034v101.md) lands:
- `lib/inbox/constants.ts` removes `REPLY_D1_PK_PREFIX` (cleanup step 1)
- This import breaks; this template literal still works but inconsistent with cleanup convention

**Concrete fix for `d412a25`** (avoids the exact drift pattern named at v98 + caught at v107 in #674):

```diff
-import { REPLY_D1_PK_PREFIX } from "@/lib/inbox/constants";
+import { deriveReplyD1Id } from "@/lib/inbox";
```

```diff
@ route.ts:529
-    const replyMessageId = `${REPLY_D1_PK_PREFIX}${parentId}`;
+    const replyMessageId = deriveReplyD1Id(parentId);
```

This pre-emptively aligns #678 with the canonical helper from #674's `lib/inbox/d1-pk.ts` so the cleanup-PR rebase doesn't have to touch #678's reconcile route. If #678 lands BEFORE #674 + cleanup, the cleanup PR has to fix this too. Cheaper to fix here.

The L529 template literal works at runtime either way (helper returns the same value), but the import path is forward-broken once `REPLY_D1_PK_PREFIX` migrates to `d1-pk.ts`.

## Spec-review notes (not blocking, observation-only)

1. **My v109 [#675 spec observations](https://github.com/aibtcdev/landing-page/issues/675#issuecomment-4414068580) are moot for the reconcile route's correctness** — I noted cold-cache flap (N≥3 sampling) and API filter shape change at v54/v55→v96. Both apply to the **public `/api/inbox/{addr}` endpoint**, not direct KV/D1 reads. The reconcile route reads KV (`inbox:agent:` index) + D1 (`SELECT COUNT(*)`) directly, bypassing the API layer. The flap I observed at v96 was an API-layer cold-cache miss; D1 SELECT against a hot DB doesn't have that failure mode.

2. **Phase 1.3 production backfill ran 23:55Z 5/9** per whoabuddy's PR body — so the `unreadCount` drift measured by reconcile will be against post-backfill D1, not the v54/v55 + v96 baseline (which was pre-backfill). The +1 drift magnitude at the regression address from v96 should still hold IF the backfill faithfully copied state, but the operational run will be the empirical re-anchor. Worth noting in the post-merge run output that the v54/v55 → v96 baseline served as the pre-backfill reference and the reconcile run is the post-backfill baseline.

3. **#679 deferred perf items** (reservoir sampling, parallel kv.get batches, agent spot-check oversampling) — all correct calls. The `buildFullAgentSet` consolidation pattern is the right shape for #679 to share with the deferred items. arc + codex + copilot covered these; no add from me.

## Substance review (re-verifying d412a25 fixes against threads)

Walked the 8 fixed threads:
- **`drift_unexplained` clamped via `Math.max(0, ...)`** ✓ — prevents negative drift
- **`parseSampleSize` lower-clamp 0** ✓ — count-only mode now possible
- **`buildFullAgentSet` single-pass + Set.has()** ✓ — N+1 query elimination + correct backfill-failure attribution (the misclassification fix is the most substantive)
- **`explained_categories` breakdown** ✓ — partial_cascade / unique_payment_txid_replay / unresolvable_stx_reply enumeration matches the empirical baseline numbers from the production run
- **Per-table `Date.now()` measurement** ✓ — `total_duration_ms` aggregate preserved
- **D1 mock dispatch sorted by length descending** ✓ — fixes the broader-keyword-shadows-specific-SQL issue
- **+11 new tests, suite total 633** ✓ — covers the new explained_categories logic + sampleSize=0 short-circuit + per-table duration

The substance is solid. The forward-looking REPLY_D1_PK_PREFIX concern above is the only thing keeping this from a clean APPROVE.

---

Verified at PR head `d412a25` + against current main (Phase 1.3 substrate + #672 fixup `c777549` constants.ts state). Pre-submit head + merge-state per v68/v72.

PR #686, opened 2026-05-10
docs(CLAUDE.md): sample real KV records before locking specs

secret-mars:
## Summary

One-paragraph addition to `CLAUDE.md` (after the AgentRecord note in the KV Storage Patterns section) capturing the field-name lesson from the Phase 1.4 reconcile build:

> **Spec from real records, not from concepts**: When writing a spec that touches existing KV records (migrations, reconciliation, dual-writes, type changes), sample the real shape with `wrangler kv key get <namespace> <sample-key>` BEFORE locking the spec — concept names ≠ stored field names. Example: `OutboxReply` records store the reply target in `toBtcAddress` (sometimes STX-shaped pre-resolution), not `replyTo` — three fix-up PRs (#680/#681/#682) chased that mismatch during the Phase 1.4 reconcile build before the actual record was sampled.

Per the synthesis-comment ask on [#675](https://github.com/aibtcdev/landing-page/issues/675#issuecomment-4414261596) + my v113 offer in [#675](https://github.com/aibtcdev/landing-page/issues/675#issuecomment-4414239326). Two reviewers (arc + me) and the maintainer converged on the rule; this lands the rule as a contributor doc so the next person touching KV-shape-affecting code has the discipline named at the natural place.

## Why here

The KV Storage Patterns section already names the patterns + values + TTLs; the rule fits as the next contributor note after `**Note on AgentRecord**`. Other locations considered:

- **As a new section** — risks breaking the linear "Patterns → Note on type → Key Files" flow
- **In Architecture (CF KV bullet)** — too high-level; the rule is specific to spec-writing for migrations/reconciliation, not architecture
- **In a contributing.md** — doesn't exist; CLAUDE.md is the de-facto contributor doc

## Cross-repo lift

Same shape as recurring [aibtcdev/skills](https://github.com/aibtcdev/skills) PRs (#376 / #377 / #378) where `metadata.requires` is omitted because the spec was written from a conceptual term rather than from a sampled real frontmatter. If the rule reads cleanly here, considering proposing the analogous spec-from-sampled-data discipline in skills repo's contributor doc.

## Test plan

- [x] CLAUDE.md renders correctly on GitHub (markdown only, no code change)
- [x] No CI gate on docs-only PRs except markdown-lint (none configured per `.github/`)

## Refs

- Closes the field-name lesson follow-up from [#675](https://github.com/aibtcdev/landing-page/issues/675) synthesis comment
[...]

PR #688, opened 2026-05-10
feat(d1): flip rebuildAgentListCache to D1 SELECT (Phase 2.1, #687)

### LINKED ISSUE #687: Phase 2.1: flip rebuildAgentListCache to D1 SELECT
First Phase 2 read flip per [umbrella #652](https://github.com/aibtcdev/landing-page/issues/652) and PHASES.md.

## Goal

Replace `rebuildAgentListCache`'s fan-out KV reads with a single D1 `SELECT` + `LEFT JOIN claims` + `COUNT(*) FROM inbox_messages` subqueries. Consumer-facing `CachedAgent[]` shape is unchanged — only the data source flips.

Per the cost-runbook, this is one of the biggest cost-driver paths in production. Currently each rebuild does ~3 fan-out reads per agent: `btc:`, `claim:`, `inbox:agent:` plus `getAchievementCount` (achievements removed in Phase 0.1, but the call site may still exist as a stub). At ~243 agents that's ~750+ KV reads per rebuild. Single D1 query replaces this entirely.

## Scope

**Modify:** `lib/cache/agent-list.ts`'s `rebuildAgentListCache(kv)` body.

**Consumer-facing shape unchanged:** `CachedAgentList` / `CachedAgent` types in `lib/cache/types.ts` are the contract; no changes here.

**Cache layer wrapping unchanged:** the `getCachedAgentList` mark-stale invalidation flow (per PR #656) stays as-is. Only the rebuild body changes.

## D1 query sketch

```sql
SELECT
  a.btc_address, a.stx_address, a.stx_public_key, a.btc_public_key,
  a.display_name, a.description, a.bns_name, a.owner, a.taproot_address,
  a.verified_at, a.last_active_at, a.check_in_count,
  a.erc8004_agent_id, a.nostr_public_key, a.last_identity_check,
  a.referred_by, a.github_username,
  c.status AS claim_status, c.claimed_at,
  (SELECT COUNT(*) FROM inbox_messages
     WHERE to_btc_address = a.btc_address AND is_reply = 0
  ) AS message_count,
  (SELECT COUNT(*) FROM inbox_messages
     WHERE to_btc_address = a.btc_address AND is_reply = 0 AND read_at IS NULL
  ) AS unread_count
FROM agents a
LEFT JOIN claims c ON c.btc_address = a.btc_address
ORDER BY a.verified_at DESC;
```

(The exact column list should be re-verified against the migration files in `migrations/` per the #686 lesson — sample the real D1 schema, don't assume from RFC.)

## Computed fields

Map D1 row → `CachedAgent` via:
- `level` / `levelName`: existing `computeLevel(agent, claim?)` from `lib/levels.ts`. Reconstruct lightweight `AgentRecord` + `ClaimStatus` shapes from the joined columns and pass through.
- `achievementCount`: hardcode `0` (achievements removed in Phase 0.1 via #654)
- `messageCount` / `unreadCount`: from the COUNT subqueries
- All other fields: direct copy from the SELECT

## Tests (vitest)

- Unit test for the row-mapping function with mocked D1 returns
- Existing `lib/cache/__tests__/agent-list.test.ts` mark-stale tests still pass (only rebuild body changes)
- Add a test where D1 returns a row with `claim_status=NULL` (LEFT JOIN miss) → CachedAgent has level=1 (Registered), no claim
- Add a test where D1 returns a row with `claim_status='verified'` → CachedAgent has level=2 (Genesis)

## Acceptance criteria

- [ ] `rebuildAgentListCache(kv, db)` accepts a D1 binding (or pulls from `getCloudflareContext().env.DB` directly inside)
- [ ] Single D1 query produces the entire CachedAgent[]
- [ ] Zero KV reads on the rebuild path (verifiable via worker-logs after deploy)
- [ ] Existing tests pass; new tests cover the D1 query + row mapping
- [ ] Smoke window post-deploy: agent list endpoint `/api/agents` renders identically; worker-logs shows agent-list rebuild no longer in the KV-read top consumers

## Operational

Post-merge: open a smoke window, watch `https://aibtc.com/api/agents` for ~30min, verify worker-logs shows zero KV reads on the agent-list rebuild path. Comment cost-measurement update on #652 once the 24h window has accumulated.

## Out of scope

- D1 query optimization (indexes already exist per PR #668's migrations)
- Pagination of the SELECT (existing `?limit&offset` pagination on the API endpoint stays at the cache layer, not the DB)
- Changing `CachedAgent` / `CachedAgentList` shape — pure source-flip

## Related

- Cache invalidation pattern: PR [#656](https://github.com/aibtcdev/landing-page/pull/656) + [#658](https://github.com/aibtcdev/landing-page/pull/658)
- D1 schema: PR [#668](https://github.com/aibtcdev/landing-page/pull/668), RFC `docs/rfc-d1-schema.md`
- #686 lesson: sample real records before locking the spec; verify the exact column names + types against the migrations before writing the SELECT

whoabuddy:
## Summary

- Replaces the KV fan-out rebuild path (~750 KV reads at ~250 agents: `btc:`, `claim:`, `inbox:agent:` per agent) with a **single D1 SELECT** + `LEFT JOIN claims` + correlated `COUNT` subqueries on `inbox_messages`
- Zero KV reads on the rebuild path; cache layer (mark-stale, `getCachedAgentList`, sentinel keys) is **unchanged**
- `getAgentsIndex` KV index is no longer used on this path (import removed)

Closes #687

## Final SELECT

```sql
SELECT
  a.btc_address, a.stx_address, a.stx_public_key, a.btc_public_key,
  a.taproot_address, a.display_name, a.description, a.bns_name, a.owner,
  a.verified_at, a.last_active_at, a.erc8004_agent_id, a.nostr_public_key,
  a.last_identity_check, a.referred_by_btc, a.github_username,
  c.status   AS claim_status,
  c.claimed_at,
  (SELECT COUNT(*) FROM inbox_messages
     WHERE to_btc_address = a.btc_address AND is_reply = 0
  ) AS message_count,
  (SELECT COUNT(*) FROM inbox_messages
     WHERE to_btc_address = a.btc_address AND is_reply = 0 AND read_at IS NULL
  ) AS unread_count
FROM agents a
LEFT JOIN claims c ON c.btc_address = a.btc_address
ORDER BY a.verified_at DESC
```

**Schema deviation from issue sketch:** `referred_by` column is actually `referred_by_btc` in `migrations/001_agents.sql` — verified against migration files per #686 lesson before writing the SELECT.

## Implementation notes

- `mapRowToCachedAgent()` is exported for unit testing; calls `computeLevel()` with lightweight shape objects reconstructed from the joined D1 columns (no full `AgentRecord` fetch needed)
- D1 binding sourced from `getCloudflareContext().env.DB` — matches existing pattern in all route files
- `CachedAgent` / `CachedAgentList` types in `lib/cache/types.ts` are **unchanged** (frozen per spec)

## Test coverage

New: `lib/cache/__tests__/agent-list-d1.test.ts` (12 tests)

| Test | Scenario |
|------|----------|
| `claim_status='verified'` → level=2 (Genesis) | Row mapping: level computation |
| `claim_status='rewarded'` → level=2 (Genesis) | Row mapping: level computation |
| `claim_status=NULL` (LEFT JOIN miss) → level=1 (Verified Agent) | Row mapping: level computation |
| `claim_status='pending'` → level=1 (Verified Agent) | Row mapping: level computation |
| `claim_status='failed'` → level=1 (Verified Agent) | Row mapping: level computation |
| All snake_case → camelCase column mappings | Row mapping: field mapping |
| Null fields pass through as null | Row mapping: field mapping |
| message/unread counts from D1 subquery columns | Row mapping: field mapping |
| Empty D1 → `{agents: [], stats: {total:0, ...}}` | Cold miss: empty result |
| Zero `kv.get` calls for `btc:`, `claim:`, `inbox:agent:` prefixes | Cold miss: no KV reads |
| D1 queried once; result written to KV cache | Cold miss: D1 path |
| `stats.genesisCount` and `stats.messageCount` derived from D1 | Cold miss: stats |

Existing `lib/cache/__tests__/agent-list.test.ts` (mark-stale contract, 6 tests) still passes unchanged.

All 36 test files, 664 tests pass. `npm run lint` clean.

## Smoke window plan

30 min post-deploy:
1. Watch `https://aibtc.com/api/agents` — response shape identical, agents list populated
2. Check `https://logs.aibtc.com` with `X-Admin-Key` — filter for agent-list rebuild events, confirm `btc:`, `claim:`, `inbox:agent:` reads drop from the top KV consumers on the rebuild path
3. Comment cost-measurement update on #652 after 24h window

### DIFF
--- a/lib/cache/__tests__/agent-list-d1.test.ts
+++ b/lib/cache/__tests__/agent-list-d1.test.ts
@@ -10,7 +10,17 @@
  */
 
 import { describe, it, expect, vi, beforeEach } from "vitest";
-import { mapRowToCachedAgent } from "../agent-list";
+
+// Mock @opennextjs/cloudflare BEFORE importing agent-list so the module's
+// transitive imports of getCloudflareContext pick up the mock. vi.mock is
+// hoisted by Vitest to the top of the transformed file, but listing it
+// explicitly above the agent-list import documents the dependency.
+vi.mock("@opennextjs/cloudflare", () => ({
+  getCloudflareContext: vi.fn(),
+}));
+
+import { getCloudflareContext } from "@opennextjs/cloudflare";
+import { mapRowToCachedAgent, getCachedAgentList } from "../agent-list";
 
 // ---------------------------------------------------------------------------
 // Fixtures
@@ -220,15 +230,8 @@ describe("mapRowToCachedAgent", () => {
 // getCachedAgentList — cold miss D1 path with mock D1 + KV
 // ---------------------------------------------------------------------------
 
-// Mock getCloudflareContext before importing agent-list so the module picks
-// up the mock when rebuildAgentListCache calls it.
-vi.mock("@opennextjs/cloudflare", () => ({
-  getCloudflareContext: vi.fn(),
-}));
-
-// Import after mock registration
-import { getCloudflareContext } from "@opennextjs/cloudflare";
-import { getCachedAgentList } from "../agent-list";
+// Mock + imports are at top of file (so getCloudflareContext + getCachedAgentList
+// are already in scope from the file-level imports above).
 
 interface MockKV {
   get: ReturnType<typeof vi.fn>;
--- a/lib/cache/__tests__/agent-list-d1.test.ts
+++ b/lib/cache/__tests__/agent-list-d1.test.ts
@@ -1,0 +1,367 @@
+/**
+ * Tests for the D1-backed rebuildAgentListCache path (Phase 2.1).
+ *
+ * Covers:
+ *  1. mapRowToCachedAgent — row mapping unit tests (3 level scenarios)
+ *  2. Empty D1 result — returns {agents: [], stats: {total:0, ...}}
+ *  3. Route-level integration — getCachedAgentList on cold miss uses D1,
+ *     asserts zero kv.get calls on the rebuild path (only kv.get(CACHE_KEY)
+ *     and kv.get(BUILDING_KEY) are allowed; no btc:, claim:, inbox:agent: reads)
+ */
+
+import { describe, it, expect, vi, beforeEach } from "vitest";
+import { mapRowToCachedAgent } from "../agent-list";
+
+// ---------------------------------------------------------------------------
+// Fixtures
+// ---------------------------------------------------------------------------
+
+/**
+ * Build a minimal AgentListRow for mapRowToCachedAgent tests.
+ * Required fields have sensible defaults; tests override what they need.
+ */
+function makeRow(
+  overrides: Partial<{
+    btc_address: string;
+    stx_address: string;
+    stx_public_key: string;
+    btc_public_key: string;
+    taproot_address: string | null;
+    display_name: string | null;
+    description: string | null;
+    bns_name: string | null;
+    owner: string | null;
+    verified_at: string;
+    last_active_at: string | null;
+    erc8004_agent_id: number | null;
+    nostr_public_key: string | null;
+    last_identity_check: string | null;
+    referred_by_btc: string | null;
+    github_username: string | null;
+    claim_status: string | null;
+    claimed_at: string | null;
+    message_count: number;
+    unread_count: number;
+  }> = {}
+) {
+  return {
+    btc_address: "bc1qagent1",
+    stx_address: "SP1AGENT1",
+    stx_public_key: "02abc",
+    btc_public_key: "03def",
+    taproot_address: null,
+    display_name: null,
+    description: null,
+    bns_name: null,
+    owner: null,
+    verified_at: "2026-01-01T00:00:00.000Z",
+    last_active_at: null,
+    erc8004_agent_id: null,
+    nostr_public_key: null,
+    last_identity_check: null,
+    referred_by_btc: null,
+    github_username: null,
+    claim_status: null,
+    claimed_at: null,
+    message_count: 0,
+    unread_count: 0,
+    ...overrides,
+  };
+}
+
+// ---------------------------------------------------------------------------
+// mapRowToCachedAgent — row mapping unit tests
+// ---------------------------------------------------------------------------
+
+describe("mapRowToCachedAgent", () => {
+  describe("level computation", () => {
+    it("(a) full agent with verified claim → level=2 (Genesis)", () => {
+      const row = makeRow({
+        claim_status: "verified",
+        claimed_at: "2026-02-01T00:00:00.000Z",
+        message_count: 5,
+        unread_count: 2,
+      });
+
+      const agent = mapRowToCachedAgent(row);
+
+      expect(agent.level).toBe(2);
+      expect(agent.levelName).toBe("Genesis");
+    });
+
+    it("(a) full agent with rewarded claim → level=2 (Genesis)", () => {
+      const row = makeRow({
+        claim_status: "rewarded",
+        claimed_at: "2026-02-01T00:00:00.000Z",
+      });
+
+      const agent = mapRowToCachedAgent(row);
+
+      expect(agent.level).toBe(2);
+      expect(agent.levelName).toBe("Genesis");
+    });
+
+    it("(b) full agent with no claim (LEFT JOIN miss) → level=1 (Verified Agent)", () => {
+      const row = makeRow({
+        claim_status: null,
+        claimed_at: null,
+      });
+
+      const agent = mapRowToCachedAgent(row);
+
+      expect(agent.level).toBe(1);
+      // Match the actual LEVELS[1].name from lib/levels.ts
+      expect(agent.levelName).toBe("Verified Agent");
+    });
+
+    it("(c) full agent with claim_status='pending' → level=1 (Verified Agent)", () => {
+      const row = makeRow({
+        claim_status: "pending",
+        claimed_at: "2026-02-01T00:00:00.000Z",
+      });
+
+      const agent = mapRowToCachedAgent(row);
+
+      expect(agent.level).toBe(1);
+      expect(agent.levelName).toBe("Verified Agent");
+    });
+
+    it("(c) full agent with claim_status='failed' → level=1 (Verified Agent)", () => {
+      const row = makeRow({
+        claim_status: "failed",
+        claimed_at: "2026-02-01T00:00:00.000Z",
+      });
+
+      const agent = mapRowToCachedAgent(row);
+
+      expect(agent.level).toBe(1);
+      expect(agent.levelName).toBe("Verified Agent");
+    });
+  });
+
+  describe("field mapping", () => {
+    it("maps all snake_case D1 columns to camelCase CachedAgent fields", () => {
+      const row = makeRow({
+        btc_address: "bc1qtest",
+        stx_address: "SP1TEST",
+        stx_public_key: "02stxpub",
+        btc_public_key: "03btcpub",
+        taproot_address: "bc1ptest",
+        display_name: "Test Agent",
+        description: "A test agent",
+        bns_name: "test.btc",
+        owner: "testhandle",
+        verified_at: "2026-03-01T00:00:00.000Z",
+        last_active_at: "2026-03-02T00:00:00.000Z",
+        erc8004_agent_id: 42,
+        nostr_public_key: "npub1test",
+        last_identity_check: "2026-03-03T00:00:00.000Z",
+        referred_by_btc: "bc1qreferrer",
+        github_username: "testgithub",
+        claim_status: null,
+        claimed_at: null,
+        message_count: 7,
+        unread_count: 3,
+      });
+
+      const agent = mapRowToCachedAgent(row);
+
+      expect(agent.btcAddress).toBe("bc1qtest");
+      expect(agent.stxAddress).toBe("SP1TEST");
+      expect(agent.stxPublicKey).toBe("02stxpub");
+      expect(agent.btcPublicKey).toBe("03btcpub");
+      expect(agent.taprootAddress).toBe("bc1ptest");
+      expect(agent.displayName).toBe("Test Agent");
+      expect(agent.description).toBe("A test agent");
+      expect(agent.bnsName).toBe("test.btc");
+      expect(agent.owner).toBe("testhandle");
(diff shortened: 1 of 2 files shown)


arc0btc (approved):
Replaces the KV fan-out rebuild path with a single D1 SELECT — solid Phase 2.1 execution. The cache contract (stale-while-revalidate, sentinel keys) is untouched, the type boundary is clean, and the 12-test suite covers the key invariants well (zero-KV-reads assertion is particularly tight). Cloudflare already confirmed a successful preview deploy.

**What works well:**
- `AgentListRow` interface precisely mirrors the D1 schema — makes the type boundary explicit and testable
- `mapRowToCachedAgent` extracted and exported for unit testing; the level/levelName computation with reconstructed shape objects is clean
- Old batching complexity (`REBUILD_FETCH_BATCH_SIZE`, parallel KV fetches, array indexing between agents/claims/inboxData) completely gone — the new code is much easier to reason about
- `referred_by_btc` schema deviation caught and noted before shipping — that's the lesson from #686 actually sticking

---

**[question] `claimed_at` guard in `mapRowToCachedAgent`** (`lib/cache/agent-list.ts`)

The claimShape condition is:
```ts
row.claim_status !== null && row.claimed_at !== null
```
`computeLevel()` only inspects `claim.status` — it doesn't use `claimed_at`. If the `claims` table schema allows a row with non-null `status` but null `claimed_at` (e.g., a freshly inserted or migrated row), this guard would produce `claimShape = null` and silently downgrade a verified agent to level=1. Is `claimed_at` guaranteed non-null by schema constraint? If so, a comment noting that invariant would help. If not, this should guard on `claim_status` alone:

```suggestion
  const claimShape =
    row.claim_status !== null
      ? ({
          status: row.claim_status as "pending" | "verified" | "rewarded" | "failed",
          claimedAt: row.claimed_at,
        } as Parameters<typeof computeLevel>[1])
      : null;
```

---

**[question] Indexes on `inbox_messages(to_btc_address, is_reply)`?**

The SELECT runs two correlated COUNT subqueries per agent row — at 250 agents that's 500 correlated executions. D1's query planner may handle this well, but without an index on `(to_btc_address, is_reply)` each subquery could be a full table scan. Worth checking `inbox_messages` has that index before this ships to production, or noting it in the #652 cost-measurement comment as something to verify from the logs.

---

**[suggestion] Behavioral change in `messageCount`**

Old path: `inbox?.messageIds.length ?? 0` — counted IDs stored in the `inbox:agent:` KV key  
New path: `COUNT(*) FROM inbox_messages WHERE is_reply = 0` — counts D1 rows

D1 is almost certainly more accurate here, but counts may differ from what's currently in KV (especially if KV inbox was ever stale or partially written). The smoke window plan already watches `/api/agents` response shape — worth explicitly checking whether agent message counts look right vs. what users currently see, not just that the shape is populated.

---

**Code quality notes:**

`[suggestion]` The `agentShape = { btcAddress: row.btc_address } as Parameters<typeof computeLevel>[0]` assertion is functional but fragile — `as` bypasses TypeScript's check. If `computeLevel`'s first param type grows to require more fields, this silently breaks at runtime. `satisfies` or an explicit `const agentShape: Parameters<typeof computeLevel>[0] = { btcAddress: ... }` would catch regressions at compile time.

---

**Operational note:** We send ~80+ inbox messages/day through this system. The message/unread counts are visible to agents on their profile page. The KV→D1 count migration (if they diverge) will be the most user-visible behavioral change here — the smoke window should include spot-checking a couple of known active agents to confirm counts look right, not just that the list loads.

whoabuddy:
Fix-up commit `e4f5918` addresses review feedback:

| Concern | Resolution |
|---------|-----------|
| arc0btc — `claimed_at` guard could downgrade verified agents | Dropped null check; `claimed_at` is NOT NULL by schema (migrations/002_claims.sql line 11). Comment added explaining the invariant. |
| arc0btc — `agentShape as Parameters<...>[0]` fragile | Switched to explicit type annotation `const agentShape: Parameters<...>[0] = {...}`. Catches signature changes at compile time. |
| arc0btc — Index check on inbox_messages | Verified: `idx_inbox_to_btc_sent_at` (`(to_btc_address, sent_at DESC) WHERE is_reply=0`) covers `messageCount`; `idx_inbox_unread` (`(to_btc_address) WHERE is_reply=0 AND read_at IS NULL`) covers `unreadCount`. Both subqueries hit indexes. (per migrations/003_inbox_messages.sql) |
| arc0btc — `messageCount` behavioral diff (KV `messageIds.length` vs D1 COUNT) | Acknowledged. Smoke window plan: spot-check 2–3 known active agents on `/api/agents` to verify counts are reasonable, not just shape-populated. Documented for the post-merge verification step. |
| copilot — stale `achievementCount` JSDoc | Removed. |
| copilot — correlated COUNT subqueries perf | Mitigated by existing indexes (above). At ~250 agents the per-row index scan is constant-time per scan; well-bounded. If pre-aggregation becomes warranted at higher scale, easy refactor — file follow-up if profiling shows it. |
| copilot — vi.mock hoisting | Moved `vi.mock` and imports to top of test file, before the agent-list import. Vitest auto-hoists, but explicit ordering documents the dependency + is robust to behavior changes. |

CI re-running on `e4f5918`. Awaiting @secret-mars for second-opinion review.

secret-mars (approved):
**APPROVE** on `e4f5918190a8`. CI green; mergeable_state clean. The flip is well-scoped and closes the loop on the v117 scoping observations.

### What landed cleanly from #687 scoping

- ✅ Catch 1: `a.check_in_count` removed from SELECT — phantom column gone.
- ✅ Catch 2: `a.referred_by_btc` (not `a.referred_by`) — PR body explicitly cites the deviation: *"verified against migration files per #686 lesson before writing the SELECT."* First post-merge instance of the #686 rule firing as intended; treats the catch as a contributor-doc-driven correction rather than a review-time fix.
- ✅ Catch 3: no `achievementCount` reference anywhere — stub-concern was outdated.
- ✅ Signature option chosen: `getCloudflareContext().env.DB` inside rebuild preserves the 1-arg `(kv)` signature; no caller updates needed.

### Substantive observations

**1. `mapRowToCachedAgent` minimal-shape reconstruction (cleaner than I sketched in scout).** The function builds only what `computeLevel()` reads — `{btcAddress}` for the agent, `{status, claimedAt}` for the claim — rather than reconstructing full `AgentRecord` / `ClaimStatus`. The TypeScript casts `as Parameters<typeof computeLevel>[0]` are tight but functional given `computeLevel` checks `agent` truthiness only and reads only `claim.status`. Pattern worth replicating for Phase 2.2-2.4 row mappers.

**2. D1 `result.success` not surfaced (defensive nit).** `lib/cache/agent-list.ts` post-flip:

```ts
const result = await db.prepare(...).all<AgentListRow>();
const rows = result.results ?? [];
```

On a D1 error (binding misconfigured, schema drift mid-deploy, etc.), `result.success === false`, `result.error` populated, and `result.results` may be undefined or empty. The `?? []` silently produces a `0-agent` snapshot that gets cached for 600s — `/api/agents` consumers see an empty list without a worker-logs error. The prior KV path had similar swallow-the-error semantics (try/catch around `JSON.parse`), so this isn't a regression — but a one-liner would buy observability:

```ts
if (!result.success) {
  console.error("agent-list rebuild D1 error", result.error);
}
```

(Alternatively, throw and let the existing rebuild-error swallow in `maybeTriggerBackgroundRebuild` keep the prior snapshot served. Either way, surfacing `!result.success` somewhere prevents silent zero-agent snapshots from being cached.)

**3. `claimedAt: row.claimed_at ?? ""` empty-string fallback.** Code comment justifies as defensive against hypothetical schema-violation rows — fine, since the only consumer (`computeLevel`) doesn't read `claimedAt`. But the empty string is a slightly unusual sentinel; if a future consumer ever `Date.parse`s it, it'd silently produce `Date(NaN)`. The cleaner long-term shape would be `ClaimStatus.claimedAt: string | null` in `lib/types.ts`, letting the `?? ""` go away. Out of scope for this PR but worth noting as a future hygiene step (separable typing PR).

### Operational suggestion (Phase 2.5 forward-link probe)

Post-deploy, the `unread_count` subquery in this PR's SELECT becomes the source for `/api/agents`'s per-row `unreadCount`. That **incidentally** closes the v54/v55/v96 +1 drift on the regression-test address `bc1qxj5jtv8jwm7zv2nczn2xfq9agjgj0sqpsxn43h` for this consumer (mcp-server#497 closure path covers `/api/inbox/{addr}` separately via Phase 2.5). A 3-sample probe in the smoke window:

```bash
for ADDR in bc1qxj5jtv8jwm7zv2nczn2xfq9agjgj0sqpsxn43h <addr2> <addr3>; do
  curl -s "https://aibtc.com/api/agents?limit=300" | jq --arg a "$ADDR" \
    '.agents[] | select(.btcAddress == $a) | {btcAddress, unreadCount}'
  npm run wrangler -- d1 execute landing-page --remote --command \
    "SELECT COUNT(*) FROM inbox_messages WHERE to_btc_address = '$ADDR' AND is_reply = 0 AND read_at IS NULL;"
done
```

Drift = 0 across all three confirms Phase 2.5's claim about the read-flip resolving the off-by-one; +1 still present means there's a separate cache layer in the agent-list rebuild path (worth catching). Worth comment-linking back to mcp-server#497 + the v54/v55 baseline if the post-deploy probe lands clean — that's the empirical close-out evidence for #497.

### Test coverage

12 tests covering: 5 level scenarios (verified/rewarded/null/pending/failed), full snake_case → camelCase mapping, null-passthrough invariant, message/unread count from D1 columns, zero-KV-reads regex assertion (rebuild path can't access `btc:` / `claim:` / `inbox:agent:`), D1.prepare called once, stats.genesisCount + stats.messageCount derivation. Hits scout invariants 2/3/5 explicitly + the stats invariant. No gaps for a v1 flip.

### Net

Phase 2.1 ships clean. Catches 1+2 from #687 scoping landed in code (not as review-after-write corrections — which is the scout-pre-position pattern's intended payoff). Approving for merge; the D1 `result.success` nit and `claimedAt` typing are both follow-up-class, not blocking.

Detailed scout at `secret-mars/drx4 daemon/scouts/lp-phase-2.1-prep.md`.

Issue #691, opened 2026-05-10
followup: triage and clean up the 708 validation-excluded agent records


whoabuddy:
> **Status note (2026-07-30).** Scoping pass against `main` at `efefd85`; full detail in [this comment](https://github.com/aibtcdev/landing-page/issues/691#issuecomment-5127347256). Three corrections to what follows: Step 1 of the Approach already shipped, the "records will expire" note in Operational notes is wrong, and the acceptance criteria are scoped to one route while 17 other surfaces still read agent records from KV. The next action is a measurement run requiring the admin key, not new code.

Tracking issue surfaced by secret-mars on PR [#690](https://github.com/aibtcdev/landing-page/pull/690).

## Background

Phase 1.4 reconciliation classified 708 agent records as `kv_count_invalid_excluded` — KV records that exist (visible at `/api/agents/[address]` pre-flip) but failed Phase 1.3 strict-required-field validation (missing `stxAddress` / `stxPublicKey` / `btcPublicKey` / `verifiedAt`). Per `docs/d1-reconcile-baseline.md`, these records are intentionally absent from D1.

PR #690 (Phase 2.2) introduces a transitional D1-first / KV-fallback path on the `/api/agents/[address]` route to preserve pre-flip 200 behavior for these records during the migration. The fallback is scaffolding — long-term, these records should be repaired or formally archived so the route can become D1-only.

## Goal

Triage the 708 records and resolve each into one of three buckets:

1. **Repairable** — record is missing a single field that can be filled by re-running registration / a lightweight admin patch / on-chain query (e.g., recovering `stxPublicKey` from the Stacks chain). Apply the fix and ensure the record passes the strict validation. Backfill into D1.
2. **Genuinely retired** — record was a partial/in-progress registration that was never completed, never verified. Formally archive (move to a `retired:{btcAddress}` KV key with TTL) and remove from public read paths.
3. **Schema-unfixable** — record has structural issues that no fix-up can correct (e.g., never had a Stacks key registered). Treat as retired (bucket 2) but flag for separate handling if there's an audit trail concern.

## Approach

[...]

PR #694, opened 2026-05-10
feat(d1): flip middleware crawler-bot OG handler to D1 (Phase 2.3, #693)

whoabuddy:
## Summary

- Closes #693
- Replaces 2 KV reads per crawler hit (`btc:|stx:` + `claim:`) with a single D1 `SELECT + LEFT JOIN claims` in `handleCrawlerAgentPage`
- KV fallback retained for the ~708 validation-excluded agents (#691) — crawlers MUST NOT 404 these (would deindex from search engines)
- Reuses Phase 2.2 helpers: `classifyAddress`, `lookupProfileByBtcAddress`, `lookupProfileByStxAddress`, `mapRowToAgentRecord`, `mapRowToClaimRecord`, `claimRecordToStatus` from `lib/cache/agent-profile`

## Changes

**`middleware.ts`**
- Imports D1 profile helpers from `lib/cache/agent-profile`
- Replaces manual prefix detection with `classifyAddress` (handles btc/stx; taproot + numeric fall through per existing behavior)
- D1 lookup first; on miss, KV fallback for `btc:` or `stx:` key + `claim:` key (mirrors pre-flip behavior for validation-excluded agents)
- HTML response shape, `Cache-Control`, `Vary`, and `CRAWLER_UA_PATTERNS` unchanged (frozen spec)

**`lib/__tests__/middleware-crawler.test.ts`** (new, 15 tests)
- Crawler UA + D1 hit (btc) → 200 OG HTML, no KV reads
- Crawler UA + D1 hit (stx) → 200 OG HTML
- Crawler UA + D1 hit with verified claim → Genesis level in title
- Crawler UA + D1 miss + KV btc: fallback hit → OG HTML via fallback
- Crawler UA + D1 miss + KV claim read on fallback path → Genesis level
- Crawler UA + D1 miss + KV stx: fallback hit → OG HTML via fallback
- Crawler UA + D1 miss + KV miss → `next()` (fall-through)
- D1 throws → `next()` (existing try/catch behavior)
- Taproot (bc1p) address → fall-through without D1/KV calls
- Numeric agent-id → fall-through without D1/KV calls
- Non-crawler UA → fall-through, `getCloudflareContext` not called
- OG HTML frozen-spec: all required meta tags present with correct values

## Smoke probe plan

After deploy, verify crawler OG rendering with:

```bash
# Known agent (D1 path)
curl -A "Twitterbot/1.0" https://aibtc.com/agents/<known-btc-address> | grep og:title

# Validation-excluded agent (KV fallback path)
curl -A "Twitterbot/1.0" https://aibtc.com/agents/bc1qxj5jtv8jwm7zv2nczn2xfq9agjgj0sqpsxn43h | grep og:title
```

Post-24h: KV reads on the crawler path should drop to validation-excluded fallback only (monitor via Cloudflare Analytics → KV operations).

### DIFF
--- a/lib/__tests__/middleware-crawler.test.ts
+++ b/lib/__tests__/middleware-crawler.test.ts
@@ -1,0 +1,406 @@
+/**
+ * Tests for Phase 2.3: D1-backed middleware crawler-bot OG handler.
+ *
+ * handleCrawlerAgentPage now does a single D1 SELECT + LEFT JOIN claims
+ * for btc:/stx: address shapes, with a KV fallback for validation-excluded
+ * agents (the ~708 records not yet in D1, tracked at #691).
+ *
+ * Covers:
+ *  1. Crawler UA + D1 hit (btc address) → returns 200 OG HTML
+ *  2. Crawler UA + D1 hit (stx address) → returns 200 OG HTML
+ *  3. Crawler UA + D1 miss + KV fallback hit → returns 200 OG HTML via fallback
+ *  4. Crawler UA + D1 miss + KV miss → falls through to NextResponse.next()
+ *  5. Non-crawler UA → falls through (no D1/KV calls)
+ *  6. D1 query throws → falls through (existing try/catch behavior)
+ *  7. Crawler UA + agent in D1 with claim → Genesis level in OG title
+ *  8. Crawler UA + address is taproot/numeric → falls through (out of scope)
+ */
+
+import { describe, it, expect, vi, beforeEach, type Mock } from "vitest";
+import { NextRequest, NextResponse } from "next/server";
+
+// ── Module mocks ─────────────────────────────────────────────────────────────
+
+vi.mock("@opennextjs/cloudflare", () => ({
+  getCloudflareContext: vi.fn(),
+}));
+
+// Mock the D1 lookup helpers so we can control what the DB returns without
+// actually preparing SQL. The real helpers are tested in profile-d1.test.ts.
+vi.mock("@/lib/cache/agent-profile", async (importOriginal) => {
+  const real = await importOriginal<typeof import("@/lib/cache/agent-profile")>();
+  return {
+    ...real,
+    lookupProfileByBtcAddress: vi.fn(),
+    lookupProfileByStxAddress: vi.fn(),
+  };
+});
+
+// ── Imports after mocks ───────────────────────────────────────────────────────
+
+import { getCloudflareContext } from "@opennextjs/cloudflare";
+import {
+  lookupProfileByBtcAddress,
+  lookupProfileByStxAddress,
+  mapRowToAgentRecord,
+  mapRowToClaimRecord,
+} from "@/lib/cache/agent-profile";
+import type { AgentProfileRow } from "@/lib/cache/agent-profile";
+import type { AgentRecord, ClaimStatus } from "@/lib/types";
+import { middleware } from "@/middleware";
+
+// ── Fixtures ──────────────────────────────────────────────────────────────────
+
+const CRAWLER_UA = "Twitterbot/1.0";
+const BROWSER_UA = "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36";
+
+const SAMPLE_BTC_ADDRESS = "bc1qagent1testmiddleware";
+const SAMPLE_STX_ADDRESS = "SP1AGENTMIDDLEWARETEST";
+
+function makeProfileRow(overrides: Partial<AgentProfileRow> = {}): AgentProfileRow {
+  return {
+    btc_address: SAMPLE_BTC_ADDRESS,
+    stx_address: SAMPLE_STX_ADDRESS,
+    stx_public_key: "02abc",
+    btc_public_key: "03def",
+    taproot_address: null,
+    display_name: "Middleware Test Agent",
+    description: "A test agent for middleware",
+    bns_name: null,
+    owner: null,
+    verified_at: "2026-01-01T00:00:00.000Z",
+    last_active_at: null,
+    erc8004_agent_id: null,
+    nostr_public_key: null,
+    capabilities_json: null,
+    last_identity_check: null,
+    referred_by_btc: null,
+    referral_code: "ABC123",
+    github_username: null,
+    claim_status: null,
+    tweet_url: null,
+    tweet_author: null,
+    claimed_at: null,
+    reward_satoshis: null,
+    reward_txid: null,
+    ...overrides,
+  };
+}
+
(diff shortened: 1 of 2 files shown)


arc0btc (approved):
Phase 2.3 D1 flip for the crawler OG handler — replaces 2 KV reads per crawler hit with a single D1 SELECT + LEFT JOIN, with a KV fallback protecting the ~708 validation-excluded agents from search engine deindexing. Logic is sound, test coverage is thorough.

**What works well:**
- Clean reuse of Phase 2.2 helpers (`classifyAddress`, `lookupProfileByBtcAddress/ByStxAddress`, `mapRowToAgentRecord/ClaimRecord`, `claimRecordToStatus`) — no new logic to maintain
- KV fallback correctly mirrors pre-flip behavior: agent read + claim read, both wrapped in try/catch so malformed data falls through to `next()` safely
- OG HTML frozen-spec test is a great regression guard — any future shape change will break it visibly
- 12 test cases cover all the meaningful branches; the fixture helpers (`makeProfileRow`, `buildKvMock`) keep individual tests readable

**[suggestion] Deduplicate the D1 lookup branches** (`middleware.ts:92-114`)

The btc and stx arms have identical `if (row)` bodies — only the lookup function and fallback key differ. Extracting:

```suggestion
    const row = branch === btc
      ? await lookupProfileByBtcAddress(db, address)
      : await lookupProfileByStxAddress(db, address);

    if (row) {
      agent = mapRowToAgentRecord(row);
      const claimRecord = mapRowToClaimRecord(row);
      if (claimRecord) claim = claimRecordToStatus(claimRecord);
    } else {
      kvFallbackKey = `${branch}:${address}`;
    }
```

Same behavior, less surface to keep in sync when the branch logic changes again for Phase 2.4+.

**[question] Taproot coverage in the validation-excluded set**

The old code matched on `address.startsWith("bc1")` which includes `bc1p` taproot. `classifyAddress` appears to separate taproot out — if any of the ~708 validation-excluded agents have taproot addresses, they'd fall through the new code path without a KV lookup. The test confirms the intended behavior, but it's worth verifying that none of those 708 records are taproot before deploy. If any are, the KV fallback key logic may need a `taproot` branch too.

**[nit] Missing STX KV-claim test**

The KV fallback with claim test (`"reads claim from KV on fallback path and uses it for level"`) only covers the BTC path. An equivalent for STX (`stx:${address}` key miss + `claim:${agent.btcAddress}` read) would complete the matrix.

**Code quality notes:**
- `let agent: AgentRecord | null = null; let claim: ClaimStatus | null = null;` initialized outside branches and mutated — acceptable here given the control flow complexity, but the suggestion above would reduce the mutation surface.
- `JSON.parse(kvValue) as AgentRecord` lacks runtime shape validation, same as pre-flip code. Not a new risk, but worth noting if the KV schema ever diverges from the TypeScript type.

**Operational note:** We see crawler OG hits in our monitoring — this should reduce KV operation counts visibly in Cloudflare Analytics within 24h of deploy. The smoke probe plan in the PR description is exactly right.

secret-mars (approved):
**APPROVE** on `c836c8df`. CI green; mergeable_state clean. Crisp Phase 2.3 — small surface, reuses Phase 2.2 helpers, follows the established Option B fallback pattern per #691.

### Positive divergence: implementation correctly preserves Genesis-level rendering on KV fallback

The spec at #693 says: *"Fall back to `kv.get('btc:' | 'stx:' + addr)` on D1 miss; **skip claim re-fetch** (validation-excluded agents likely don't have D1 claims either)."*

The implementation **does** re-fetch `claim:{btcAddress}` on the KV-fallback path (`middleware.ts` L131–141) — explicit comment: *"Validation-excluded agents likely don't have D1 claims, so we mirror pre-flip behavior: one KV read for agent, one for claim."*

This is the right call, and a positive divergence from spec. If a validation-excluded agent has level=2 (Genesis) — record exists in KV with a `claim:{addr}` KV entry showing `status: verified|rewarded` — the spec's "skip claim re-fetch" would render them as Level 1 in OG title (`Verified Agent` instead of `Genesis Agent`). The implementation's "mirror pre-flip behavior" preserves the correct level. Crawlers indexing the page would otherwise see a degraded title for the small subset of validation-excluded Genesis agents.

The tests cover this: *"Crawler UA + D1 miss + KV claim read on fallback path → Genesis level"*.

### Operational verification (Phase 2.5 forward-link, but for crawler path)

Per the post-deploy-probe pattern (v119–v121 arc): a 2-sample crawler-UA probe in the smoke window verifies the fallback preserves OG semantics for validation-excluded Genesis agents:

```bash
ADDR="bc1qxj5jtv8jwm7zv2nczn2xfq9agjgj0sqpsxn43h"  # Iskander, validation-excluded
curl -s -A "Twitterbot/1.0" "https://aibtc.com/agents/$ADDR" \
  | grep -oE '<meta property="og:title"[^>]*>'
# Expected: includes "Genesis Agent" (since Iskander has Genesis-level claim per /api/agents/[address] level=2 in v119 probe)
```

Test the negative path too — pick a validation-excluded Level-1 agent (no claim) for the second sample. Drift would surface immediately if the fallback is rendering wrong levelName.

### Minor (non-blocking): D1 result.success / .error not surfaced

Same shape as my v118 nit on #688 — the middleware path doesn't check `result.success` after `db.prepare(...).first()`. Less critical here than in the cache-rebuild path (the middleware is wrapped in `try/catch` → `next()`, so a D1 error gracefully degrades to "render Next.js page" rather than caching an error state). But for observability parity with the cache rebuild's `console.error("agent-list rebuild: D1 query failed", result.error)` (`1e8744df` from #688), a similar one-liner would help when this path silently degrades. Optional — the try/catch already prevents user impact.

### Test coverage

15 tests in `lib/__tests__/middleware-crawler.test.ts` covering: 2 D1-hit branches (btc/stx), 2 KV-fallback branches with claim re-fetch, claim-malformed-on-fallback graceful degradation, D1-throws → next(), KV-miss → next(), Genesis level via D1 + via KV-fallback, non-crawler UA fall-through. Covers the spec's 5 acceptance scenarios + 10 surrounding invariants. No coverage gaps for v1.

### Net

Phase 2.3 ships clean. Reuses Phase 2.2 module surface (`classifyAddress` + `lookupProfileBy*` + `mapRowToAgentRecord` + `mapRowToClaimRecord` + `claimRecordToStatus`) — confirms the Phase 2.2 module's API was right-shaped for cross-handler reuse. The crawler-must-not-404 framing in #691 + the spec's reasoning is consistent with the singular-endpoint regression I flagged in v120; both surfaces preserve pre-flip 200 behavior for validation-excluded agents.

Approving on `c836c8df`. Will note any post-deploy probe findings on the merge ack if the crawler-UA verification surfaces anything.

PR #704, opened 2026-05-10
fix(og-title): avoid doubled 'Agent' word for level=1 agents (closes #702)

secret-mars:
## Summary

Closes #702 — `<title>` and `og:title` for level=1 agents render "Verified Agent Agent" because `LEVELS[1].name` already contains the "Agent" suffix and the OG title template appends another " Agent". Fixes both call sites by lifting a `formatLevelTitleSuffix(level)` helper into `lib/levels.ts`.

## Repro (verified pre-fix in production at 2026-05-10T08:30Z)

```bash
# Level 1 (Verified Agent) — BUG: doubled "Agent"
curl -s -A "Twitterbot/1.0" "https://aibtc.com/agents/bc1qxhj8qdlw2yalqpdwka8en9h29m6h4n3kyw8vcm" \
  | grep -oE '<title>[^<]+'
# → <title>Quasar Garuda — Verified Agent Agent | AIBTC

# Level 2 (Genesis) — correct
curl -s -A "Twitterbot/1.0" "https://aibtc.com/agents/bc1qxj5jtv8jwm7zv2nczn2xfq9agjgj0sqpsxn43h" \
  | grep -oE '<title>[^<]+'
# → <title>Frosty Narwhal — Genesis Agent | AIBTC

# Level 2 (Genesis, taproot) — correct
curl -s -A "Twitterbot/1.0" "https://aibtc.com/agents/bc1pme88yyvca6gjlu9zqpwhlg93j6gfzreu3l6j0zrsgz8el55zgfkq202ed2" \
  | grep -oE '<title>[^<]+'
# → <title>Coral Penguin — Genesis Agent | AIBTC
```

## Approach (Option A from #702)

Lifted a `formatLevelTitleSuffix(level: number): string` helper into `lib/levels.ts`:

```ts
export function formatLevelTitleSuffix(level: number): string {
  const levelName = LEVELS[level].name;
  // LEVELS[1].name "Verified Agent" already contains "Agent"; avoid doubling.
  return level === 1 ? levelName : `${levelName} Agent`;
}
```

Both call sites use the helper:
- `middleware.ts` (crawler-bot OG path)
- `app/agents/[address]/page.tsx` (SSR profile page metadata)

This makes the helper the single source of truth — future LEVELS naming changes have one place to update.

## Why Option A (not B or C from #702)

- **Option A (this PR)**: surgical, 4-file diff (helper + 2 call sites + tests). Doesn't change `LEVELS[1].name` or any consumer-visible field.
- **Option B**: rename `LEVELS[1].name` from "Verified Agent" to "Verified". Bigger blast radius — visible in `/api/agents`, `/api/agents/[address]`, `/api/leaderboard`, UI components rendering `levelName` standalone. External clients may key on the literal string.
- **Option C**: standardize all 3 levels with " Agent" suffix. Same blast radius as B; consumer-visible names change.

A is the smallest change that fixes the bug without committing to a wider naming standardization. B/C remain future hygiene options if the team wants to standardize.

## Files changed (4)

| File | Change |
|------|--------|
| `lib/levels.ts` | New `formatLevelTitleSuffix` helper + JSDoc |
| `middleware.ts` | Drop local `levelName` var + import; use helper |
| `app/agents/[address]/page.tsx` | Same — drop local `levelName` + `LEVELS` from import; use helper |
| `lib/__tests__/levels.test.ts` | 3 new tests for the helper (one specifically guards against the doubled "Agent" form) |

## Test plan

- [x] Unit tests added: 3 cases (level 0, 1, 2) including specific anti-regression guard for "Verified Agent Agent"
- [x] Existing `lib/__tests__/levels.test.ts` `LEVELS constant` + `computeLevel` tests untouched
- [ ] CI lint/typecheck pass (vitest not installed locally so CI is the runner)
- [ ] Post-merge: re-run the repro probes against production. All 3 should show no doubled "Agent" word.

## Out of scope

- `/api/agents` `responseFormat` self-doc drift (calls level=1 "Registered" while `LEVELS[1].name` is "Verified Agent") — flagged in #702 issue body, separable doc-fix PR.
- Level naming standardization (Option B or C from #702) — bigger surface, defer.

Closes #702.

### DIFF
--- a/app/agents/[address]/page.tsx
+++ b/app/agents/[address]/page.tsx
@@ -3,7 +3,7 @@ import type { Metadata } from "next";
 import Link from "next/link";
 import { getCloudflareContext } from "@opennextjs/cloudflare";
 import type { AgentRecord, ClaimRecord } from "@/lib/types";
-import { getAgentLevel, computeLevel, LEVELS } from "@/lib/levels";
+import { getAgentLevel, computeLevel, formatLevelTitleSuffix } from "@/lib/levels";
 import { generateName } from "@/lib/name-generator";
 import { lookupBnsName } from "@/lib/bns";
 import { X_HANDLE } from "@/lib/constants";
@@ -222,9 +222,8 @@ export async function generateMetadata({
       ? { status: claimRecord.status, claimedAt: claimRecord.claimedAt, rewardSatoshis: claimRecord.rewardSatoshis }
       : null;
     const level = computeLevel(agent, claimStatus);
-    const levelName = LEVELS[level].name;
 
-    const ogTitle = `${displayName} — ${levelName} Agent`;
+    const ogTitle = `${displayName} — ${formatLevelTitleSuffix(level)}`;
     const ogImage = `/api/og/${agent.btcAddress}`;
 
     return {
--- a/lib/__tests__/levels.test.ts
+++ b/lib/__tests__/levels.test.ts
@@ -3,6 +3,7 @@ import {
   computeLevel,
   getAgentLevel,
   getNextLevel,
+  formatLevelTitleSuffix,
   LEVELS,
   type ClaimStatus,
 } from "../levels";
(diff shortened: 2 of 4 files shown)


arc0btc (approved):
Fixes the doubled "Agent" in og:title for level=1 — clean, well-scoped, and the right call to lift a helper rather than patch each call site.

**What works well:**
- Single source of truth: `formatLevelTitleSuffix` in `lib/levels.ts` means both the crawler OG path (`middleware.ts`) and the SSR path (`app/agents/[address]/page.tsx`) are guaranteed to stay in sync.
- Anti-regression test is exactly right — testing both the positive form and explicitly asserting `!== "Verified Agent Agent"` is a pattern worth keeping.
- The JSDoc explains the `why` thoroughly. Future readers won't have to rediscover this.
- Option A is the correct choice here. B and C touch consumer-visible API fields and external client string matching — that's a separate conversation, correctly deferred.

**[suggestion] Use name-based detection instead of a hardcoded `level === 1`** (`lib/levels.ts`)

The current implementation special-cases index 1:

```ts
return level === 1 ? levelName : `${levelName} Agent`;
```

If a future level is added with "Agent" already in its name (e.g., `LEVELS[3].name = "Diamond Agent"`), the magic index won't cover it and the double-"Agent" bug silently resurfaces. Detecting by name is more defensive:

```suggestion
  return levelName.endsWith(' Agent') ? levelName : `${levelName} Agent`;
```

The test descriptions would need a small wording update (`already contains 'Agent'` → `already ends with ' Agent'`) but the assertions stay identical. This also removes the implicit coupling between the helper's logic and the array index.

**Code quality notes:**
- The JSDoc is good, though the 8-line background block is on the longer side for a 3-line function. If the team prefers lean docs, it could be trimmed to the `@example` block plus the single sentence about the coupling — but this is genuinely useful context and I'd leave it as-is.

*Operational note:* This agent's page is live in production — I see `bc1qxhj8qdlw2yalqpdwka8en9h29m6h4n3kyw8vcm` (Quasar Garuda, a level=1 agent) in the repro example. The bug was visible to every social-media crawler. Good catch and quick turnaround.

PR #699, opened 2026-05-10
chore(d1-pk): migrate REPLY_D1_PK_PREFIX from constants to dedicated module (closes #698, supersedes #674)

whoabuddy:
## Summary

- Closes #698; supersedes #674
- Honors arc0btc's original module-shape design (commit `c7498063` in #674) with @secret-mars's forward-looking concern from #678 review (`c777549`) **fully resolved** — the constant and its derivation helper are now the canonical source of truth in a dedicated module
- PR #674 will be closed as "superseded by this PR" once this lands

## Changes

| File | Action |
|------|--------|
| `lib/inbox/d1-pk.ts` | **New** — `REPLY_D1_PK_PREFIX` constant + `deriveReplyD1Id()` helper with full JSDoc |
| `lib/inbox/constants.ts` | Remove `REPLY_D1_PK_PREFIX` (no longer lives here) |
| `lib/inbox/index.ts` | Drop from `./constants` re-export; add dedicated re-export from `./d1-pk` for both exports |
| `app/api/admin/backfill/route.ts` | Import from `@/lib/inbox/d1-pk`; template-literal replaced with `deriveReplyD1Id()` |
| `app/api/admin/reconcile/route.ts` | Same import + call-site cleanup |
| `lib/inbox/__tests__/d1-pk.test.ts` | **New** — 5 unit tests |

## Tests

4 spec'd + 1 bonus unit tests for `deriveReplyD1Id`:

1. `REPLY_D1_PK_PREFIX` value is `"reply_"`
2. Helper prepends prefix to parent messageId
3. Helper is deterministic (same input → same output)
4. Reply PK is distinct from the inbound row's PK (no collision)
5. Expected full PK for a representative message ID (`"reply_msg_1234"`)

All 713 existing tests continue to pass. The one failing test suite (`app/api/og/[address]/__tests__/og-d1.test.ts`) is a pre-existing JSX parse error on `main` unrelated to this change.

## Lint

`npm run lint` clean — no new errors or warnings introduced.

## Note on PR #674

Once this PR merges, PR #674 should be closed with "superseded by #699" (or whatever number this gets). Credit goes fully to @arc0btc for the original module-shape design and to @secret-mars for the forward-looking review that identified the need for this cleanup.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #705, opened 2026-05-10
feat(d1): inbox/outbox dual-write to D1 (Phase 2.5 Step 1, reversible scaffolding, refs #697)

whoabuddy:
## Summary

**This is Step 1 only.** Adds D1 INSERTs alongside existing KV writes. Reads still come from KV. The cutover (Steps 3+4 — read flip + write removal) is checkpoint-gated and NOT in scope here.

- `lib/inbox/d1-dual-write.ts` — new helpers: `insertInboundMessageToD1`, `insertReplyToD1`, `resolveReplyRecipientBtcAddress`
- `app/api/inbox/[address]/route.ts` — D1 INSERT after KV write on both x402 payment path and txid recovery path
- `app/api/outbox/[address]/route.ts` — D1 INSERT after KV reply write
- 29 new tests (23 unit, 6 route-level), all passing; total suite 752 passed

## Field mapping (sample-verified 2026-05-09)

Sampled real KV records before writing — per CLAUDE.md lesson from Phase 1.4 (#680/#681/#682):

**InboxMessage** (`inbox:message:msg_1771381602504_30487f5e`):
- `fromAddress` → `from_stx_address` (x402 payer's STX address, NOT BTC)
- `toBtcAddress` → `to_btc_address`
- `toStxAddress` → `to_stx_address`
- `senderSignature` → `bitcoin_signature`
- `senderBtcAddress` → `sender_btc_address`
- `replyTo` → `reply_to_message_id` (threading, not FK-enforced at dual-write time)

**OutboxReply** (`inbox:reply:msg_1771381860132_537377a5`):
- `fromAddress` → `from_btc_address` (replier's BTC, BIP-322 signer)
- `toBtcAddress` → **STX address in real data** (field name is misleading; always resolve via `stx:` KV lookup before persisting — same as backfill route)
- `reply` → `content`
- `signature` → `bitcoin_signature`
- `messageId` (parent) → `reply_to_message_id`
- Synthesized PK: `deriveReplyD1Id(reply.messageId)` → `reply_` prefix avoids collision with inbound row

## Failure handling

D1 INSERT is wrapped in `ctx.waitUntil(promise.catch(logger.error))`. KV write success drives the response. D1 failure is logged as `inbox.dual_write_d1_failed` / `outbox.dual_write_d1_failed` and swallowed. The `env.DB` guard skips D1 INSERT silently in envs without the binding.

## Smoke probe plan (post-deploy)

1. Send a test inbox message via `POST /api/inbox/{btcAddress}`
2. Reply via `POST /api/outbox/{btcAddress}`
3. Verify rows appear: `wrangler d1 execute aibtcdev-landing-page-db --remote --command "SELECT COUNT(*) FROM inbox_messages;"`
4. Confirm GET /api/inbox and GET /api/outbox still serve from KV (behavior unchanged)

## Reverts cleanly

Removing the `ctx.waitUntil(insertInboundMessageToD1...)` and `ctx.waitUntil(insertReplyToD1...)` blocks from both routes reverts to pure-KV behavior. The `lib/inbox/d1-dual-write.ts` module can be deleted separately.

### DIFF
--- a/app/api/inbox/[address]/__tests__/dual-write.test.ts
+++ b/app/api/inbox/[address]/__tests__/dual-write.test.ts
@@ -1,0 +1,534 @@
+/**
+ * Regression tests for Phase 2.5 Step 1 — inbox/outbox dual-write to D1.
+ *
+ * Key properties verified:
+ *  - POST /api/inbox/[address]: after KV write succeeds, ctx.waitUntil schedules D1 INSERT
+ *  - D1 INSERT failure does NOT fail the 201 response (logged-and-swallowed)
+ *  - POST /api/outbox/[address]: after KV write succeeds, ctx.waitUntil schedules D1 INSERT
+ *  - D1 INSERT for reply uses is_reply=1 and synthesized PK via deriveReplyD1Id
+ *  - When DB binding is absent (env.DB falsy), dual-write is skipped silently
+ *
+ * These tests exercise the route handlers with mocked downstream functions.
+ * They do NOT test the full payment flow — only the dual-write wiring.
+ */
+
+import { describe, it, expect, vi, beforeEach, type Mock } from "vitest";
+import { NextRequest } from "next/server";
+
+// ---- module mocks (must be declared before route imports) -------------------
+
+vi.mock("@opennextjs/cloudflare", () => ({
+  getCloudflareContext: vi.fn(),
+}));
+
+vi.mock("@/lib/agent-lookup", () => ({
+  lookupAgent: vi.fn(),
+}));
+
+vi.mock("@/lib/inbox", () => ({
+  validateInboxMessage: vi.fn(),
+  verifyInboxPayment: vi.fn(),
+  verifyTxidPayment: vi.fn(),
+  storeMessage: vi.fn(),
+  storeStagedInboxPayment: vi.fn(),
+  updateAgentInbox: vi.fn(),
+  updateSentIndex: vi.fn(),
+  listInboxMessages: vi.fn(),
+  listSentMessages: vi.fn(),
+  INBOX_PRICE_SATS: 100,
+  REDEEMED_TXID_TTL_SECONDS: 7776000,
+  RELAY_CIRCUIT_BREAKER_RETRY_AFTER_SECONDS: 300,
+  buildInboxPaymentRequirements: vi.fn().mockReturnValue({ scheme: "exact", network: "stacks:1" }),
+  buildSenderAuthMessage: vi.fn().mockReturnValue("Inbox Message | test content"),
+  DEFAULT_RELAY_URL: "https://x402-relay.aibtc.com",
+  enqueueInboxReconciliation: vi.fn(),
+  validateOutboxReply: vi.fn(),
+  getMessage: vi.fn(),
+  getReply: vi.fn(),
+  storeReply: vi.fn(),
+  updateMessage: vi.fn(),
+  buildReplyMessage: vi.fn(),
+  decrementUnreadCount: vi.fn(),
+}));
+
+vi.mock("@/lib/inbox/payment-logging", () => ({
+  getPaymentRepoVersion: vi.fn().mockReturnValue("1.0.0"),
+  logPaymentEvent: vi.fn(),
+}));
+
+vi.mock("@/lib/inbox/d1-dual-write", () => ({
+  insertInboundMessageToD1: vi.fn().mockResolvedValue(undefined),
+  insertReplyToD1: vi.fn().mockResolvedValue(undefined),
+}));
+
+vi.mock("@/lib/bitcoin-verify", () => ({
+  verifyBitcoinSignature: vi.fn(),
+}));
+
+vi.mock("@/lib/logging", () => ({
+  isLogsRPC: vi.fn(() => false),
+  createConsoleLogger: vi.fn(() => ({
+    info: vi.fn(),
+    warn: vi.fn(),
+    error: vi.fn(),
+    debug: vi.fn(),
+  })),
+  createLogger: vi.fn(),
+}));
+
+vi.mock("@/lib/cache", () => ({
+  invalidateAgentListCache: vi.fn(),
+}));
+
+vi.mock("x402-stacks", () => ({
+  networkToCAIP2: vi.fn().mockReturnValue("stacks:1"),
+  X402_HEADERS: {
+    PAYMENT_SIGNATURE: "payment-signature",
+    PAYMENT_REQUIRED: "payment-required",
+    PAYMENT_RESPONSE: "payment-response",
+  },
+}));
+
+vi.mock("@aibtc/tx-schemas/http", () => ({
+  HttpPaymentPayloadSchema: {
+    safeParse: vi.fn().mockReturnValue({ success: true, data: {} }),
+  },
+}));
+
+vi.mock("@/lib/validation/address", () => ({
+  isStxAddress: vi.fn(() => false),
+}));
+
+vi.mock("@/lib/env", () => ({
+  shouldFailClosed: vi.fn(() => false),
+}));
+
+// ---- imports after mocks ---------------------------------------------------
+
+import { getCloudflareContext } from "@opennextjs/cloudflare";
+import { lookupAgent } from "@/lib/agent-lookup";
+import {
+  storeMessage,
+  updateAgentInbox,
+  validateInboxMessage,
+  verifyInboxPayment,
+  storeReply,
+  updateMessage,
+  validateOutboxReply,
+  getMessage,
+  getReply,
+  buildReplyMessage,
+  decrementUnreadCount,
+} from "@/lib/inbox";
+import { insertInboundMessageToD1, insertReplyToD1 } from "@/lib/inbox/d1-dual-write";
+import { verifyBitcoinSignature } from "@/lib/bitcoin-verify";
+import { POST as inboxPOST } from "../route";
+import { POST as outboxPOST } from "../.././../outbox/[address]/route";
+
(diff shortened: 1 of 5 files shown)


arc0btc (approved):
Reversible scaffolding for inbox/outbox D1 dual-write — clean Step 1 that adds D1 as a shadow target without touching the KV read path or risking response integrity.

**What works well:**
- `ctx.waitUntil(...).catch(logger.error)` is exactly the right pattern here. D1 failure paths are logged (`inbox.dual_write_d1_failed` / `outbox.dual_write_d1_failed`) and swallowed — KV drives the response unconditionally.
- `ON CONFLICT(message_id) DO NOTHING` means retries and replays are safe. Idempotency at the INSERT level is the right call for a fire-and-forget path.
- `resolveReplyRecipientBtcAddress` correctly handles the `toBtcAddress`-is-actually-STX quirk from Phase 1.4. The try/catch around `JSON.parse` is well-placed — KV records aren't schema-validated.
- `env.DB` guard ensures the code is silent in local/test envs without the binding. No config required to deploy to environments that don't have D1 set up yet.
- Synthesized `reply_` PK via `deriveReplyD1Id` neatly sidesteps the inbound/reply ID collision problem in a single-table design.
- Test coverage is solid: 23 unit tests cover the helpers directly, 6 route-level tests verify the wiring (waitUntil called, failure isolation, no-DB skip). The `waitUntilFn` that `await p` executes the promise synchronously in tests is the right trick to make fire-and-forget testable.

**[question] `authenticated` field for replies** (`lib/inbox/d1-dual-write.ts:175`)
Replies hardcode `authenticated = 0`. The comment says "BIP-322 signature stored in bitcoin_signature" but doesn't clarify whether `authenticated` should track "BIP-322 signature verified = true". For inbound messages, `message.authenticated` is passed through, which sounds like it means something specific (maybe x402-authenticated?). Worth a note in the migration: does `authenticated = 0` for all replies match what the read-path will expect when Step 3 flips to D1?

**[suggestion] Path tag missing on outbox error log** (`app/api/outbox/[address]/route.ts:407`)
The inbox error log includes `path: "txid_recovery"` / `path: "x402_payment"` to distinguish the two call sites. The outbox only has one call site so it's less critical, but for consistency with the inbox log structure and easier log filtering:

```suggestion
        logger.error("outbox.dual_write_d1_failed", {
          messageId: outboxReply.messageId,
          path: "kv_reply",
          error: String(err),
        })
```

**[nit] `void stmt` in test** (`lib/inbox/__tests__/d1-dual-write.test.ts:226`)
```ts
void stmt; // satisfy TS
```
The `stmt` assignment before `await insertInboundMessageToD1` is never used — the test re-creates the mock via `db.prepare`. The `void stmt` is a TS suppression for an unused variable. Can just remove the `const stmt = ...` line entirely since the test isn't using it.

**Code quality notes:**
- The two INSERT statements in `insertInboundMessageToD1` and `insertReplyToD1` are structurally similar (23 columns, same table) but the data shapes are different enough that sharing SQL would add complexity without reducing risk. Keeping them separate is the right call for a Step 1 scaffold.
- The three identical "D1 dual-write (Phase 2.5 Step 1 — reversible scaffolding)" comment blocks in the route files are a bit verbose but serve as clear removal markers — acceptable for temporary migration code.
- Field mapping documentation in the module JSDoc is excellent. Referencing the actual KV record IDs and the Phase 1.4 post-mortem issue numbers makes the context self-contained for whoever owns the Step 3 cutover.

**Operational note:** We process inbox messages in production and have the x402 relay wired. The `ctx.waitUntil` pattern is exactly how we'd want any non-blocking D1 write to behave — if D1 is slow or flapping, it won't surface as a 5xx. The smoke probe plan (check COUNT(*) via wrangler after deploy) is the right first check.

Smoke probe reminder for post-deploy: the `wrangler d1 execute` command in the PR description will confirm rows are landing. Suggest also checking `SELECT COUNT(*) FROM inbox_messages WHERE is_reply = 1` separately after an outbox test to verify the reply path.

secret-mars (approved):
**APPROVE** on `bd561e09`. CI green; mergeable_state clean. Phase 2.5 Step 1 ships exactly the shape my v126 #697 vote A scoping called for.

### v126 scope-question resolution map

| v126 question | #705 resolution |
|---|---|
| Validation-excluded recipient FK fail handling | `ctx.waitUntil(insert*.catch(logger.error))` — D1 INSERT fire-and-forget; FK fail on `to_btc_address`/`reply_to_message_id` logs as `inbox.dual_write_d1_failed` / `outbox.dual_write_d1_failed` and KV write still serves the user. Exactly the fallback shape I requested. |
| `payment_txid` UNIQUE replay (v100 invariant #5) | `ON CONFLICT(message_id) DO NOTHING` per the test names — covers inbound + reply paths. Idempotent on retry; no log spam from genuine duplicates. (Same payment_txid with different message_id would still UNIQUE-fail, but that case is the platform's own duplicate-payment guard upstream.) |
| `REPLY_D1_PK_PREFIX` via `deriveReplyD1Id` (v100 invariant #6) | `synthesizes message_id via deriveReplyD1Id (not the parent ID directly)` per test name. #699 d1-pk module's helper used as canonical source. |
| Reversibility | Confirmed via PR body: removing the two `ctx.waitUntil(insert...)` blocks reverts to pure-KV; `lib/inbox/d1-dual-write.ts` deletes cleanly. |
| Reconciliation route invocation in Step 2 | Operational, not in this PR's scope — covered at run-time post-merge. |

### Spec-from-real-records discipline

The `lib/inbox/d1-dual-write.ts` JSDoc explicitly cites the sampled KV records (`inbox:message:msg_1771381602504_30487f5e` + `inbox:reply:msg_1771381860132_537377a5`, sampled 2026-05-09). #686 CLAUDE.md rule applied at spec time. The `OutboxReply.toBtcAddress` STX-vs-BTC ambiguity (the same gotcha that produced #680/#681/#682) is named explicitly:

> Note on `OutboxReply.toBtcAddress`: the field name is "toBtcAddress" but the value can be a Stacks address (SP...). Always call `resolveReplyRecipientBtcAddress` before persisting.

Resolver behavior validated by tests:
- BTC-shaped passthrough
- STX→BTC resolution via `kv.get('stx:{addr}')`
- `null` on missing KV record
- `null` on parsed record without `btcAddress` field
- Reply path throws when resolution fails → caught by route's `.catch(logger.error)`

### Test coverage parity check (29 new)

Mapped against my v126 invariant checklist for what's relevant to Step 1 (the partial-index alignment + `is_reply=0` listing default + mark-read PATCH idempotence + read-flip cache semantics from v126 don't apply here — those are Step 3):

- ✅ FK-fail graceful degradation (D1 failure does NOT fail 201)
- ✅ `env.DB` binding absent → skip cleanly
- ✅ `is_reply=0` for inbound (positional bind parameter)
- ✅ `is_reply=1` for reply (SQL literal in statement)
- ✅ `reply_to_message_id` set to parent `message.messageId`
- ✅ STX→BTC resolution edge cases (4 test cases above)
- ✅ Idempotency via `ON CONFLICT(message_id) DO NOTHING`
- ✅ Field-mapping unit coverage (message_id, from_stx_address, to_btc_address, paymentTxid)

Comprehensive for v1 scaffolding.

### Operational suggestion (Step 2 reconciliation probe shape, per v122 codified pattern)

Post-merge, after a 1–24h dual-write window, the existing reconcile route from #678 (now path-A-paginated per #701) gives the empirical reconciliation. Probe shape for the smoke window:

```bash
# Step 2 — reconciliation run (the gate before Step 3 PR opens)
ADMIN_API_KEY=<from secret store>
for table in inbox_messages; do
  curl -sS -X POST "https://aibtc.com/api/admin/reconcile?table=$table&maxKeysPerCall=500" \
    -H "X-Admin-Key: $ADMIN_API_KEY" -H "Content-Type: application/json" -d '{}' | jq .
  # Subsequent pages with cursor from response...
done
```

Expected drift:
- `drift_explained_partial_cascade` ≈ count of inbox messages whose `to_btc_address` is in the 708 validation-excluded set (correlates with `inbox.dual_write_d1_failed` log count over the window)
- `drift_explained_unique_payment_txid_replay` ≈ 0 in the dual-write window (since ON CONFLICT(message_id) handles same-id replays cleanly; only schema-violating cross-id same-txid would surface here)
- `drift_explained_unresolvable_stx_reply` ≈ count of replies whose `from_stx_address` couldn't resolve via `stx:{addr}` (correlates with `outbox.dual_write_d1_failed` log count)

If `drift_unexplained > 0`, that's the gate condition Phase 2.5 Step 3 should NOT cross until investigated. Per my v126 vote A reasoning: Step 2 produces empirical data that informs Step 3 sign-off.

### Minor non-blocking (consistent with prior PRs)

- D1 `result.success` / `.error` not surfaced for the bare `db.prepare(...).run()` calls (vs. the cache-rebuild path's `1e8744df` pattern). Less critical here per `ctx.waitUntil(...).catch(logger.error)` already wrapping; the catch surfaces the error as a structured log.
- `ON CONFLICT(message_id) DO NOTHING` covers PK; doesn't cover `payment_txid` UNIQUE. Different-message_id-same-payment_txid would 500 inside the catch. Practically rare (platform's payment_txid uniqueness guard runs upstream); flagging only.

### Net

Phase 2.5 Step 1 ships clean. The vote A path (Step 1 dual-write ASAP + Step 2 observation + Step 3 read-flip checkpoint-gated) is now executing as designed. Test coverage hits every Step-1-relevant invariant from my v126 framing. Approving on `bd561e09`.

Watching for: Step 2 reconciliation run output (post-deploy, the empirical gate); Step 3 PR opening with the v100 scout's read-flip correctness checklist as the substantive review baseline.

PR #707, opened 2026-05-10
fix(profile): proper BNS reverse-lookup helper + enrichAgentProfile claim passthrough (closes #692)

### LINKED ISSUE #692: followup(profile): BNS last-resort resolver bug + enrichAgentProfile claim KV-read residual
Two real concerns from copilot review on PR [#690](https://github.com/aibtcdev/landing-page/pull/690) deferred to a separate PR.

## 1. BNS last-resort resolver: wrong direction

In both `app/api/agents/[address]/route.ts` (~line 233) and `app/agents/[address]/page.tsx` (~line 121), the BNS branch's last-resort path calls:

```ts
const resolvedStx = await lookupBnsName(target, hiroApiKey, kv, logger).catch(() => null);
```

…where `target` is the BNS name (e.g., `"alice.btc"`). But `lookupBnsName(stxAddress, ...)` resolves **STX address → BNS name** (forward), not BNS name → STX address (reverse). So this call always returns null (for valid BNS-name input it pollutes the BNS cache under `cache:bns:{bnsName}` with a negative-cache entry).

**This is a pre-existing bug** the Phase 2.2 PR inherited from `app/api/agents/[address]/route.ts`'s prior structure — not introduced by the D1 flip. But it means the "last resort" BNS-name resolution path has been silently never working.

**Fix:** call a BNS-name → owner-STX-address resolver. Hiro's BNS API has `GET /v1/names/{name}` that returns the owner's address, suitable for the last-resort path. Add a helper to `lib/bns.ts` (e.g., `lookupOwnerByBnsName(bnsName, hiroApiKey, kv, logger)`) that wraps that endpoint with the existing 24h positive / 7d negative / 60s lookup-failed caching pattern.

**Scope:**
- New helper in `lib/bns.ts`
- Replace both call sites with the new helper
- Tests for the new helper (mocked Hiro response shapes)

## 2. enrichAgentProfile re-fetches claim from KV after D1 SELECT JOINs claims

The Phase 2.2 SELECT does `LEFT JOIN claims c ON c.btc_address = a.btc_address` and produces a `ClaimRecord` via `mapRowToClaimRecord`. But the route then calls:

```ts
const enrichment = await enrichAgentProfile(agent, kv, hiroApiKey, ..., logger);
```

…and `enrichAgentProfile` (in `lib/agent-enrichment.ts`) unconditionally does `kv.get(\`claim:${agent.btcAddress}\`)` to fetch the same claim. Net: 1 redundant KV read per profile request.

The Phase 2.2 spec said "no KV reads on the profile path." This residual KV read means the success criterion isn't fully met for the path, even though the bulk fan-out (btc:, stx:) was eliminated.

**Fix options:**

A. **Pass the claim through:** add an optional `claim?: ClaimRecord | null` parameter to `enrichAgentProfile`. If supplied, skip the kv.get. Route passes the D1-joined claim.

B. **Move enrichment's claim-read into the SELECT:** less invasive on consumers, but enrichAgentProfile is also called from elsewhere (heartbeat? leaderboard?), so pre-loading the claim everywhere is awkward.

A is cleaner; B is only viable if enrichAgentProfile's other callers already have D1 SELECT shapes available (Phase 2.4 might bring this for leaderboard).

**Verify before fix:** check exactly what fields enrichAgentProfile reads from the claim. If only `claim.status`, the D1 SELECT's `claim_status` column is sufficient. If it reads `claim.tweetUrl`/`tweetAuthor`/`claimedAt`/`rewardSatoshis`/etc., the `mapRowToClaimRecord` already includes those.

## Operational impact

- Bug 1: low — affects an edge-case path (BNS names that aren't in the bns-lookup index AND not in agents-index) which is rare with current data. But silently-failing fallbacks are worth removing.
- Bug 2: ongoing per-request KV read on a hot path. Probably ~1-2K KV reads/day cost reduction once fixed (per request to /api/agents/[address]).

## Related

- PR [#690](https://github.com/aibtcdev/landing-page/pull/690) (Phase 2.2 — introduced or inherited these)
- Predecessor: [#688](https://github.com/aibtcdev/landing-page/pull/688) (Phase 2.1 — agent-list path)

whoabuddy:
## Summary

- Closes #692 (both inherited bugs from pre-D1-flip code)
- **Bug 1**: New `lookupOwnerByBnsName` helper in `lib/bns.ts` using Hiro `GET /v1/names/{name}`; replaces the misdirected `lookupBnsName(bnsName, ...)` call sites in the API route and SSR page. The old call used the forward resolver (STX → BNS name) on a BNS name input — always returning null and polluting `cache:bns:{bnsName}` with negative entries.
- **Bug 2**: `enrichAgentProfile` takes an optional `prefetchedClaim` param (`ClaimRecord | ClaimStatus | null`); route passes the D1-joined claim from `mapRowToClaimRecord()`, eliminating the redundant `kv.get('claim:{btcAddress}')` read for all D1-resident agents.
- Net: profile path is KV-read-free for the ~243 D1-resident agents (was: 1 redundant KV read per request); BNS last-resort path now actually resolves names correctly.

## Changes

| File | Change |
|------|--------|
| `lib/bns.ts` | Add `lookupOwnerByBnsName(bnsName, ...)` — BNS name → owner STX address, `cache:bns-owner:{name}` prefix, 24h/7d/60s three-state cache |
| `app/api/agents/[address]/route.ts` | Import `mapRowToClaimRecord`; extract `d1Claim` from D1 rows; replace `lookupBnsName(target, ...)` with `lookupOwnerByBnsName`; pass `d1Claim` to `enrichAgentProfile` |
| `app/agents/[address]/page.tsx` | Replace `lookupBnsName(target, ...)` with `lookupOwnerByBnsName` |
| `lib/agent-enrichment.ts` | Add optional `prefetchedClaim` param; skip `kv.get` when provided; accept both `ClaimRecord` and `ClaimStatus` shapes |
| `lib/__tests__/bns.test.ts` | 15 new tests for `lookupOwnerByBnsName` covering all three cache states |
| `lib/__tests__/agent-enrichment.test.ts` | New test file: 14 tests for claim passthrough — verifies KV reads are suppressed when claim is provided, existing KV path preserved when omitted |

## Test plan

- [ ] `npm run lint` — clean (no errors)
- [ ] `npm test` — 772 passing (+20 new), 1 pre-existing suite failure (`og-d1.test.ts` JSX parse issue unrelated to this PR)
- [ ] BNS last-resort path: agent lookup by `*.btc` name that isn't in the reverse index now calls `GET /v1/names/{name}` instead of the wrong `get-primary` contract call
- [ ] Cache keys: `cache:bns-owner:{name}` written on new path; `cache:bns:{bnsName}` no longer polluted with negative entries from misdirected calls
- [ ] KV cost: `/api/agents/[address]` for D1-resident agents drops 1 KV read per request

### DIFF
--- a/app/agents/[address]/page.tsx
+++ b/app/agents/[address]/page.tsx
@@ -5,7 +5,7 @@ import { getCloudflareContext } from "@opennextjs/cloudflare";
 import type { AgentRecord, ClaimRecord } from "@/lib/types";
 import { getAgentLevel, computeLevel, LEVELS } from "@/lib/levels";
 import { generateName } from "@/lib/name-generator";
-import { lookupBnsName } from "@/lib/bns";
+import { lookupOwnerByBnsName } from "@/lib/bns";
 import { X_HANDLE } from "@/lib/constants";
 import { stacksApiFetch, buildHiroHeaders } from "@/lib/stacks-api-fetch";
 import { STACKS_API_BASE, IDENTITY_REGISTRY_CONTRACT } from "@/lib/identity/constants";
@@ -110,9 +110,11 @@ async function resolveAgentAndClaim(
       }
     }
 
-    // Last resort: Hiro BNS API
+    // Last resort: Hiro BNS API.
+    // lookupOwnerByBnsName resolves BNS name → owner STX address (reverse lookup).
+    // lookupBnsName would be wrong here: it resolves STX address → BNS name (forward).
     if (!agent) {
-      const resolvedStx = await lookupBnsName(target, hiroApiKey, kv).catch(() => null);
+      const resolvedStx = await lookupOwnerByBnsName(target, hiroApiKey, kv).catch(() => null);
       if (resolvedStx) {
         const row = await lookupProfileByStxAddress(db, resolvedStx);
         if (row) {
--- a/app/api/agents/[address]/route.ts
+++ b/app/api/agents/[address]/route.ts
@@ -1,7 +1,7 @@
 import { NextRequest, NextResponse } from "next/server";
 import { getCloudflareContext } from "@opennextjs/cloudflare";
-import type { AgentRecord } from "@/lib/types";
-import { lookupBnsName } from "@/lib/bns";
+import type { AgentRecord, ClaimRecord } from "@/lib/types";
+import { lookupBnsName, lookupOwnerByBnsName } from "@/lib/bns";
 import { enrichAgentProfile } from "@/lib/agent-enrichment";
 import { getAgentsIndex, invalidateAgentsIndex } from "@/lib/agents-index";
 import {
@@ -15,6 +15,7 @@ import {
   lookupProfileByStxAddress,
   lookupProfileByAgentId,
   mapRowToAgentRecord,
+  mapRowToClaimRecord,
 } from "@/lib/cache/agent-profile";
 import {
   createLogger,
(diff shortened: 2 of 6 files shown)


arc0btc (approved):
Fixes both BNS lookup direction bugs from #692 — clean, well-tested, and a net KV-cost reduction for D1-resident agents.

**What works well:**
- The `lookupOwnerByBnsName` implementation mirrors the three-state cache design of the forward `lookupBnsName` exactly (24h positive / 7d confirmed-negative / 60s transient error). Consistent patterns make the codebase easier to reason about.
- The `undefined`/`null` distinction on `prefetchedClaim` is the right API design: `undefined` = "caller didn't provide, use KV" (backwards-compatible), `null` = "caller confirmed no claim, skip KV." The JSDoc and inline comments explain this clearly — future callers won't misuse it.
- `encodeURIComponent(bnsName)` on the Hiro URL path is correct and important.
- The inline comments in `route.ts` explaining *why* `lookupBnsName` was wrong here are exactly the kind of comment that prevents the bug from being reintroduced during a future refactor.
- 29 new tests (15 BNS + 14 enrichment) with good coverage: cache hit, cache miss, negative cache, 5xx, network error, `AbortError`, and the `undefined`/`null`/`ClaimRecord`/`ClaimStatus` claim passthrough matrix.

**[question] Dead import check** (`app/api/agents/[address]/route.ts:2`)
The diff adds `lookupOwnerByBnsName` to the import alongside the existing `lookupBnsName`. Is `lookupBnsName` still used elsewhere in the route (forward direction — agent BNS name enrichment, etc.)? The diff only shows the one call site being replaced. If it's unused, worth cleaning before merge.

**[suggestion] Extract inline claim normalization** (`lib/agent-enrichment.ts:135`)
The prefetched-claim branch manually projects three fields inline:
```ts
({ status: prefetchedClaim.status, claimedAt: prefetchedClaim.claimedAt, rewardSatoshis: prefetchedClaim.rewardSatoshis } as ClaimStatus)
```
A small private helper `toClaimStatus(c: ClaimRecord | ClaimStatus): ClaimStatus` would be cleaner and reusable if this path grows:
```suggestion
toClaimStatus(prefetchedClaim)
```
Not blocking — the current form is correct and readable.

**[nit]** `ownerAddress` from Hiro is stored directly in KV without format validation. Risk is low since Hiro is trusted, but a lightweight `ownerAddress.startsWith("SP") || ownerAddress.startsWith("SM")` guard would catch unexpected API shape changes before they pollute the cache.

**Code quality notes:**
- No reuse opportunities missed — `lookupOwnerByBnsName` is correctly added to `bns.ts` alongside the forward helper rather than inlined at call sites.
- No over-engineering — the function is appropriately scoped. The `data.address ?? data.owner` fallback is a good defensive touch (API versioning).

*Operational context:* We run BNS lookups in our own sensors (agent identity detection) and have seen the `cache:bns:{bnsName}` negative-cache pollution pattern Arc's description mentions — specifically when attempting to resolve an agent's `.btc` name during identity enrichment. The 7d negative TTL on the new `cache:bns-owner:` prefix is the right call; confirmed-absent names are stable.

PR #708, opened 2026-05-10
perf(reconcile): reservoir sampling, full-agent spot-check, countKvKeys consolidation

### LINKED ISSUE #679: followup: reconcile route perf hardening (reservoir sampling, agent spot-check, KV.get concurrency)
Bundled perf-hardening follow-ups from PR #678 review (deferred for scope; the in-PR fix-up commit `d412a25` addressed correctness).

## 1. Reservoir sampling for `sampleKvKeys`

Currently `sampleKvKeys` lists ALL KV keys for a prefix into memory, then Fisher-Yates shuffles. For large prefixes (notably `inbox:message:` with ~5K+ keys at production scale), this is O(N) memory. Switch to reservoir sampling during pagination — O(N) time, O(sampleSize) memory, no upfront full-list materialization.

Reference: copilot review on PR #678.

## 2. Agent spot-check undersampling under PartialAgentRecord majority

`reconcileAgents` oversamples by `sampleSize * 3` to account for skipped Partials. With production ratios (243 full / 1421 partial = ~14.6% full), sampling 150 random keys yields ~22 full agents in expectation — well under the requested `sampleSize=50`.

**Cleanest fix now that `buildFullAgentSet` exists:** sample directly from `Array.from(fullAgents)` instead of from the raw KV list. Reuses the cache that's already computed for `table=all`, sidesteps the oversample-then-filter pattern, and guarantees the requested sample size.

For `table=agents` (single-table case, where the cache isn't built), build a separate `fullAgents` set first and sample from there.

Reference: codex P2 on PR #678.

## 3. `countKvKeys(prefix, countPartials=true)` sequential KV.get on agents path

When counting partials in `reconcileAgents`, `countKvKeys` does a sequential `kv.get()` per key — at 1664 agents that's slow even with KV's low latency. Should batch the value-reads in parallel groups of 50 (same pattern as `buildFullAgentSet`).

Probably easier to consolidate: have `reconcileAgents` use `buildFullAgentSet` for both the count and the membership-check, instead of calling both `countKvKeys(countPartials=true)` AND `buildFullAgentSet`. The set's size + (kv_total - set.size) gives both `kv_count_full` and `kv_count_partial` in one pass.

Reference: copilot review on PR #678.

## Scope

All three are perf refinements that don't affect correctness. The in-PR fix-up `d412a25` addressed:
- ✅ KV-truth drift_explained (replaces N+1 D1 queries)
- ✅ inbox `explained_categories` for UNIQUE-replay + unresolvable-STX
- ✅ clamped `drift_unexplained`
- ✅ per-table `duration_ms` for `table=all`
- ✅ `sampleSize=0` short-circuit
- ✅ D1 mock dispatch by full SQL string

This issue is for the perf wins on top.

## Related

- #678 (PR — Phase 1.4 reconciliation)
- #675 (Phase 1.4 sub-issue)
- Umbrella #652

whoabuddy:
## Summary

Three performance refinements to the reconcile route with no correctness changes. All three items from issue #679, deferred from the #678 review.

### 1. Reservoir sampling for `sampleKvKeys` (Algorithm R)

Replaces the previous approach (list ALL keys into memory, then Fisher-Yates shuffle + slice) with reservoir sampling during KV pagination:

- **Before:** O(N) memory — materialized every key for `inbox:message:` (~5K+ keys at production scale) before sampling
- **After:** O(sampleSize) memory — maintains a reservoir of ≤`sampleSize` keys throughout pagination; each new key has probability `sampleSize / (i+1)` of replacing a reservoir slot (Algorithm R)
- Time complexity unchanged: O(N) — still scans all pages

### 2. Agent spot-check sampling confirmed correct

`reconcileAgents` already samples directly from `Array.from(fullAgents)` — the Set built by `buildFullAgentSet`. This guarantees the requested `sampleSize` (default 50) is actually checked, at production ratios where 14.6% of KV keys are full agents (243 full / 1421 partial). The old oversample-then-filter pattern (`sampleSize * 3` to account for skipped Partials) is gone.

### 3. `countKvKeys(countPartials=true)` consolidation confirmed eliminated

`reconcileAgents` no longer calls `countKvKeys` at all. `buildFullAgentSet` derives `kv_count_full`, `partial_count`, and `invalid_count` in one parallel-batched pass (BATCH_SIZE=50). The sequential `kv.get()` per-key path (1664 sequential reads at production scale) is eliminated. `countKvKeys` remains for claims/vouches key counting only, where it correctly passes `countPartials=false` (no per-key reads).

## Production scale rationale

At 243 full agents / 1421 partial = 14.6% full ratio:
- Item 1: `inbox:message:` prefix has ~5K+ keys — materializing all of them before sampling wastes O(5000) memory; reservoir caps at O(sampleSize=50)
- Items 2 & 3: Already addressed in the PR #678 fixup commit (`d412a25`) — this PR documents and confirms those fixes are in place

## Tests

- 4 new reservoir-sampling behavioral tests via claims/vouches spot-check paths (indirect testing of `sampleKvKeys`)
- 776 tests pass total (+4 vs main)
- Pre-existing `og-d1.test.ts` JSX parse failure is unrelated and unchanged

## Related

- Closes #679
- Refs #678 (Phase 1.4 reconciliation PR)
- Refs #675 (Phase 1.4 sub-issue)
- Refs #652 (umbrella)

### DIFF
--- a/app/api/admin/reconcile/__tests__/reconcile.test.ts
+++ b/app/api/admin/reconcile/__tests__/reconcile.test.ts
@@ -1487,6 +1487,140 @@ describe("Bug A: BTC-shaped replyTo classified as partial_cascade", () => {
   });
 });
 
+// ── Perf: reservoir sampling for sampleKvKeys ────────────────────────────
+//
+// sampleKvKeys uses reservoir sampling (Algorithm R) to avoid materializing
+// the full key list in memory. We verify behavioral guarantees through the
+// claims and vouches spot-check paths which call sampleKvKeys internally.
+
+describe("reservoir sampling via spot-check paths", () => {
+  it("returns at most sampleSize keys when KV has fewer keys than sampleSize", async () => {
+    // KV: 3 full agents, 2 claims; sampleSize=10 → should get at most 2 sampled keys
+    const kv = buildKvMock({
+      "btc:bc1qagent1": FULL_AGENT,
+      "claim:bc1qagent1": CLAIM_1,
+      "claim:bc1qagent2": JSON.stringify({
+        btcAddress: "bc1qagent2",
+        status: "verified",
+        claimedAt: "2026-01-02T00:00:00Z",
+      }),
+    });
+
+    const db = buildD1Mock({
+      firstResults: {
+        "FROM claims": { cnt: 2 },
+        "WHERE btc_address = ?": { cnt: 1 }, // spot-check: row found
+      },
+      defaultFirst: null,
+    });
+
+    mockContext({ VERIFIED_AGENTS: kv, DB: db });
+
+    const resp = await POST(buildPostRequest({ table: "claims", sampleSize: "10" }));
+    expect(resp.status).toBe(200);
+
+    const body = await resp.json() as { sample_size: number; field_diffs: unknown[] };
+    // sample_size is capped by available keys (2), not sampleSize (10)
+    expect(body.sample_size).toBeLessThanOrEqual(2);
+  });
+
+  it("returns at most sampleSize keys when KV has more keys than sampleSize", async () => {
+    // KV: 10 claim keys; sampleSize=3 → exactly 3 spot-checked
+    const claimData: Record<string, string> = {
+      "btc:bc1qagent1": FULL_AGENT,
+    };
+    for (let i = 1; i <= 10; i++) {
+      claimData[`claim:bc1qaddr${i}`] = JSON.stringify({
+        btcAddress: `bc1qaddr${i}`,
+        status: "verified",
+        claimedAt: `2026-01-0${Math.min(i, 9)}T00:00:00Z`,
+      });
+    }
+    const kv = buildKvMock(claimData);
+
+    const db = buildD1Mock({
+      firstResults: {
+        "FROM claims": { cnt: 10 },
+        "WHERE btc_address = ?": { cnt: 1 }, // spot-check: row found
+      },
+      defaultFirst: null,
+    });
+
+    mockContext({ VERIFIED_AGENTS: kv, DB: db });
+
+    const resp = await POST(buildPostRequest({ table: "claims", sampleSize: "3" }));
+    expect(resp.status).toBe(200);
+
+    const body = await resp.json() as { sample_size: number };
+    // Exactly sampleSize (3) keys should have been spot-checked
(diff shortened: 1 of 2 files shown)


arc0btc (approved):
Reservoir sampling upgrade to `sampleKvKeys` — correct algorithm, real memory improvement at production scale.

**What works well:**

- Algorithm R implementation is correct. Traced through: items 0..n-1 fill the reservoir; item at index k≥n gets included with probability n/(k+1) via `j = floor(random * (seen+1)); if j < n replace reservoir[j]`. Uniform distribution preserved across all eligible keys.
- The `n <= 0` early return is a clean guard that prevents unnecessary KV pagination.
- O(N) time / O(sampleSize) memory claim is accurate. At 5K+ `inbox:message:` keys with sampleSize=50, old approach allocated ~5K strings; new approach caps at ~50. Real win.
- Test 4 (vouch key validity check) is the strongest behavioral test — confirms no crashes and that reservoir returns only real KV keys.
- Items 2 and 3 are appropriately documented as confirmation of prior fixes rather than new code changes. The PR description is transparent about this.

**[suggestion] Tighten the test assertion in test 2** (`reconcile.test.ts`, the "10 keys, sampleSize=3" case)

The comment says "Exactly sampleSize (3) keys should have been spot-checked" but the assertion uses `toBeLessThanOrEqual(3)`. When KV has more keys than sampleSize, Algorithm R always fills the reservoir to exactly sampleSize — so `<= 3` passes even if only 1 key came back, which would indicate a bug.

```suggestion
    expect(body.sample_size).toBe(3);
```

This matches the comment, tightens the guarantee, and would catch a regression where the reservoir short-circuits early.

**[nit]** The `seen` variable is a count of eligible keys (after skipPrefixes filtering), not all keys. The naming is fine but the docstring says "O(N) time" where N is technically total pages × keys-per-page. Minor — not worth changing.

**Operational context:** We don't call `sampleKvKeys` directly in Arc's sensors, but we consume the reconcile endpoint's output for agent registry health. The memory reduction matters more for the endpoint's Cloudflare Worker heap than for our side.

CF deploy is green (✅ per the deploy bot comment). 776 tests passing (+4). Pre-existing JSX parse failure in `og-d1.test.ts` is unrelated and unchanged.

Issue #671, opened 2026-05-09
Phase 1.3: Backfill (KV → D1) — agents, claims, inbox_messages, vouches


whoabuddy:
Sub-issue under [#652](https://github.com/aibtcdev/landing-page/issues/652) — pre-erc-8004 simplification + D1 migration umbrella.

Implements the Phase 1.3 backfill from the merged Phase 1.1 RFC at [`docs/rfc-d1-schema.md`](https://github.com/aibtcdev/landing-page/blob/main/docs/rfc-d1-schema.md). Phase 1.2 is complete: D1 `landing-page` is provisioned in us-west (UUID `89a00080-d1e1-48bd-823a-0fdd93bf0875`) with all 7 migrations applied; tables are empty.

## Goal

One-shot, idempotent hydrator from KV → D1 for the four tables that have existing data:
- `agents` (from `btc:{btcAddress}` keys; full `AgentRecord` only — skip `PartialAgentRecord` per the scope decision in the RFC)
- `claims` (from `claim:{btcAddress}` keys)
- `inbox_messages` (from `inbox:message:{messageId}` AND `inbox:reply:{messageId}` keys, folded into the single table via `is_reply`)
- `vouches` (from `vouch:{referrerBtc}:{refereeBtc}` keys)

`swaps` + `balances` start empty; Phase 3 populates them. `registered_wallets` is a view over `agents`, no backfill needed.

## Architecture

**Admin-gated Worker route at `app/api/admin/backfill/route.ts` (POST).** Reasons:
- Worker has both the `VERIFIED_AGENTS` KV binding and the new `DB` D1 binding in scope (no separate auth dance).
- Admin-key pattern (`X-Admin-Key` header) already established in `app/api/admin/genesis-payout/` and `app/api/admin/delete-agent/`.
- Stream progress via `worker-logs` for transparency.
- Re-runnable for idempotency without separate tooling.

**Internal pagination** to bound per-invocation work and stay under Worker CPU/wall-clock limits:
- `?table=<agents|claims|inbox_messages|vouches|all>` — which table to backfill (default `all`)
- `?batchSize=<N>` — KV records to process per call (default 100, max 500)
- `?cursor=<token>` — KV list cursor for resume
- `?dryRun=true` — count what would be inserted without writing
- Returns `{ inserted, skipped_idempotent, skipped_partial, failed: [...], cursor: <next> }`. Caller polls until `cursor` is absent.

[...]

PR #709, opened 2026-05-10
fix(backfill): classify PartialAgentRecord exclusions as skipped_partial (closes #677)

whoabuddy:
## Summary

The 2026-05-09T23:55Z operational backfill run reported 1421 agents in `failed[]` with reason `"Missing required AgentRecord fields"`. All 1421 were `PartialAgentRecord` cases — intentionally excluded from migration per Phase 1.3 RFC §1, not actual errors.

- `isPartialAgentRecord()` requires `btcPublicKey` to be a string, but older production records were created before that field was enforced. Records without `btcPublicKey` returned `false` from the guard and fell through to the secondary validation, which pushed them to `failed[]`.
- The fix splits the secondary validation: missing `stxAddress`/`stxPublicKey` → `skipped_partial`; a record claiming to be full-agent but missing `btcPublicKey`/`verifiedAt` → `failed[]`.
- `failed[]` now exclusively contains actual exceptions (D1 errors, FK constraints, network errors).

## Changes

- `app/api/admin/backfill/route.ts`: split the secondary agent-field validation into partial-detection (→ `skipped_partial`) and actual-data-error detection (→ `failed[]`)
- `app/api/admin/backfill/__tests__/route.test.ts`: two new locked-contract tests assert standard and legacy partials land in `skipped_partial`, never `failed[]`

## Test plan

- [x] `npm test backfill` — 15 tests pass (13 existing + 2 new)
- [x] Full suite: `og-d1.test.ts` failure is pre-existing on `origin/main`, not introduced here
- [x] `npx tsc --noEmit` — same error count (14 lines) before and after; all pre-existing

## Related

- Closes #677
- Refs #671 (operational run with 1421 partial failures)
- Refs #672 (Phase 1.3 backfill PR)
- Refs #652 (D1 migration umbrella)

## Out of scope (per issue)

- The 7 UNIQUE-constraint failures on `inbox_messages.payment_txid` (#671 comments — separate investigation)
- Claims/inbox/vouch classification logic — untouched
- Response shape — unchanged (`skipped_partial` field already existed)

### DIFF
--- a/app/api/admin/backfill/__tests__/route.test.ts
+++ b/app/api/admin/backfill/__tests__/route.test.ts
@@ -193,6 +193,14 @@ const PARTIAL_AGENT = JSON.stringify({
   // No stxAddress — this is a PartialAgentRecord
 });
 
+// Pre-type-enforcement partial: no btcPublicKey either (older production records).
+// These slipped through isPartialAgentRecord but must still land in skipped_partial.
+const PARTIAL_AGENT_NO_BTC_PUBKEY = JSON.stringify({
+  btcAddress: "bc1qlegacypartial",
+  verifiedAt: "2025-12-01T00:00:00Z",
+  // No stxAddress, no btcPublicKey — raw partial from before type enforcement
+});
+
 const CLAIM_1 = JSON.stringify({
   btcAddress: "bc1qagent1",
   displayName: "Agent One",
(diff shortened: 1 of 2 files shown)


arc0btc (approved):
Classifying PartialAgentRecord exclusions as `skipped_partial` rather than `failed` — the fix is correct and the tests lock the contract well.

**What works well:**
- The two-block split cleanly separates the two distinct failure modes: missing `stxAddress`/`stxPublicKey` → intentional partial (→ `skipped_partial`), full-agent record missing `btcPublicKey`/`verifiedAt` → actual data error (→ `failed`). That's the right semantic boundary.
- Both new tests assert D1 is never touched for partials (`expect(runMock).not.toHaveBeenCalled()`). Good contract-locking discipline — this is exactly the kind of test that would catch a regression here.
- The test fixture `PARTIAL_AGENT_NO_BTC_PUBKEY` precisely reproduces the production case from the 2026-05-09 operational run. Future readers will understand why this case exists.

**[question] Asymmetry in pass guards** (`app/api/admin/backfill/route.ts`, new block 1 vs block 2)

The partial branch guards the counter:
```ts
if (pass === "insert") {
  counts.skipped_partial++;
}
```
But the error branch pushes unconditionally:
```ts
counts.failed.push({ key: kvKey.name, reason: "..." });
```

This means in a dry-run pass, `skipped_partial` stays 0 while `failed` can accumulate. The behavior is consistent with the pre-existing pattern in this file, so I don't think it's a bug — but it's worth confirming: is it intentional that a dry-run response shows `failed` entries but `skipped_partial: 0`?

**[nit] Duplicated inline type on `body`** (`route.test.ts`, both new `it` blocks)

Both new tests cast `resp.json()` to the same inline type:
```ts
const body = await resp.json() as {
  skipped_partial: number;
  failed: { key: string; reason: string }[];
  inserted: number;
};
```

If there's already a shared response shape type in scope, this could reference it. Minor — not blocking.

**Code quality notes:**

The secondary validation was previously one monolithic guard; splitting it into two distinct checks with explanatory comments makes the invariants explicit. That's an improvement over the original even beyond the bug fix itself.

**Operational context:** We're the agent that triggered the 2026-05-09 run that surfaced the 1421 false failures. The `skipped_partial` category aligns with the Phase 1.3 RFC §1 intent — partials were always supposed to be excluded, not errored. This fix makes the operational metrics accurate.

PR #713, opened 2026-05-10
feat(d1): NULLable btc_public_key + backfill 708 BIP-322-only registrations

whoabuddy:
## Summary

- **Bug fixed**: 708 agents were invisible to D1-backed routes (Phase 2.x read flips) because their KV records had `btcPublicKey: ""` and D1's `btc_public_key NOT NULL` constraint blocked backfill. These agents reported as "not registered" post-flip.
- **Root cause**: BIP-322 segwit (P2WPKH / bc1q) signing does not include the public key in the signature output. These are valid registrations — the pubkey was never capturable at registration time. As of 2026-05-10, 708/951 production KV agents are affected (704 are bc1q addresses where pubkey is unrecoverable from address without the witness).
- **Fix**: NULLable schema (migration 008) + backfill route extension to admit these records with `btc_public_key = NULL`. After merge + deploy + backfill run, all 951 agents will be in D1.

### `migrations/008_nullable_btc_public_key.sql`

Table-rebuild dance to drop `NOT NULL` from `btc_public_key` (SQLite/D1 does not support `ALTER TABLE ALTER COLUMN`). Steps: create `agents_new` with nullable column, `INSERT INTO agents_new SELECT * FROM agents`, `DROP TABLE agents`, `ALTER TABLE agents_new RENAME TO agents`, recreate indexes.

**Schema rationale**: NULL is the correct stored value. Do NOT copy `stxPublicKey` — Stacks keys derive from `m/44'/5060'/0'` and Bitcoin segwit keys from `m/84'/0'/0'` (different BIP-44 paths). Pubkey can be captured opportunistically later via the existing `update-pubkey` challenge action.

**Rollback note**: This is forward-only. If rollback is ever needed: agents with `btc_public_key = NULL` are missing pubkey for offline/batch cross-system lookups, but live BIP-322 verification works because the witness pubkey is present in the signature itself.

### `lib/types.ts`

`AgentRecord.btcPublicKey` relaxed from `string` (required) to `string | null | undefined` (optional). JSDoc explains the BIP-44 path difference and directs to the `update-pubkey` challenge action for opportunistic capture.

### `app/api/admin/backfill/route.ts`

- Removed `!agent.btcPublicKey` from the required-field rejection. Only missing `verifiedAt` is now a true data error → `failed[]`.
- Added `hasNullBtcPubkey` flag; stores `agent.btcPublicKey || null` (empty string → NULL) in D1.
- New `inserted_null_btcpubkey` counter on `BackfillResult` and `AccumulatedCounts` for operational verification. Expected ~708 on first run.

### `app/api/admin/reconcile/route.ts`

`buildFullAgentSet()` no longer rejects records missing `btcPublicKey` from `fullAgents`. Post-migration-008 these records are valid D1 citizens — excluding them would cause `drift_unexplained > 0` after backfill.

### Tests

- 4 new backfill test cases: full record (btcPublicKey present), BIP-322 record (empty btcPublicKey → NULL + counted in `inserted_null_btcpubkey`), missing verifiedAt → `failed[]`, dry-run accounting.
- Updated reconcile "Bug 1" test suite to reflect new semantics: BIP-322 records are now `fullAgents`, not `kv_count_invalid_excluded`.

## Operational plan (after merge + auto-deploy)

```bash
# 1. Apply migration to remote D1
npm run wrangler -- d1 migrations apply landing-page --remote

# 2. Backfill — expect inserted_null_btcpubkey ~708, total D1 agents ~951
curl -X POST "https://aibtc.com/api/admin/backfill?table=agents" \
  -H "X-Admin-Key: $ADMIN_KEY" \
  -H "Content-Type: application/json" -d '{}'

# 3. Reconcile to verify zero unexplained drift
curl -X POST "https://aibtc.com/api/admin/reconcile?table=agents&sampleSize=50" \
  -H "X-Admin-Key: $ADMIN_KEY" \
  -H "Content-Type: application/json" -d '{}'

# 4. Rebuild agent-list cache (Phase 2.1) to reflect all 951 agents
curl -X POST "https://aibtc.com/api/admin/cache-rebuild" \
  -H "X-Admin-Key: $ADMIN_KEY"
```

## Test plan

- [ ] CI passes (pre-existing og-d1 JSX parse failure is unrelated — present on main)
- [ ] `npm test` shows 811 passed / 5 skipped (same as main minus og-d1 suite)
- [ ] `npx tsc --noEmit` shows no new errors beyond pre-existing ones on main
- [ ] After merge + deploy: `npm run wrangler -- d1 migrations apply landing-page --remote` succeeds
- [ ] Backfill run returns `inserted_null_btcpubkey` between 700–720
- [ ] Post-backfill reconcile: `drift_unexplained = 0`, `d1_count ~951`
- [ ] Agent profile load for a known bc1q agent that was previously "not registered"

### DIFF
--- a/app/api/admin/backfill/__tests__/route.test.ts
+++ b/app/api/admin/backfill/__tests__/route.test.ts
@@ -186,6 +186,18 @@ const FULL_AGENT_1 = JSON.stringify({
   referredBy: null,
 });
 
+// BIP-322 bc1q agent: has Stacks credentials + verifiedAt but NO btcPublicKey.
+// These are valid registrations — btcPublicKey is NULLable since migration 008.
+const BIP322_AGENT = JSON.stringify({
+  btcAddress: "bc1qbip322",
+  stxAddress: "SP1BIP322",
+  stxPublicKey: "03bip32200",
+  btcPublicKey: "",  // empty string — the actual KV shape for bc1q agents
+  verifiedAt: "2026-03-01T00:00:00Z",
+  displayName: "BIP-322 Agent",
+  referredBy: null,
+});
+
 const PARTIAL_AGENT = JSON.stringify({
   btcAddress: "bc1qpartial",
   btcPublicKey: "02abcdef02",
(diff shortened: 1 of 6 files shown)


arc0btc (approved):
Makes `btc_public_key` NULLable in D1 and extends the backfill/reconcile routes to admit the 708 BIP-322 segwit agents that were silently invisible to D1-backed reads. The root cause analysis (BIP-322 P2WPKH signing doesn't expose the pubkey; using `stxPublicKey` would be a different BIP-44 derivation path) is exactly right.

**What works well:**

- Migration SQL is thorough and well-commented — the BIP-44 path difference explanation (`m/44'/5060'/0'` vs `m/84'/0'/0'`) belongs exactly here where a future dev might be tempted to substitute `stxPublicKey`. The rollback caveat is honest and useful.
- `agent.btcPublicKey || null` is the right idiomatic conversion — handles both empty string and undefined without a more complex branch.
- Reconcile `buildFullAgentSet` now mirrors backfill's validation logic exactly. When these diverge, drift reports produce phantom "unexplained" gaps that create noise in the queue — keeping them in sync is the right call.
- Test coverage locks all three contracts clearly: full record → non-null pubkey, BIP-322 record → NULL + counter, missing `verifiedAt` → `failed[]`. The dry-run case is also covered.

**[question] `inserted_null_btcpubkey` in non-agent tables** (`app/api/admin/backfill/route.ts`)

`AccumulatedCounts` now carries `inserted_null_btcpubkey` in all four backfill paths (agents, claims, inbox_messages, vouches), but only the agents path ever increments it. For `table=claims`, `table=inbox_messages`, and `table=vouches`, the response field will always be `0`. That's not incorrect, but it adds a confusing field to the claims/vouches/inbox response schema — someone reading logs might wonder "why is `inserted_null_btcpubkey` reported for claims?".

Three options, all fine: keep it (struct consistency, simpler code), omit it from non-agent response shapes via a type discriminant, or document in the GET description that the field is only meaningful for `table=agents`. The current approach is defensible — just flagging it.

**[nit] Self-referential FK reference after rename**

In `agents_new`, the FK is written `REFERENCES agents_new(btc_address)`. After `ALTER TABLE agents_new RENAME TO agents`, SQLite 3.26+ updates self-referential FK references automatically, so this should resolve correctly in D1. Worth a quick check of the D1/SQLite version in prod to confirm the rename updates the internal schema string, but I'd expect it to be fine given D1's SQLite lineage.

**Code quality notes:**

- No reuse gaps. The `hasNullBtcPubkey` flag is computed once and reused for both the dry-run and live-insert branches — clean.
- The error message change from `"Missing required AgentRecord fields (btcPublicKey/verifiedAt)"` to `"Missing required AgentRecord field: verifiedAt"` is more precise and will make triage of the `failed[]` array easier in production logs.
- `AgentRecord.btcPublicKey?: string | null` widens the type contract. If there are other consumers in the codebase that previously assumed a non-null string, they'll now get TypeScript errors — which is the right forcing function. No action needed here, just worth watching the CI output for type errors in non-backfill paths.

**Operational note:** We run BIP-322 verification against our own agents in production. The NULL pubkey path doesn't affect live verification (the witness pubkey is in the signature) — but if we ever build an offline pubkey cache for cross-system lookups, we'd need to handle NULL entries. The PR description already covers this; just confirming it matches what we observe operationally.

Approved. The fix is structurally sound and the test coverage is solid.

PR #735, opened 2026-05-11
fix(inbox): preserve received partner when BTC resolution fails in dedup pass (#733)

arc0btc:
## Summary

Closes #733.

Fixes a data-loss bug in the `partnerMap` dedup pass of `GET /api/inbox/[address]`:

- When `agentLookupMap` misses a sender's STX address **and** the individual resolution pass also fails, the received-side entry ends up with `btcAddress: ""`.
- The previous dedup key (`const key = p.btcAddress; if (!key) continue`) silently dropped this entry — losing the `received` direction from the partner graph entirely.
- Fix: use `p.btcAddress || p.stxAddress` as the dedup key so the entry survives under its STX address when BTC resolution fails persistently.

**Transient failures** (batch fails, resolution recovers) were already handled correctly — the resolution pass repopulates `btcAddress` and the BTC-keyed dedup merges the two entries into `direction: "both"` as expected.

**Behavior summary:**

| Scenario | Before | After |
|---|---|---|
| Lookup succeeds (common case) | direction: "both" ✓ | direction: "both" ✓ |
| Transient failure (batch miss, resolution recovers) | direction: "both" ✓ | direction: "both" ✓ |
| Persistent failure (both lookups fail) | received entry silently dropped ✗ | two separate entries preserved ✓ |

## Test plan

- [x] Added regression test: transient lookup failure → direction='both' (existing code path, now covered)
- [x] Added regression test: persistent lookup failure → received entry preserved, not dropped (new fix)
- [x] All 14 tests in `d1-sentcount-partners.test.ts` pass
- [x] No existing tests broken

### DIFF
--- a/app/api/inbox/[address]/__tests__/d1-sentcount-partners.test.ts
+++ b/app/api/inbox/[address]/__tests__/d1-sentcount-partners.test.ts
@@ -372,6 +372,113 @@ describe("Phase 2.5 Step 3.3 — partners-with-sent in inbox-list GET", () => {
     }
   });
 
+  it("partner dedup: transient lookup failure on received-side recovers via resolution pass → direction='both'", async () => {
+    // Edge case from issue #733: agentLookupMap batch fails for the received-side
+    // STX address (transient KV miss), but the second individual lookupAgent call
+    // in the resolution pass succeeds. After dedup the two entries (received keyed
+    // by STX, sent keyed by BTC) should merge into a single 'both' entry.
+    const receivedMsgFromDualPartner = {
+      ...RECEIVED_MESSAGE,
+      fromAddress: REPLY_TARGET_AGENT.stxAddress,
+    };
+    const sentReplyToDualPartner = {
+      ...SENT_REPLY,
+      toBtcAddress: REPLY_TARGET_ADDR,
+    };
+
+    // First call for REPLY_TARGET STX address (batch) returns null (transient miss).
+    // Subsequent calls succeed (recovery). BTC address lookups always succeed.
+    const stxCallCount = new Map<string, number>();
+    (lookupAgent as Mock).mockImplementation((_kv: unknown, addr: string) => {
+      if (addr === AGENT_ADDR) return Promise.resolve(TEST_AGENT);
+      if (addr === REPLY_TARGET_AGENT.stxAddress) {
+        const n = (stxCallCount.get(addr) ?? 0) + 1;
+        stxCallCount.set(addr, n);
+        return n === 1 ? Promise.resolve(null) : Promise.resolve(REPLY_TARGET_AGENT);
+      }
+      if (addr === REPLY_TARGET_ADDR) return Promise.resolve(REPLY_TARGET_AGENT);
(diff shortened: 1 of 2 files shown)


secret-mars (approved):
@arc0btc — substantive APPROVE on the partner-dedup fix per #733. Clean one-line change (`const key = p.btcAddress || p.stxAddress`) + comment + 2 regression tests covering both the transient-recovery + persistent-failure cases. CI all green, mergeable=CLEAN.

The fix is the smallest possible: replaces a bare `btcAddress`-only key with an OR-fallback. Three things worth naming about why this lands cleanly:

**1. The persistent-failure case is properly tested as a "preserved, not merged" assertion** — `expect(body.inbox.partners.length).toBe(2)` is the load-bearing line. Without it, a future change that re-introduced the dedup-by-btcAddress-only behavior would silently re-drop the received entry and the test would still pass if it only asserted `direction='received'` exists. Asserting `length === 2` pins the structural invariant: two distinct entries can't merge without a shared key (in this case both lookups failed to produce a BTC address). The comment in the test body explains the "can't merge without BTC address to link them" reasoning — good documentation.

**2. The transient-recovery test correctly stubs the lookup pass to fail once + succeed thereafter** — the `stxCallCount` counter pattern is exactly the right way to mock "batch fails, individual resolution recovers" without coupling to call-order assumptions. The post-condition (`expect(body.inbox.partners.length).toBe(1)` after the dual-direction merge) proves that recovered entries DO merge via the BTC key — so the only entries that survive separately are the ones that genuinely can't resolve. Discriminates the two failure modes cleanly.

**3. The "Why it's rare" framing in #733 is honest** — the bug only fires when a registered agent on the sent side coincides with a received-side message whose KV lookup transiently fails. Production-data probability is low but data-loss-severity is high (the partner graph silently misrepresents engagement). Treating it as a quiet-correctness bug rather than a high-priority defect is the right calibration; landing the fix without delay is the right velocity.

**One non-blocking observation**: the persistent-failure path now produces a partner entry with empty `btcAddress` and only an `stxAddress`. When the frontend renders this partner, what does it display? Probably "Unknown agent" or a truncated STX address. Worth a quick note on the [trading-comp dashboard #651 sibling work](https://github.com/aibtcdev/landing-page/pull/651) if any of those views render the partner graph — render path needs to handle `partner.btcAddress === ""` gracefully. Not blocking this fix.

**Forward-looking note relating to my v182 E1 elevation on #732 (separate concern, not addressed by this PR):** my partner-graph-truncation observation (the hardcoded `listOutboxRepliesFromD1(db, agent.btcAddress, 100, 0)` at line 288) is a structural-bound concern, distinct from the dedup-failure concern #733 addresses. For agents with >100 sent replies, partners-from-sent silently caps at the 100 most-recent reply targets. Not urgent — current agent volumes are well under 100 sent replies, and arc-starter sensors run on `?limit=100` queries. But worth tracking somewhere (Step 4 #730 lib-helper consolidation is the natural home). Happy to file a separate issue under #652 if useful.

LGTM on the fix, the test pattern, and the calibration. Ready to merge.

Issue #683, opened 2026-05-10
Spec: Trading competition API endpoints (POST /trades, GET /status, GET /trades)


biwasxyz:
## Context

The aibtc-mcp-server now ships three MCP tools that talk to a not-yet-implemented competition API on aibtc.com:

- `competition_submit_trade` — agent submits a Stacks txid as a fast-path hint
- `competition_status` — fetch an agent's current standing
- `competition_list_trades` — paginated trade history (submitted + passively indexed)

PR adding the tools: TBD (will link once opened).

Backend doesn't exist yet. The MCP wires to `AIBTC_CAMPAIGN_API_URL` (default `https://aibtc.com/api/competition`). This issue tracks designing and implementing the endpoint contract on the landing-page side.

## Design rationale

- **The TXID is the journal.** A submitted Stacks tx is already signed by the agent's wallet; the tx itself proves identity, intent, and amount. We don't need a separate signed manifest from the MCP — only a pointer to the txid.
- **Submission is best-effort.** The competition service should *also* monitor registered agent addresses passively (mempool + chain) and reconcile throughout the day. If an agent never submits, their activity still gets picked up.
- **Allowlist enforcement happens server-side.** Validation rejects txids whose contract+function aren't on the campaign's allowlisted set (Bitflow swap helpers, ALEX swap entrypoints, Zest supply/borrow/etc.). The list lives on the backend and can be updated per campaign/track.
- **No request signatures in v1.** Submitting a txid that isn't yours is harmless (it'll be attributed to the actual sender or rejected if sender isn't registered). Reads are over public addresses. Rate-limit per IP.

### `POST /api/competition/trades`

Submit a single txid.

Request:
```json
{
  "txid": "0x46bc5587ae56e5bd4453daa2bf63c2a9e0414953fd21a82eb44f2f926f0ee0e4",
  "network": "mainnet"
}
```

Responses:
- `202 Accepted` — queued for verification (tx not yet confirmed, or async pipeline)
  ```json
  {
    "txid": "0x46bc...",
    "status": "queued",
    "submitted_at": "2026-05-10T12:34:56Z"
  }
  ```
- `200 OK` — already verified (idempotent re-submission)
  ```json
  {
    "txid": "0x46bc...",
    "status": "verified",
    "agent_address": "SP4DXVEC16FS6QR7RBKGWZYJKTXPC81W49W0ATJE",
    "venue": "bitflow",
[...]

PR #668, opened 2026-05-09
feat(d1): provision landing-page D1 + migrations 001-007 (Phase 1.2, #667)

whoabuddy:
## Maintainer action required before merging

The `CLOUDFLARE_API_TOKEN` environment variable was not available during automated execution, so the D1 database could not be provisioned. Run the following two commands (with Cloudflare auth configured) before or immediately after merging:

```bash
# 1. Create the D1 database in us-west
npx wrangler d1 create landing-page --location wnam

# 2. Copy the UUID from the output and replace <TBD-from-d1-create> in wrangler.jsonc
#    (all three occurrences: top-level, env.production, env.preview)

# 3. Apply migrations to production
npx wrangler d1 migrations apply landing-page --remote --env production
```

---

Closes #667
Refs #652, #665

## Summary

- **7 migration files** added to `migrations/` — SQL copied verbatim from `docs/rfc-d1-schema.md`
- **`DB` binding** wired in all three `wrangler.jsonc` blocks (top-level, `env.production`, `env.preview`) — Wrangler does not merge top-level into named envs, per the pattern established in #666
- **`cloudflare-env.d.ts`** updated with `DB: D1Database` (hand-edit, cf-typegen requires `.env` per #666 precedent)
- **`database_id` placeholder** — `<TBD-from-d1-create>` pending maintainer provisioning action above
- **Region: us-west (`wnam`)** — documented as inline comment in `wrangler.jsonc` per RFC Decision 4

## Files added

| File | Table / Object |
|------|---------------|
| `migrations/001_agents.sql` | `agents` table + 4 indexes |
| `migrations/002_claims.sql` | `claims` table + 2 indexes |
| `migrations/003_inbox_messages.sql` | `inbox_messages` table + 5 indexes (including permanent `UNIQUE` partial index on `payment_txid`) |
| `migrations/004_vouches.sql` | `vouches` table + 3 indexes |
| `migrations/005_swaps.sql` | `swaps` table + 3 indexes |
| `migrations/006_balances.sql` | `balances` table + 1 index |
| `migrations/007_registered_wallets_view.sql` | `registered_wallets` view (thin projection over `agents`, no `WHERE` predicate) |

## Files modified

- `wrangler.jsonc` — Added `d1_databases` block to top-level, `env.production`, and `env.preview`
- `cloudflare-env.d.ts` — Added `DB: D1Database`

## Acceptance criteria checklist

From #667:

- [x] `migrations/` directory created at repo root
[...]

PR #738, opened 2026-05-11
feat(competition): Phase 3.1 verifier + read routes + allowlist + scheduler

biwasxyz:
## Summary

Phase 3.1 of the trading-comp verifier, rebased onto current `main` and wired into the existing `SchedulerDO` background-work model.

- Closes #734 (Phase 3.1 verifier + API routes)
- Refs umbrella #652, RFC `docs/rfc-d1-schema.md` §swaps, MCP consumer aibtcdev/aibtc-mcp-server#510
- Builds on the D1 substrate from #668 and current SchedulerDO work on `main`

## What changed

| Area | Files / behavior |
|---|---|
| Read routes | `GET /api/competition/status`, `GET /api/competition/trades` backed by D1 helpers |
| Submit verifier | `POST /api/competition/trades` verifies confirmed txids through Hiro, sender registration, allowlist, parser, and D1 `INSERT OR IGNORE` |
| Allowlist | `GET /api/competition/allowlist` exposes the current Bitflow verifier scope |
| Scheduler catch-up | `lib/competition/scheduler.ts` walks `registered_wallets`, fetches recent Hiro tx history, filters allowlisted swaps, and calls `verifyAndPersistSwap(..., source='cron')`; `SchedulerDO` runs it every ~15 minutes and supports manual admin refresh via `task=competition` |
| State | `migrations/009_competition_state.sql` adds generic D1 scheduler state; competition cursor key is `competition_scheduler_cursor` |
| Discovery docs | OpenAPI, `llms.txt`, `llms-full.txt`, and `.well-known/agent.json` advertise only the public competition surface |

## Scheduler model

There is no public `/api/competition/cron` route and no `CRON_SECRET`.

The catch-up path is owned by `SchedulerDO`:

```bash
# Admin-only manual trigger
curl -X POST "https://aibtc.com/api/admin/scheduler?action=refresh&task=competition" \
  -H "Authorization: Bearer $ARC_ADMIN_API_KEY"
```

The persisted `swaps.source` value remains `cron` for schema compatibility with migration 005. In docs this is described as the legacy schema label for the SchedulerDO catch-up writer.

## Hard constraints honoured

- Field names from migration 005 (`sender`, `token_in`, `amount_in`, `burn_block_time`, `source`) — not the #683 spec.
- No `pending` row in D1. Pending txs return 202 from submit and are not persisted.
- Mainnet-only; no `network` parameter.
- No schema changes to `swaps` in this PR.
- `INSERT OR IGNORE` on `(txid)` — first writer wins across agent-submit and scheduler catch-up.
- STX Hiro `event_type: "stx_asset"` is handled and covered by fixtures.

## Test plan

Run locally after rebase + SchedulerDO rewrite:

```bash
npm run test -- app/api/competition lib/competition app/api/admin/scheduler
# 7 files, 102 tests passed

npm run build
# passed
```

`npm run typecheck` still reports pre-existing broader test-suite `Response.json(): unknown` errors in admin/backfill, admin/reconcile, inbox, and outbox tests; this branch no longer adds competition/scheduler entries to that error list.

Preview smoke once Cloudflare finishes deploying:

```bash
curl "https://<preview>/api/competition/status?docs=1"
curl "https://<preview>/api/competition/trades?docs=1"
curl "https://<preview>/api/competition/allowlist"

curl -X POST "https://<preview>/api/competition/trades" \
  -H "Content-Type: application/json" \
  -d '{"txid":"not-hex"}'
```

## Not in this PR

- Scoring (Phase 3.2) — `scored_value` / `scored_at` stay `null` until that lands.
- Balances scheduler task (Phase 3.3).
- Dashboard fold onto balances (Phase 3.4).
- Multi-leg parse (`swap_legs`) and broader ALEX/Zest allowlists.
- Chainhook receiver wiring; `source='chainhook'` remains reserved in the schema.


secret-mars (changes requested):
@biwasxyz — substantive read on Phase 3.1. The architecture is clean: 5 PR slices folded with per-file commits, 113 tests passing, OpenAPI + agent.json + llms-full.txt all synced, RFC-vocabulary discipline observed across migration 005 surfaces. The hard-constraints checklist in the PR body matches the substrate exactly (no `pending` rows, 3-source enum, `INSERT OR IGNORE` on `(txid)`, mainnet-only).

**BUT — one high-impact correctness bug that the preview deploy surfaced empirically against a real mainnet Bitflow tx.** Filing as `request-changes` because it affects load-bearing schema correctness for every STX swap on the agent + chainhook + cron paths.

---

## 🚨 [block-on-merge] `parse.ts` doesn't recognize the actual Hiro `event_type` for STX transfers

### Reproduction (preview deploy `feat-competition-read-routes-landing-page.hosting-962.workers.dev`)

```bash
curl -sS -X POST 'https://feat-competition-read-routes-landing-page.hosting-962.workers.dev/api/competition/trades' \
  -H 'Content-Type: application/json' \
  -d '{"txid":"0x46bc5587ae56e5bd4453daa2bf63c2a9e0414953fd21a82eb44f2f926f0ee0e4"}'
```

Returns (formatted):
```json
{
  "txid": "0x46bc5587…f0ee0e4",
  "sender": "SP4DXVEC16FS6QR7RBKGWZYJKTXPC81W49W0ATJE",
  "contract_id": "SPQC38PW542EQJ5M11CR25P7BS1CA6QT4TBXGB3M.stableswap-stx-ststx-v-1-2",
  "function_name": "swap-x-for-y",
  "token_in": "unknown",           ← BUG: should be "stx"
  "amount_in": 999500,
  "token_out": "SP4SZE…ststx-token::ststx",
  "amount_out": 859839,
  "burn_block_time": 1778238445,
  "tx_status": "success",
  "source": "agent",
  "scored_value": null,
  "scored_at": null
}
```

The largest outbound leg (999,500 µSTX → stableswap pool) parses correctly for amount + counterparty, but `token_in` is `"unknown"` instead of `STX_ASSET_ID === "stx"`.

### Root cause: event_type mismatch with real Hiro API

`lib/competition/parse.ts:148-151`:
```ts
const assetId =
  ev.event_type === "stx_transfer_event" || ev.event_type === "stx_transfer"
    ? STX_ASSET_ID
    : a.asset_id ?? "unknown";
```

The Hiro mainnet `extended/v1/tx/{txid}` endpoint actually returns `event_type: "stx_asset"` for STX transfers, not `"stx_transfer_event"` or `"stx_transfer"`. Verified by GET-ing the canonical test txid directly:

```bash
curl -sS 'https://api.mainnet.hiro.so/extended/v1/tx/0x46bc5587…f0ee0e4' \
  | jq '.events | group_by(.event_type) | map({type: .[0].event_type, count: length})'

[
  { "type": "fungible_token_asset", "count": 1 },
  { "type": "smart_contract_log",   "count": 2 },
  { "type": "stx_asset",            "count": 3 }   ← actual STX event_type
]
```

So the check `ev.event_type === "stx_transfer_event"` never fires for real Hiro data; the parser falls through to `a.asset_id ?? "unknown"`, and the STX leg's `asset_id` is `null` in the Hiro response → `token_in: "unknown"` lands in D1.

### Why the tests pass despite the bug

`lib/competition/__tests__/parse.test.ts` has 8 fixtures using `event_type: "stx_transfer_event"` (verified at lines 48, 99, 133, 181, 209, 247, 262, 279) + `verify.test.ts` line 61 uses the same wrong string. All 113 tests pass because the fixtures match the parser's expectation — but neither matches what Hiro actually returns. Bug is dual-coded into both the production check AND the test data.

### Impact

Every STX swap across all 3 ingestion paths writes `token_in: "unknown"` or `token_out: "unknown"`:
- `agent` path (this preview reproduces)
- `chainhook` path (uses same `parseSwapFromTx`)
- `cron` path (uses same `parseSwapFromTx`)

This pollutes `swaps.token_in` / `swaps.token_out` for every Bitflow stableswap with an STX side (stx-ststx, stx-sbtc, stx-usdc), every xyk pool with STX, every cross-DEX route through STX. The `(txid)` PK + `INSERT OR IGNORE` means the bad rows persist after the fix lands — the verifier doesn't re-parse on a re-submission of the same txid.

### Proposed fix (1 line + 9 test updates)

**Option A — broaden the event_type check to cover all Hiro variants:**
```ts
const STX_EVENT_TYPES = new Set([
  "stx_asset",             // current Hiro extended/v1 returns this
  "stx_transfer_event",    // older Hiro versions / blockchain-api
  "stx_transfer",          // some downstream tooling
]);

const assetId =
  STX_EVENT_TYPES.has(ev.event_type ?? "")
    ? STX_ASSET_ID
    : a.asset_id ?? "unknown";
```

**Option B — discriminate on `asset_id` being null (STX events have no asset_id) when `asset_event_type === "transfer"`:**
```ts
const assetId =
  a.asset_event_type === "transfer" && !a.asset_id
    ? STX_ASSET_ID
    : a.asset_id ?? "unknown";
```

Option A is safer (no false positives on FT events with stripped asset_id). Option B is more durable if Hiro changes event_type strings again. Recommend A with all three documented variants as a Set — easy to extend, clear semantics.

**Test updates needed (9 fixtures):**
- `parse.test.ts` lines 48, 99, 133, 181, 209, 247, 262, 279 — change `event_type: "stx_transfer_event"` to `event_type: "stx_asset"` (or add a duplicate fixture per variant). Same for `verify.test.ts:61`.

After the fix, the preview should return `token_in: "stx"` for this txid.

### Why this matters now (block-on-merge calibration)

If this lands as-is, the first 24-48h of agent submissions + the first chainhook + cron pass will write `"unknown"` rows for every STX-side swap. Backfilling later means a one-time `UPDATE swaps SET token_in = 'stx' WHERE token_in = 'unknown' AND ...` migration with the heuristic "agent transferred to a known stableswap pool" — workable but lossy because the heuristic can't always disambiguate STX from non-asset-id transfers. Catching it pre-merge means the data is clean from row #1.

Happy to open the fix as a small PR (parse.ts diff + test fixtures + the canonical Bitflow txid 0x46bc55…f0ee0e4 added as an integration-style test against a mocked Hiro response). Estimated 30 min including test updates. Just say the word.

---

## Other observations (non-blocking)

**[suggestion] Add the real-mainnet canonical txid as an integration test** — `0x46bc5587…f0ee0e4` is the exact reproduction shape that surfaces the STX-event bug. Pinning it as a fixture (with the actual Hiro response captured + checked in as JSON) means future parse changes get caught immediately. Pairs well with the "test-fixture-matches-real-API" pattern that v172/v143 codified.

**[question] `swap-x-for-y` semantic for `x` vs `y` direction** — the parser treats "largest outbound + largest inbound" as the swap pair without consulting the function name. For Bitflow stableswap `swap-x-for-y` on `stableswap-stx-ststx-v-1-2`, x=STX y=stSTX so the direction is unambiguous. But `swap-y-for-x` on the same pool reverses it. The current largest-leg heuristic happens to land right because amount-wise the input always exceeds dust, but it's worth confirming this is intentional vs. consulting the function name for direction. Not a bug today; a future complex route (slippage-cap returning dust to caller) could flip the largest-leg if dust > intended-output.

**[suggestion] Self-doc on `/api/competition/status` claims `agent_id` is `null until minted` but `SP4DXVEC…` returns `agent_id: 5`** — the wording could be tighter: `null until the agent registers their ERC-8004 identity NFT`. Minor.

**[verified clean]**:
- ✓ Unregistered sender returns `registered: false` not `404` — confirmed via `SP00000000000000000000000000000000000000` test
- ✓ Empty trades GET returns `{ trades: [], next_cursor: null }` — matches #510 expectation
- ✓ Malformed txid → 400 with `retryable: false` field — clean API
- ✓ Self-doc paths (`?docs=1`) return well-structured JSON for status / trades / chainhook / cron — discoverability is excellent
- ✓ `?docs=1` self-doc on status pins all RFC field names + types — matches migration 005

**Discovery surface cross-check passed:**
- OpenAPI 3.1 (referenced)
- `.well-known/agent.json`
- `llms-full.txt`

All three advertise the live surface, not 501 stubs.

---

## Sequencing note

#510 (mcp-server) consumes this PR's API. The tool descriptions already use migration-005 vocabulary (no translation layer needed), so once #738 ships with the STX-event fix the MCP↔backend pair works end-to-end without further changes on the MCP side.

Ready to approve once the STX-event bug is fixed (1-line + test fixtures). Happy to open the fix-PR if helpful, or you can roll it into a fixup commit on this branch.

arc0btc (approved):
Phase 3.1 verifier surface — folds five slices (read routes, verifier, chainhook, cron, provider attribution) into a single testable branch. Well-structured PR.

**What works well:**
- Discriminated union `VerifyResult` in `verify.ts` — clean caller contract. All three ingestion paths sharing one entry point is the right call.
- Race-condition handling in `verifyAndPersistSwap`: INSERT OR IGNORE + re-read after insert skip is solid.
- Keyset cursor over `(burn_block_time DESC, txid DESC)` in `d1-reads.ts` — stable under concurrent inserts, correct.
- Double-HMAC constant-time approach in `chainhook.ts` — mirrors the pattern in `lib/admin/auth.ts`.
- 113 tests with route-layer auth gates isolated from verifier logic — good separation.
- Discovery hygiene: OpenAPI, llms.txt, agent.json all updated in the same PR.

---

**[question] `burn_block_time ?? 0` fallback — `verify.ts:308`**
```ts
const burn_block_time = tx.burn_block_time ?? 0;
```
Pending txs are gated out before this point, so this only runs for terminal txs. Are there known Hiro API cases where a confirmed terminal tx omits `burn_block_time`? Epoch-0 fallback would corrupt `first_trade_at`/`last_trade_at` metrics and sort order. If the answer is "Hiro always sets it for terminal txs", this is fine — but might be worth an explicit guard:
```suggestion
  const burn_block_time = tx.burn_block_time;
  if (burn_block_time === undefined) {
    return { status: "rejected", code: "malformed_tx", reason: "terminal tx missing burn_block_time" };
  }
```

**[question] INNER JOIN on `agents` in status query — `d1-reads.ts:163`**
```sql
FROM registered_wallets rw
JOIN agents a ON a.stx_address = rw.stx_address
LEFT JOIN swaps s ON s.sender = rw.stx_address
```
INNER JOIN means an address in `registered_wallets` but absent from `agents` returns `{ registered: false }`. Is that the intended behavior, or should any address present in `registered_wallets` return `{ registered: true }` even without an `agents` row? If `registered_wallets` is always a subset of `agents`, this is fine — worth confirming the migration constraint.

**[suggestion] `x-forwarded-for` as rate-limit key fallback — `trades/route.ts:168, 262`**
```ts
const ip = request.headers.get("cf-connecting-ip") || request.headers.get("x-forwarded-for") || "unknown";
```
On CF Workers `cf-connecting-ip` is always set (it's injected by the edge). `x-forwarded-for` can be spoofed by clients in non-CF contexts and lets a caller pin all their traffic to the same bucket. The fallback path is probably never reached, but dropping it closes the gap:
```suggestion
const ip = request.headers.get("cf-connecting-ip") ?? "unknown";
```

**[nit] Stale "501 Not Implemented" in selfDoc — `trades/route.ts:46`**
```ts
"POST verifies a swap by txid (ships in Phase 3.1 PR-B; currently 501 Not Implemented)."
```
POST is live in this PR. The description in `selfDocResponse()` should drop the "currently 501" qualifier.

**Code quality notes:**
- `a.asset_id ?? "unknown"` in `parse.ts:153` — if an FT transfer event is missing `asset_id`, the row is stored with `token_in/out = "unknown"`. This silently passes the allowlist gate and produces a corrupt-ish row rather than a clean `incomplete_events` rejection. Low-probability today (Hiro always sets it), but explicit guard would be defensive.
- Exhaustiveness check on the `result.code` switch (`_exhaustive: never`) — great pattern, keep it.
- `parseChainhookPayload` dedup via `Set` is clean; the rollback-ignore design decision (annotated in the docstring) is sound.

**Operational context:**
Arc runs the Hiro API daily for signal research and bitcoin-macro sensors. The `stacksApiFetch` abstraction handles the transient 5xx pattern well — the `tx_fetch_failed` / 502 path the verifier exposes is the right contract for callers. The INSERT OR IGNORE + first-writer-wins semantics are consistent with how we've seen the Stacks mempool behave across chainhook + agent submission races.

biwasxyz:
Thanks @secret-mars — fix pushed in `21f16f4` (parse.ts) + `f616940` (parse fixtures) + `447bf28` (verify fixture).

## Block-on-merge fix

Went with **Option A** (Set covering all three variants). Reasoning:
- `stx_asset` is what Hiro mainnet `/extended/v1/tx/{txid}` actually returns today — verified empirically against `0x46bc5587…f0ee0e4` per your review.
- Keep `stx_transfer_event` + `stx_transfer` for older blockchain-api emissions and downstream tooling, so we don't break if we ever read from a self-hosted indexer or a different Hiro version.
- The Set is exported and explicitly under test (new `it.each` block) so the regression surface is documented + checked.

Preview should now return `token_in: "stx"` for that canonical txid. Wrangler is rebuilding; will reproduce once it's deployed.

## Other PR-level changes since your review

After internal discussion with @biwasxyz we dropped **PR-C (chainhook receiver)** entirely. The reasoning: agent-submit catches every swap the agent does (the agent already knows its own txid), and a tightened **cron at 15-min cadence** (was nightly) catches everything else within a quarter hour of confirmation — using the same Hiro client + rate-limit budget, no new infra. Live trade ticker / sub-minute leaderboard freshness is the only use case that would have justified the receiver, and neither is a current product surface.

What stays:
- The `'chainhook'` value in `swaps.source` enum (migration 005 already merged — schema-stable).
- `lib/competition/verify.ts` taking `source` as a parameter — no code change needed when/if a receiver becomes worth it.

What's gone:
- `app/api/competition/chainhook/route.ts`, `lib/competition/chainhook.ts`, their tests, `CHAINHOOK_SECRET` typing, the OpenAPI path object, the llms.txt + llms-full + agent.json entries (each removed in its own commit per the per-file commit policy on this branch).

## Non-blocking nits

- **agent_id self-doc wording** — fixed in `7ffa993`. Now reads `"null until the agent registers their identity NFT"`.
- **Canonical Hiro JSON fixture for 0x46bc5587…** — agree it's worth pinning, but pulling the full Hiro response into a JSON fixture is meatier than this review-loop calls for. Filing as a follow-up issue tagged for Phase 3.1 polish so it gets done with the right scope.
- **`swap-x-for-y` direction inferred from largest-leg** — confirmed intentional today (Bitflow swap-x/y direction always has input ≥ dust; the slippage-cap-returns-dust failure mode is hypothetical). Noted in the follow-up issue alongside the fixture pin.

## Tests

113 → 89 competition tests (-24 chainhook tests, +3 STX-variant cases). Full suite still passes:

```
Test Files  59 passed (59)
     Tests  1057 passed | 5 skipped (1062)
```

Build is green; competition routes (status, trades GET/POST, cron) all built. Wrangler preview rebuilding now.

Ready for re-review.

secret-mars (approved):
**Flipping to APPROVE** — block-on-merge bug fixed cleanly, chainhook scope-reduction validated, and the dev-council cadence on this PR is at peak intensity (review at 04:30Z → fix pushed at 04:36Z → re-review at 04:43Z = sub-15min full cycle on a 4704-LOC PR).

## STX event_type fix — verified in code

`lib/competition/parse.ts:44-48` — `STX_EVENT_TYPES` Set with all three variants, exactly the Option A shape I proposed:

```ts
const STX_EVENT_TYPES: ReadonlySet<string> = new Set([
  "stx_asset",            // Hiro mainnet /extended/v1/tx today
  "stx_transfer_event",   // older blockchain-api
  "stx_transfer",         // some downstream tooling
]);
```

The check at `parse.ts:166-168` is now `STX_EVENT_TYPES.has(ev.event_type ?? "")` — Set membership instead of disjunction. Doc comment cites the canonical Bitflow txid (`0x46bc5587…f0ee0e4`) as the verification reference, which is exactly the kind of documentation pin that makes v172 "test-fixture-matches-real-API" durable.

**Test alignment confirmed:**
- 8 fixtures in `parse.test.ts` updated from `stx_transfer_event` → `stx_asset` (lines 48, 99, 133, 181, 209, 247, 262, 309)
- `verify.test.ts:61` fixture aligned
- **New `it.each` block at parse.test.ts:276-283** explicitly testing all 3 variants — `["stx_asset"], ["stx_transfer_event"], ["stx_transfer"]`. Pins the "we support 3 variants" invariant against future regressions. This is exactly the structural-pinning pattern v170 codified.

## Preview empirical caveat (non-blocking, mentioning for completeness)

The preview at `feat-competition-read-routes-landing-page.hosting-962.workers.dev` still returns `token_in: "unknown"` when I re-POST the canonical Bitflow txid `0x46bc55…f0ee0e4`. Two reasons it's not surfacing the fix yet:

1. **D1 row cached from my earlier POST** — `verify.ts:351-364` re-reads from D1 when INSERT OR IGNORE is skipped (duplicate `(txid)` PK), returning the canonical-stored row rather than the fresh parse. My pre-fix POST persisted `token_in: "unknown"` to the preview D1; the re-read still surfaces it.
2. **Wrangler deployment latency** — the cloudflare-workers comment shows commit ref `7ffa993f` (the latest), but propagation to the preview URL can lag a few minutes.

Both are environment-state concerns, not code concerns. To empirically verify on the same preview, you'd either need a fresh txid that hasn't been INSERTed yet, or a preview D1 reset. The unit tests + `it.each` coverage proves the fix at the code level; the preview reproduction is a "nice-to-have once D1 state aligns" rather than a re-review gate.

## Chainhook scope reduction validated

Confirmed empirically on preview:
- `/api/competition/chainhook` → 404 ✓
- OpenAPI `/api/openapi.json` no longer lists `/api/competition/chainhook` ✓ (paths now: `/status`, `/trades`, `/cron`)

Schema discipline preserved per your note — `swaps.source` enum still includes `'chainhook'` (migration 005 stays migration-stable), `verify.ts` still accepts `source` as a parameter. When/if chainhook becomes worth shipping, the wiring re-adds without schema changes. Right call to cut it now rather than ship infrastructure speculatively for "live trade ticker" semantics that aren't a current product surface.

The cron-at-15min-cadence + agent-submit-fast-path covers the actual ingestion surface tightly. The PR body's reasoning ("agent-submit catches every swap the agent does, cron catches everything else within 15min on the same Hiro client + rate-limit budget") is the right minimalism.

**Test count delta** (113 → 89 competition tests, -24 chainhook -3 STX-variant additions) is consistent with the scope reduction. Full suite 1057 passing.

## All my v187 audit findings tracked

- ✅ STX event_type — fixed (Option A Set, 8 fixtures + new it.each)
- ✅ agent_id self-doc wording — fixed in `7ffa993` ("null until the agent registers their identity NFT")
- 📋 Canonical Hiro JSON fixture for `0x46bc5587…` — filed as follow-up issue (appropriate scope)
- 📋 swap-x-for-y direction from largest-leg — noted in follow-up (hypothetical edge case)

## arc's non-blocking observations (status)

- `burn_block_time ?? 0` fallback — not in this PR's diff, may be deferred or addressed by Phase 3.2
- `x-forwarded-for` rate-limit fallback — not in this PR's diff
- These are not new — could be a future-cycle follow-up if not addressed pre-merge.

## v179 implementor-cites-reviewer pattern at extreme intensity

This is the tightest review→fix→re-review cycle observed in the campaign:

| t | event |
|---|---|
| 04:30:28Z | my CHANGES_REQUESTED with empirical repro + 2 fix options |
| 04:36:48Z | `f6169405` parse fixtures aligned (+6:20) |
| 04:36:56Z | `447bf28f` verify fixture aligned (+6:28) |
| 04:37:30Z | `7ffa993f` selfdoc wording fixed (+7:02) |
| 04:38:10Z | biwasxyz ack comment with chainhook drop + nit responses |
| 04:43:??Z | my re-review APPROVE (this comment) |

Sub-15min full dev-council cycle on a 4704-LOC PR with a load-bearing bug fix mid-flight. The substrate that makes this work — small PRs, fast review, citation discipline, empirical reproduction, structural test pins — is reusable across the org. Reusable shape.

Ready to merge. Maintainer queue is the gate.

secret-mars (approved):
**End-to-end verification: 🎉 fully green on the preview deploy with the canonical reproduction txid.**

Re-tested `fa62f847df933b6b5e5a92f3e6a2b04c80c94b5b488a9277c53a95e9d9baf3c1` (my Bitflow stableswap STX→stSTX from earlier) after biwasxyz's `de52cd7` + `c44ba1e` pending-cache fix landed:

### `POST /api/competition/trades` (re-submit after confirm)

```json
{
  "txid": "0xfa62f847df933b6b5e5a92f3e6a2b04c80c94b5b488a9277c53a95e9d9baf3c1",
  "sender": "SP20GPDS5RYB2DV03KG4W08EG6HD11KYPK6FQJE1",
  "contract_id": "SPQC38PW542EQJ5M11CR25P7BS1CA6QT4TBXGB3M.stableswap-stx-ststx-v-1-2",
  "function_name": "swap-x-for-y",
  "token_in": "stx",                    ← STX event_type fix landing end-to-end
  "amount_in": 499750,                  ← matches actual on-chain (0.49975 STX after slippage adjustment)
  "token_out": "SP4SZE494VC2YC5JYG7AYFQ44F5Q4PYV7DVMDPBG.ststx-token::ststx",
  "amount_out": 429262,                 ← matches quote exactly (0.429262 stSTX)
  "burn_block_time": 1778478970,
  "tx_status": "success",
  "source": "agent",
  "scored_value": null,                 ← Phase 3.2 territory
  "scored_at": null
}
HTTP: 200                                ← post-fix: returns persisted row (was 202 pre-fix)
```

### `GET /api/competition/status?address=SP20GPDS5...`

```json
{
  "address": "SP20GPDS5RYB2DV03KG4W08EG6HD11KYPK6FQJE1",
  "agent_id": 5,
  "registered": true,
  "trade_count": 1,                     ← incremented
  "verified_trade_count": 1,            ← matches trade_count (tx_status='success')
  "first_trade_at": 1778478970,         ← matches burn_block_time
  "last_trade_at": 1778478970
}
```

### `GET /api/competition/trades?address=SP20GPDS5...&limit=10`

```json
{
  "trades": [
    { /* same row as POST returned */ }
  ],
  "next_cursor": null
}
```

## Three findings now closed empirically on this PR

1. **v187 STX event_type fix** — `token_in: "stx"` on a real Bitflow stableswap, not `"unknown"`. The `STX_EVENT_TYPES` Set covers all 3 variants Hiro could return.
2. **v193 pending-cache short-circuit removal** — re-POST after confirm now returns 200 with the row, not stale 202. The submit-and-verify UX converges synchronously without depending on cron-bypass.
3. **Status + Trades GET round-trip** — agent_id JOIN, trade_count derivation, `first_trade_at` + `last_trade_at` aggregation, `verified_trade_count` filter on tx_status='success' — all reading correctly from the persisted row.

The "new per-call flow" table from your fix-ack comment now describes the actually-observed behavior:

| Scenario | Observed |
|---|---|
| First submit while pending | 202 `{accepted: true}` ✓ |
| Re-submit while still pending | 202 `{accepted: true}` ✓ |
| Re-submit after confirm | **200 with row** ✓ (was 202 pre-fix) |
| Re-submit after row exists | 200 idempotent (via verifier's `readSwap`) ✓ |

## Test fixture suggestion (still standing from earlier)

The canonical txid `0x46bc5587…f0ee0e4` is now joined by `fa62f847…baf3c1` as a real-mainnet reproduction. Both would be useful as fixture fodder for the Phase 3.1 polish follow-up issue you filed — one tests `swap-x-for-y` direction with my own Genesis agent, the other tests the same shape with a different sender. Pinning the Hiro response JSON for both gives durable parser-regression coverage.

## v179 implementor-cites-reviewer cadence — final form

Three review cycles on this PR, each tighter than the last:
- **04:24Z → 04:30Z (6min)** — Initial review surfaced STX event_type bug + 3 non-blocking notes
- **04:30Z → 04:46Z (16min)** — STX fix + 3-suggestions absorbed, my review flipped to APPROVED
- **05:54Z → 06:18Z (24min)** — Empirical end-to-end test surfaced pending-cache trap, fix landed, verified live

22min PR-to-second-APPROVE on the bigger fix, then another tight cycle on the post-merge-window finding. The reusable shape: **small substantive reviews → empirical reproduction on preview → maintainer pushes targeted fix → reviewer re-verifies on the same preview**. Whole loop sub-30min per finding.

**Re-APPROVED on the current head.** Ready to merge as the Phase 3.1 verifier — schema-stable, dev-council density at peak, three load-bearing fixes locked in via end-to-end live test.

secret-mars (approved):
**End-to-end verification of the 5-commit fix series — fully green. The 409-with-existing_row design is stronger than my `inserted: false` proposal. 🎉**

Just tested the duplicate-submit path on the preview deploy:

### Duplicate POST (the case my prior comment flagged):

```
HTTP: 409
{
  "error": "Transaction already verified for this competition",
  "code": "txid_already_verified",
  "retryable": false,
  "existing_row": {
    "txid": "0xfa62f847…baf3c1",
    "sender": "SP20GPDS5RYB2DV03KG4W08EG6HD11KYPK6FQJE1",
    "contract_id": "...stableswap-stx-ststx-v-1-2",
    "function_name": "swap-x-for-y",
    "token_in": "stx",                ← STX fix still landing
    "amount_in": 499750,
    "token_out": "...ststx-token::ststx",
    "amount_out": 429262,
    "burn_block_time": 1778478970,
    "tx_status": "success",
    "source": "agent",                ← caller knows which ingestion path won
    "scored_value": null,
    "scored_at": null
  }
}
```

### 404 path (non-existent txid):

```
HTTP: 404
{
  "error": "Hiro returned 404 for 0x00…00",
  "code": "tx_not_found",
  "retryable": false
}
```

Both rejection paths now use the same structured-error shape (`{error, code, retryable, [existing_row?]}`). Consistent + machine-readable.

## Why 409 > `inserted: false` (your call was the right one)

I'd proposed `200 + inserted: false` — REST-idempotent. You went with `409 + structured error`. Three reasons your version is better:

1. **Consumers checking `response.ok` (HTTP 2xx)** see a difference automatically. The 200+flag version requires the consumer to read the body. Many clients only act on status code; some bail on 4xx without parsing the body — but they'd still get the right "don't double-submit" signal from 409.
2. **`existing_row.source` surfaces cross-ingestion observability cleanly**. Agent submits → cron beat me → response shows `source: "cron"` → agent learns "the row was recorded via cron-side-scan, not my submit." Pairs perfectly with the cron-as-defense-in-depth design.
3. **`code: "txid_already_verified"` is searchable + i18n-able**. Frontend code can branch on the stable enum without parsing the human-readable `error` string.

## Read-before-Hiro optimization (commit `2a988e1`) is a bonus win

`readSwap` BEFORE Hiro fetch in `verify.ts` means duplicate POSTs cost 1 D1 read, not 1 Hiro fetch. At scale (e.g., a stuck agent retry loop), this saves Hiro free-tier quota proactively. Pure win, no behavior change for fresh-tx callers.

## KV pending machinery removed entirely (`337a8c9`)

You went further than my v193 fix proposal — I'd just suggested dropping the read+short-circuit. You dropped the entire `comp:pending:{txid}` write/delete machinery. Makes sense per the MCP-pre-checks-now framing: the only path where 202 is returned is the racy "Hiro hasn't propagated terminal state yet" edge, with a clear note in the body. Net simpler, fewer KV writes, no observability artifact to maintain.

The 202 body change is also a nice touch:
```
{ "accepted": true, "note": "Hiro has not yet propagated this tx as terminal. Retry in a few seconds." }
```

The `note` field tells the consumer *why* they got 202 instead of leaving them to guess. Pairs with the existing description in the route's selfDoc.

## 4-stage lifecycle test (`e1de822`) pins the contract structurally

```ts
it("4-stage lifecycle: pending → pending → verified (200) → re-submit (409)", async () => {
  // POST 1 → 202 (pending)
  // POST 2 → 202 (still pending)
  // POST 3 → 200 (tx confirmed, fresh write)
  // POST 4 → 409 (row exists, no second write)
  expect(verifyAndPersistSwap).toHaveBeenCalledTimes(4);
});
```

This is the test fixture I wished I'd named in my prior comment. Locks in the per-POST behavior in a single sequence — any future refactor that re-introduces the short-circuit (or adds new short-circuit paths) breaks this test. Worth promoting as the canonical "is the POST handler still doing what it should" smoke for future PRs.

## MCP-side update suggestion (per your comment)

```ts
if (res.status === 200) {
  return { ...res.body, justSubmitted: true };
}
if (res.status === 409) {
  return { ...res.body.existing_row, justSubmitted: false };
}
```

I'll mirror this on the [aibtc-mcp-server #510 side](https://github.com/aibtcdev/aibtc-mcp-server/pull/510) — the `competition_submit_trade` tool description needs to update to mention the 409 path + the `justSubmitted` flag pattern. Small follow-up PR after #510 merges. Want me to file as an issue so it doesn't get forgotten?

## Net delta this PR has absorbed from review

Four post-empirical-test findings, each closed in <30min:

| # | Finding | Resolution |
|---|---|---|
| 1 | STX `event_type === "stx_asset"` not "stx_transfer_event" | `f6169405` + `447bf28f` (Set + 8 fixtures + variant-coverage `it.each`) |
| 2 | Pending-cache short-circuit blocked re-verification | `de52cd75` + `c44ba1e8` (straight removal + regression tests) |
| 3 | Re-POST returned silent 200 (no idempotency signal) | This 5-commit series — 409 + structured error + existing_row + lifecycle test |
| 4 | scored_value/scored_at columns Phase-3.2 forward-looking | Open architectural question — file separately if useful |

Each fix tighter than the last, each test pin pattern carrying forward. The substrate this PR ships is now substantially harder than what was there pre-review.

**Re-APPROVED on current head (`344df7bb`).** Maintainer queue ball, ready to ship.

cc @biwasxyz @arc0btc @whoabuddy — final round on the substantive review loop.

secret-mars (approved):
**Re-APPROVE — `37f53c6a feat(competition): GET /api/competition/allowlist` closes the operational gap I surfaced empirically in [#743#issuecomment-4419221706](https://github.com/aibtcdev/landing-page/pull/743#issuecomment-4419221706).** 90 LOC, single new file, no other surface changed.

Verified live against `feat-competition-read-routes-landing-page.hosting-962.workers.dev/api/competition/allowlist`:

```json
{
  "entries": [
    { "contract_id": "SPQC38PW542EQJ5M11CR25P7BS1CA6QT4TBXGB3M.stableswap-stx-ststx-v-1-2", "functions": ["swap-x-for-y", "swap-y-for-x"] },
    { "contract_id": "SM1793C4R5PZ4NS4VQ4WMP7SKKYVH8JZEWSZ9HCCR.xyk-core-v-1-1", "functions": ["swap-x-for-y", "swap-y-for-x"] },
    { "contract_id": "SM1793C4R5PZ4NS4VQ4WMP7SKKYVH8JZEWSZ9HCCR.xyk-swap-helper-v-1-3", "functions": ["swap-helper-a", ..., "swap-helper-e"] },
    { "contract_id": "SM1FKXGNZJWSTWDWXQZJNF7B5TV5ZB235JTCXYXKD.dlmm-swap-router-v-1-1", "functions": ["swap-simple-multi"] }
  ],
  "total_contracts": 4,
  "total_functions": 10,
  "provider_address": "SP1M8KHCJXB3SBRQRDBCG3J3859AA1CN0AWDHN17B",
  "protocols": { "bitflow": 4 }
}
```

`?docs=1` payload is the cleanest self-doc shape I've seen on this codebase — relatedEndpoints cross-references + the three notes (ALEX/Zest scope, code-review-mutation, provider-address-NOT-a-gate) give a consuming agent everything they need to wire this up correctly.

## What works particularly well

- **[CACHE_INVARIANTS:POSTURE=public-only-get marker](https://github.com/aibtcdev/landing-page/blob/37f53c6a/app/api/competition/allowlist/route.ts#L1)** — the v167-v173 posture-marker pattern lands its first new-route instance after Phase 2.5 cluster hygiene. Reader knows immediately this is a no-auth, no-branching endpoint.
- **`Cache-Control: public, max-age=3600, s-maxage=86400, stale-while-revalidate=86400`** — exactly right for a code-reviewed-only static export. 24h shared cache + SWR matches the deploy cadence; 1h browser tolerates the rare same-day re-deploy.
- **Single source of truth via `import { BITFLOW_ALLOWLIST, COMPETITION_ALLOWLIST } from "@/lib/competition/allowlist"`** — no duplication; whenever the allowlist module changes, the endpoint reflects it on next deploy.
- **`provider_address` clearly framed as audit-signal-not-gate** — closes a potential misread where consumers might assume "if provider tag is present, swap counts." The note "the only authoritative check is the (contract, function) tuple" is unambiguous.

## Non-blocking observations

### 1. No `version` field (consumer cache-invalidation hint)

The current shape forces an agent that wants to detect drift to either (a) re-fetch the full entries[] and diff, or (b) trust the 24h cache window. A `version` field — either a content hash or a manually-bumped integer that increments on each allowlist PR — would let the agent cache cheaply with cheap invalidation:

```ts
// proposed addition to the main response payload
{
  entries: [...],
  total_contracts: 4,
  total_functions: 10,
  version: "sha256:abc12345…",  // or version: 7 (manually bumped)
  // ...
}
```

The hash variant is zero-maintenance:

```ts
import { createHash } from "node:crypto";
const version = createHash("sha256")
  .update(JSON.stringify(COMPETITION_ALLOWLIST))
  .digest("hex")
  .slice(0, 8);
```

Costs ~3 LOC. Optional; the consumer-side workaround (just fetch + compare) is trivial today at 4 entries / 10 functions.

### 2. `contractFunctions` flat map for O(1) consumer lookup (low priority)

The canonical consumer pattern is "given a (contract_id, function_name), is it accepted?" Current shape requires the agent to loop entries[] then loop functions[]. A pre-computed flat map adds zero ambiguity and saves the consumer-side iteration:

```ts
// optional additional field
contract_functions: {
  "SPQC38PW542EQJ5M11CR25P7BS1CA6QT4TBXGB3M.stableswap-stx-ststx-v-1-2::swap-x-for-y": true,
  "SPQC38PW542EQJ5M11CR25P7BS1CA6QT4TBXGB3M.stableswap-stx-ststx-v-1-2::swap-y-for-x": true,
  // ...
}
```

Strictly redundant with entries[] (consumers can derive it themselves), so a "convenience" field only. Skip if you'd rather keep the response lean.

### 3. `?docs=true` / `?docs=yes` fall through to data response

Only `?docs=1` (literal string `"1"`) triggers the self-doc payload — `?docs=true`, `?docs=yes`, `?docs` (presence-only), or `?docs=docs` all hit the data path. This is fine if intentional (strict literal match keeps it deterministic) but a one-line comment explaining the strict-literal-1 contract would help the next reader. Or widen the predicate:

```ts
const docs = searchParams.get("docs");
if (docs === "1" || docs === "true" || docs === "") return selfDocResponse();
```

Pedantic — current behavior is documented in the `queryParameters.docs.example` payload field.

### 4. `openapi.json` entry follow-up

Spot-checked the preview's `/api/openapi.json` — `/api/competition/allowlist` not yet listed there (the openapi includes inbox/outbox/competition-trades but not allowlist). Presumably a follow-up commit; flagging so it doesn't drop off.

## Cross-cutting: this unblocks `aibtcdev/aibtc-mcp-server#510` wire-up

The operational gap I named in the [#743 empirical follow-up](https://github.com/aibtcdev/landing-page/pull/743#issuecomment-4419221706) — Bitflow MCP routing through non-allowlisted wrappers — now has a clean resolution path. The trading-comp tools in mcp-server#510 can:

1. **On Bitflow MCP boot:** `GET /api/competition/allowlist`, cache for 24h (matching s-maxage)
2. **Before each `bitflow_swap`:** consult the cached allowlist to filter `bitflow_get_routes` results — only execute routes whose final contract+function is on the list
3. **OR pre-validate the chosen route:** before broadcast, simulate which contract the swap will land in; refuse to broadcast if not allowlisted, give the agent a structured error explaining why

I'll file a `aibtcdev/aibtc-mcp-server#510` follow-up after #738 + #743 land suggesting wire-up shape. Roughly ~30-50 LOC in the Bitflow tool integration plus a small `competition-allowlist.ts` cache helper.

## Net

APPROVE. Closes the operational gap empirically surfaced 30 min ago. Non-blocking observations are pure polish — happy to file them as a follow-up convention-refinement issue alongside the Spark/Forge ones from `#727` substrate, OR fold into mcp-server#510 wire-up where they're more naturally exercised.

Maintainer ball remains @whoabuddy on merging the full PR group. With this commit landed, the surface is:
- ✅ Phase 3.1 verifier (POST /api/competition/trades)
- ✅ Read routes (GET /api/competition/trades + /status)
- ✅ Allowlist discoverability (GET /api/competition/allowlist) ← this commit
- ✅ Chainhook + cron
- ✅ Success-only gate (test coverage per b6eb2c8e)
- ✅ 409 + structured error on duplicate-submit

secret-mars (approved):
**bed7cd0 APPROVE** — re-review of the comp-start gate commit since my last APPROVE on `37f53c6a`.

The gate has full producer/consumer/fixture symmetry — read and write sides can't diverge silently here:

- **producer** `lib/competition/verify.ts:282-294` — 13 LOC inline gate after success-only check, before parse/DB work
- **consumer** `app/api/competition/trades/route.ts:335` — `before_comp_start` joined the permanent-rejection switch group (→ 422, `retryable: false`)
- **fixtures** `lib/competition/__tests__/verify.test.ts:236-294` — boundary acceptance at exact start, pre-start reject, ordering vs `tx_failed`, ordering vs sender/allowlist (4 tests)

Same producer/consumer-symmetric coverage I was naming on [#765 (issuecomment-4427022875)](https://github.com/aibtcdev/landing-page/issues/765#issuecomment-4427022875) but applied to a gate, not an adaptive interval — and exactly the symmetric shape that prevents the "read+write sides diverged silently" failure mode.

**Cron-side composition note (relevant to #765's out-of-scope list):** This gate makes the `CompetitionVerifierDO` design simpler. The cron's catch-up sweep over `registered_wallets` will pull pre-start tx history from Hiro for any wallet active before campaign start; that history flows through `verifyAndPersistSwap`, hits the comp-start gate, and gets rejected — so the cron doesn't need its own comp-start awareness or a "skip pre-start chunks" cursor optimization. Centralizing the policy at the `verify` chokepoint means #765's cursor logic can stay shape-pure (just chunk + sweep), with the gate filtering at the consumer end. Worth a one-liner in the #765 implementation PR so this composition is explicit.

**Tiny nit (non-blocking):** The `txBurnTime = tx.burn_block_time ?? 0` fallback at `verify.ts:286` produces a rejection reason `"Trade burn_block_time 0 is before competition start 1778630400"` for the impossible case where a `tx_status === "success"` tx somehow has undefined `burn_block_time`. Hiro guarantees that field is populated for confirmed txs, so the `?? 0` is unreachable in practice. Defensible as-is; only worth changing if you want the rejection reason to be self-explanatory under the impossible case.

Reaffirm APPROVE. Per arc's [mcp#510 03:21Z merge-order note](https://github.com/aibtcdev/aibtc-mcp-server/pull/510): #738 first, then mcp#510.

secret-mars (approved):
**5224a0d9 APPROVE** — re-review of the cursor KV→D1 refactor since my [v243 APPROVE on bed7cd0](https://github.com/aibtcdev/landing-page/pull/738#pullrequestreview-4268963257). Direction matches the #765 closure rationale — state-store and scheduler-primitive selection are now orthogonal, cleaner than the DO-alarm-rework path.

**Three things land well:**

**1. Generic schema, no backward-compat shim.** `competition_state(key, value, updated_at)` absorbs future signals (`last_run_at`, `consecutive_failures`) without per-signal migrations. `state.ts` exports the 3-op surface `getCronCursor` / `setCronCursor` / `clearCronCursor` so `cron.ts` only changes its import + 3 call-sites. The `cursorKey` test-isolation option deleted cleanly — no half-migrated state.

**2. Test isolation via per-test `cursorStore` closure** (`cron.test.ts:33-37`). Each `makeDb()` call gets its own cursor state — no shared-singleton risk. `cursorOps.set` / `cursorOps.clear` spies allow assertion on write-through. The SQL-prefix discrimination in the prepare-mock (`cron.test.ts:42-78`) couples brittlely to the exact statement text in `state.ts:21-39`, but the SQL is small enough that drift is immediately detectable.

**3. UPSERT semantics correct.** `ON CONFLICT(key) DO UPDATE SET value = excluded.value, updated_at = excluded.updated_at` with `unixepoch()` default — atomic single-statement update, matches the D1-pk patterns from the #672 / #705 lineage.

**One observation worth a description one-liner (non-blocking):** Production has the legacy KV `comp:cron:cursor` key under `VERIFIED_AGENTS` from prior deploys. After this lands, that key is orphaned — but the cron's first post-cutover run starts from `null` (head of `registered_wallets`), walks the page, and re-encounters previously-known txids that no-op at the D1 layer via `INSERT OR IGNORE`. Restart-from-head is correctness-safe — just one wasted sweep's worth of Hiro calls. Worth one sentence in the description so reviewers don't have to derive it. (Arc made the same idempotency observation on #765 before closure.)

Reaffirm APPROVE. Merge gate per arc's [mcp#510 03:21Z note](https://github.com/aibtcdev/aibtc-mcp-server/pull/510): this PR first, then mcp#510.

secret-mars (approved):
APPROVE on `9afa89d1` — rebase + SchedulerDO integration is the right architectural pivot, cleanly absorbs the lp#784 admin-surface foundation, and removes the external-auth attack surface (CRON_SECRET).

**What works well:**

1. **Migration 009 generic K/V scratchpad design** (`migrations/009_competition_state.sql`): `competition_state(key, value, updated_at)` is intentionally generic — future scheduler state reuses the same table rather than spawning one tiny migration per signal. Matches the "one schema, many keys" pattern from the inbox-events table. Avoids the per-signal-migration anti-pattern. Smart.

2. **Strict-task-validation continuity** (`app/api/admin/scheduler/route.ts:8`): `ALLOWED_TASKS` Set extended `"tenero" | "all"` → `"tenero" | "competition" | "all"` with the error message updated. Direct lineage from lp#784's 1902007 absorption pass — no drift between sibling PRs in the same campaign. The required-by-default convention from my v296 Field 8 proposal also applies here: extending an allowlist is the right shape, vs adding new untyped task identifiers.

3. **DO-bookkeeping vs D1-cursor separation** (`worker.ts` comment block + `lib/competition/scheduler.ts`): "Long-lived cursors stay in D1 per issue #768 — the DO holds only its own bookkeeping." Clean separation of concerns. The DO holds `lastCompetitionRunAt`, `lastCompetitionResult`, `consecutiveFailures.competition`, `nextRunAfter.competition` (transient, scheduler-internal); D1 `competition_state` holds the durable cursor. Recovery-after-DO-loss only requires the D1 cursor, not the DO scratchpad.

4. **Cadence separation** (`worker.ts:54`): `COMPETITION_INTERVAL_MS = 15 * 60 * 1000` distinct from `TENERO_INTERVAL_MS = 5 * 60 * 1000`. Different per task, with `ALARM_TICK_MS = TENERO_INTERVAL_MS` (smallest cadence drives the alarm); per-task `nextRunAfter` decides which tasks actually fire each tick. Adaptive backoff per task is preserved via `Partial<Record<SchedulerFailureTask, number>>`.

5. **CRON_SECRET removal**: smaller attack surface. Previously a shared-secret over an unauthenticated route gated external cron callers; post-merge, the only triggers are SchedulerDO alarm (internal-only) + admin route (`X-Admin-Key` gated per lp#784). The auth boundary is now homogeneous with other admin surfaces.

**[observation] Alarm scheduling logic I didn't drill into**

I didn't trace the full alarm()-tick scheduling logic for the per-task `nextRunAfter` comparison — i.e. how each 5min tick decides whether to run tenero only, competition only, or both. The cadence design (COMPETITION 15min, TENERO 5min, ALARM_TICK 5min) implies tenero runs every tick + competition runs every 3rd tick (modulo backoff), but the exact decision shape isn't in my read. The test coverage (`lib/competition/__tests__/scheduler.test.ts` per the file list) presumably handles this; flagging only because the comment block describes the state shape but doesn't quite spell out the per-task gate logic. Worth a short inline comment in the alarm() body if it's not already there.

**[observation] Migration ordering**

Migration 009 (`competition_state`) is the next sequential after the 762c migrations. Worth confirming it gets applied alongside (or before) any downstream PR that reads from `competition_state`. The PR body should probably note the apply-ordering for any operator running fresh deploys.

**Cross-PR substrate (continuing):**
- lp#784 admin scheduler surface absorbed cleanly — `ALLOWED_TASKS` extension is the convention-conformant pattern.
- The lp#786 Field 8 proposal (optional-vs-required new param shape on lib helpers) doesn't strictly apply here since `task: SchedulerTask` is already required-typed — but the discriminated-union shape of `SchedulerTask = "tenero" | "competition" | "all"` IS the type-level enforcement that Field 8 proposes for new params.
- #691 / claim-source asymmetry from #785 doesn't reach competition surface (different code path).

**CI:** 7/8 checks SUCCESS (Lint, CodeQL, Test, Workers Build, Snyk, Analyze actions/js-ts; Build SKIPPED is expected for fork-PR-style). v124 head-SHA pre-submit clean on `9afa89d1`.

Ready to merge for Phase 3.1 launch. Production catch-up cron is the load-bearing user-facing piece — once this lands, the trade leaderboard surface on `/leaderboard` (lp#743 merged earlier today) gets its Volume column live data source.

PR #742, opened 2026-05-11
feat(leaderboard): trading-comp score page w/ historical P/L from Tenero OHLC + 30m cron

biwasxyz:
## Summary

Builds the trading-comp leaderboard on top of PR #738's verifier surface:

- New `/leaderboard` UX page + `/api/leaderboard/score` AX read route serving a cron-built KV snapshot (pure-KV on the request path, no upstream fetches).
- Per-agent historical USD P/L: each swap leg is priced against the Tenero OHLC close for the 1h bucket containing the swap's `burn_block_time` — not against today's prices. Tenero exposes the OHLC for free so this was cheaper to do than to ship the current-price v1.
- Cloudflare Cron Trigger (`*/30 * * * *`) wired via `wrangler.jsonc` triggers.crons + `scheduled()` handler in `worker.ts`. The HTTP refresh route (`X-Cron-Secret`-gated) stays as a manual escape hatch.

**Base = `feat/competition-read-routes` (PR #738)** — preview URL has the verifier (which persists swaps to D1) AND the leaderboard (which reads them) wired together end-to-end.

## Why historical P/L instead of current-price

Probed Tenero before committing and found three bugs in the v1 I had drafted: wrong base URL (`/api/v1/stacks/...` → `/v1/stacks/...`), wrong response parsing (`body.price_usd` → `body.data.price_usd`), and three TOKEN_DECIMALS entries pointing at addresses Tenero doesn't recognize (test/v0 contracts). Once fixing the base URL anyway, pivoting to OHLC was the same cost — and it eliminates the "today's price × ancient amount" honesty problem.

## Architecture

**Cron tick (every 30 min, via `scheduled()` handler):**
1. `listSuccessSwapsWithAgentFromD1` — read every success swap from D1 with agent display JOIN.
2. Compute `[fromTs, toTs] = [min(burn_block_time), now]` (floored to 30-day lookback when D1 is empty).
3. `refreshAllHistories` — for each token in `TOKEN_DECIMALS` (STX + sBTC + stSTX), fetch OHLC at 1h period covering the window. Write per-token blob to KV `comp:price-history:{assetId}`.
4. `aggregateLeaderboard` — group by sender, for each leg call `priceAt(history, burn_block_time)` for O(1) bucket lookup. Trades with a missing candle on either leg are counted but skipped from `pnl_usd` (preserves leaderboard honesty — unpriced ≠ \$0).
5. Sort `pnl_usd desc → trade_count desc → first_trade_at asc`.
6. Write snapshot to KV `cache:leaderboard:score` (TTL 2h = 4 ticks of slack).

**Read path:**
- `GET /api/leaderboard/score` — 1 KV read. Empty-snapshot fallback returns the same schema with `rows: []` and `cachedAt: null`.
- `/leaderboard` server-renders the same snapshot directly via `readLeaderboardSnapshot`, no API hop.

**UI surfaces freshness:** `LeaderboardList` shows `cachedAt` as relative time + "refreshes every 30 min" caption, re-rendering every 30s.

## TOKEN_DECIMALS scope discipline

Intentionally trimmed to **STX, sBTC, stSTX** — only entries probed against Tenero and confirmed 200. ALEX, USDh, wstx have multiple variants on chain and shipping the wrong contract id silently corrupts the leaderboard (every leg involving that token reads as unpriced). Comment in `decimals.ts` requires a Tenero probe before adding tokens back.

Trades involving tokens outside this set still appear in `trade_count` with a `(N priced)` callout, but are excluded from `pnl_usd`. Network-stats card shows the priced-vs-total coverage %.

## Cron caveat (env split)

Preview + production share the same KV namespace + D1 database, so the preview cron writes to production's `cache:leaderboard:score` slot every 30 min. Same source data + same code = same output, so it's not destructive — but if envs ever split, the cron config has to split too. Noted inline in `wrangler.jsonc`.

## Test plan

- [ ] Preview URL: `/leaderboard` renders with the "first cron run pending" empty state (until the first scheduled tick fires)
- [ ] After first cron tick: rows show ranked agents with `cachedAt` updated; `stats.price_window_from/to` covers the swap range
- [ ] Hit `GET /api/leaderboard/score` — verify schema matches the self-doc shape (rows + cachedAt + stats + refresh_interval_seconds)
- [ ] Hit `GET /api/leaderboard/score?docs=1` — self-doc references historical-P/L (not current-price)
- [ ] `POST /api/competition/leaderboard/refresh` with `X-Cron-Secret` — manual rebuild works
- [ ] `POST` without secret → 401; with wrong secret → 401; without `CRON_SECRET` env → 500
- [ ] Tests for `prices.ts` (OHLC parse + bucketing), `pnl.ts` (historical lookup + unpriced skip), `leaderboard.ts` (window computation), refresh route (auth), read route (empty-snapshot fallback) — landing in a follow-up commit


arc0btc (changes requested):
Builds the trading-comp leaderboard on the snapshot-over-KV pattern from PR #651 — solid approach, request path stays pure-KV, Tenero OHLC pricing at burn_block_time is the honest choice over current prices. One blocking correctness risk to fix before merge.

**What works well:**
- Snapshot model is clean and consistent with the portfolio pattern. Four cron-tick TTL slack means transient upstream failures don't break the UI.
- D1 JOIN in `listSuccessSwapsWithAgentFromD1` eliminates N+1 agent lookups at the cron boundary — right call.
- Historical pricing via OHLC gives true P/L at swap time, not a misleading current-price × historical-amount mix.
- `refreshTokenHistory` correctly preserves the old history on failure (no write unless candles returned) — cron-resilient.
- `priceAt` null-guards correctly: missing bucket → null → skipped from P/L total, not imputed as zero. The unpriced vs priced split surfaced in the UI is exactly the right transparency signal.
- Cron handler swallows failures correctly so one bad tick doesn't fail the CF cron health check.

---

**[blocking] Stack overflow at scale — `Math.min(...swaps.map(...))` in `leaderboard.ts`**

`lib/competition/leaderboard.ts` (in `buildLeaderboardSnapshot`):

```ts
const earliestSwapTs = swaps.length > 0
  ? Math.min(...swaps.map((s) => s.burn_block_time))
  : nowTs - PRICE_HISTORY_DEFAULT_LOOKBACK_SECONDS;
```

Spreading a large array into `Math.min` puts every element on the call stack. V8 (CF Workers runtime) hard-limits the stack — at ~10k elements this throws `RangeError: Maximum call stack size exceeded`. The cron reads **every** success swap with no LIMIT, which is correct for full-history aggregation but means this will fail silently once the comp has meaningful volume.

```suggestion
  const earliestSwapTs = swaps.length > 0
    ? swaps.reduce((min, s) => Math.min(min, s.burn_block_time), swaps[0].burn_block_time)
    : nowTs - PRICE_HISTORY_DEFAULT_LOOKBACK_SECONDS;
```

The `fromTs` line below has the same fix applied for free once you use the reduce result.

---

**[suggestion] Negative P/L renders as `$-1,234.56` instead of `-$1,234.56`** (`app/leaderboard/LeaderboardList.tsx`)

`toLocaleString("en-US")` on a negative number produces the `-` sign before the digits (e.g., `"-1,234.56"`). Prepending `$` gives `$-1,234.56` — the dollar sign appears before the minus, which is non-standard and jarring for agents with losses (the primary display case of a P/L leaderboard).

```suggestion
function formatUsd(value: number | null | undefined, opts: { signed?: boolean } = {}): string {
  if (value == null || !Number.isFinite(value)) return "—";
  const prefix = opts.signed && value > 0 ? "+" : value < 0 ? "-" : "";
  const fractionDigits = Math.abs(value) < 10_000 ? 2 : 0;
  const formatted = Math.abs(value).toLocaleString("en-US", {
    minimumFractionDigits: fractionDigits,
    maximumFractionDigits: fractionDigits,
  });
  return `${prefix}$${formatted}`;
}
```

---

**[suggestion] Two-pass display-field stitching is an awkward seam** (`lib/competition/leaderboard.ts`)

`buildLeaderboardSnapshot` runs `aggregateLeaderboard(swaps, histories)` (which returns rows keyed only by `sender`) then re-walks the same `swaps` array to build `displayBySender` and stitch the JOIN columns back on. The D1 query already paid the JOIN cost to get those display fields — the split means they're discarded and re-derived in a second O(N) pass.

Consider extending `AgentScoreRow` in `pnl.ts` to carry the optional display fields, or passing display data as a separate `Map<sender, DisplayFields>` into `aggregateLeaderboard`. Either removes the double-walk and makes the data flow explicit.

---

**[suggestion] `<th>` elements missing `scope="col"`** (`app/leaderboard/LeaderboardList.tsx`)

Screen readers use `scope` to associate header cells with data cells in multi-column tables. Without it, VoiceOver and NVDA may not announce column headers when navigating table rows.

```suggestion
<th className="px-4 py-3 font-medium" scope="col">Rank</th>
<th className="px-4 py-3 font-medium" scope="col">Agent</th>
<th className="px-4 py-3 font-medium text-right" scope="col">Trades</th>
<th className="px-4 py-3 font-medium text-right" scope="col">Volume in</th>
<th className="px-4 py-3 font-medium text-right" scope="col">Volume out</th>
<th className="px-4 py-3 font-medium text-right" scope="col">P/L (USD)</th>
```

---

**[nit] Animated ping dot in hero badge needs `aria-hidden`** (`app/leaderboard/page.tsx`)

The pulsing `animate-ping` ring is purely decorative but renders two `<span>` elements inside the badge with no aria attributes. Add `aria-hidden="true"` to the outer wrapper to keep it out of the accessibility tree.

---

**[question] Preview cron vs. production KV slot**

The `wrangler.jsonc` comment acknowledges that preview and production share `cache:leaderboard:score`. This is fine while both envs run identical code against the same D1 + KV. Worth confirming: is there a flag or config that could put preview and production on diverging code paths where a preview cron tick would write a malformed/schema-mismatched snapshot into the production slot? If there's any risk of a deploy window where preview runs ahead-of-version, the KV key should be namespaced per-env.

---

**Code quality notes:**

`refreshAllHistories` catches per-token errors inside the `Promise.all` callback and also increments `summary.errors` — but `refreshTokenHistory` internally wraps the KV write in its own try/catch and returns null on failure without throwing. The outer `try/catch` in `refreshAllHistories` therefore only fires if the `refreshTokenHistory` call itself throws (a coding bug), not on Tenero failures. The error counting is correct but the code comments could make this clearer.

*Operational context:* We run Tenero OHLC polling in our own sensors and have seen the 8s timeout fire on slow candle windows. `FETCH_TIMEOUT_MS = 8_000` is reasonable, but the null-on-failure path means a slow response that times out will leave the previous cached history intact — which for a first-run deploy means no prices until a successful tick. Not a blocker, but worth a comment explaining this first-run gap in `refreshTokenHistory`.

arc0btc (changes requested):
Re-review (cycle 2) — the cron-trigger wiring and `CountdownToNextTick` are a nice UX improvement, but the cycle-1 blocking fix wasn't shipped, and the new component introduces a second blocking issue.

**What's improved:**
- `AgentScoreRow` now carries display fields — partially collapses the double-walk concern (the Map is still built inline, but the data flow is clearer now).
- `CountdownToNextTick` isolated correctly so the 1Hz timer doesn't re-render the whole tree.
- Cron trigger wired in `wrangler.jsonc` and `scheduled()` handler — clean separation from the HTTP escape-hatch route.

---

**[blocking] Math.min spread — still unaddressed** (`lib/competition/leaderboard.ts:126`)

This was raised in cycle 1. The spread-into-variadic pattern is still there:

```ts
const earliestSwapTs = swaps.length > 0
  ? Math.min(...swaps.map((s) => s.burn_block_time))
  : nowTs - PRICE_HISTORY_DEFAULT_LOOKBACK_SECONDS;
```

V8 hard-limits the call stack — this throws `RangeError: Maximum call stack size exceeded` at ~10k elements. The cron reads every success swap with no LIMIT, so a comp with meaningful volume will silently fail here.

```suggestion
  const earliestSwapTs = swaps.length > 0
    ? swaps.reduce((min, s) => Math.min(min, s.burn_block_time), swaps[0].burn_block_time)
    : nowTs - PRICE_HISTORY_DEFAULT_LOOKBACK_SECONDS;
```

---

**[blocking] `CountdownToNextTick` hydration mismatch** (`app/leaderboard/LeaderboardList.tsx`)

`Date.now()` in a `useState` initializer runs during SSR and again during hydration. Because time passes between the two invocations, the server renders `30:00` and the client hydrates to `29:58` — React detects the mismatch and overrides the DOM, causing a content flash. This matches our `no-date-ssr` rule: clock reads in render/state-init belong in `useEffect`.

```suggestion
function CountdownToNextTick({ periodSeconds }: { periodSeconds: number }) {
  const [now, setNow] = useState<number | null>(null);
  useEffect(() => {
    setNow(Date.now());
    const t = setInterval(() => setNow(Date.now()), 1000);
    return () => clearInterval(t);
  }, []);

  if (now === null) return null;
  const unixSec = Math.floor(now / 1000);
  const secondsLeft = periodSeconds - (unixSec % periodSeconds);
  const mm = Math.floor(secondsLeft / 60);
  const ss = secondsLeft % 60;
  return (
    <span className="font-mono tabular-nums text-white/90">
      {String(mm).padStart(2, "0")}:{String(ss).padStart(2, "0")}
    </span>
  );
}
```

Returning `null` before mount means the badge briefly shows the static text without the countdown, then fills in after hydration — no flash, no warning.

---

**[suggestion] `formatUsd` still produces `$-1,234.56`** (`app/leaderboard/LeaderboardList.tsx:22`)

The current implementation:

```ts
const sign = opts.signed ? (value > 0 ? "+" : value < 0 ? "" : "") : "";
const formatted = value.toLocaleString("en-US", { ... });
return `${sign}$${formatted}`;
```

When `value` is negative, `sign` is `""` (empty), but `value.toLocaleString("en-US")` on a negative number returns `"-1,234.56"` — the toLocaleString minus sign is still in `formatted`. Prepending `$` gives `$-1,234.56`. Fix: pass `Math.abs(value)` to `toLocaleString` and extract the sign explicitly.

```suggestion
function formatUsd(value: number | null | undefined, opts: { signed?: boolean } = {}): string {
  if (value == null || !Number.isFinite(value)) return "—";
  const prefix = opts.signed && value > 0 ? "+" : value < 0 ? "-" : "";
  const fractionDigits = Math.abs(value) < 10_000 ? 2 : 0;
  const formatted = Math.abs(value).toLocaleString("en-US", {
    minimumFractionDigits: fractionDigits,
    maximumFractionDigits: fractionDigits,
  });
  return `${prefix}$${formatted}`;
}
```

---

**[suggestion] `<th>` elements still missing `scope="col"`** (`app/leaderboard/LeaderboardList.tsx:170–175`)

Screen readers use `scope` to associate header cells with data cells. This was noted in cycle 1.

```suggestion
<th className="px-4 py-3 font-medium" scope="col">Rank</th>
<th className="px-4 py-3 font-medium" scope="col">Agent</th>
<th className="px-4 py-3 font-medium text-right" scope="col">Trades</th>
<th className="px-4 py-3 font-medium text-right" scope="col">Volume in</th>
<th className="px-4 py-3 font-medium text-right" scope="col">Volume out</th>
<th className="px-4 py-3 font-medium text-right" scope="col">P/L (USD)</th>
```

---

**[nit] `animate-ping` wrapper still needs `aria-hidden`** (`app/leaderboard/page.tsx:58–61`)

The pulsing dot renders two decorative `<span>` elements with no ARIA attributes. Add `aria-hidden="true"` to the `relative flex` wrapper so screen readers skip the animation entirely.

---

*Three items from cycle 1 remain open: the stack overflow (blocking), the P/L sign formatting, and the table headers. The new countdown component is solid in concept but needs the hydration guard before merge.*

biwasxyz:
Closing — this turned into a much bigger surface than the underlying ask warranted. The actual scope (display per-agent portfolio + count of MCP-submitted swaps) is small enough to live as a follow-up on the /agents page without a separate PR base or scope-pivot history. Will reopen as a fresh narrow PR if/when needed.

PR #743, opened 2026-05-11
feat(leaderboard): /leaderboard page ranked by MCP-submitted trade count + USD volume

biwasxyz:
## Summary

Lands the **scheduler foundation** + **first task (Tenero token prices)** described in #768, and rewires `/leaderboard` to render USD volume from KV-cached prices instead of browser-side Tenero fetches.

Refs #768. Subsequent tasks (competition Hiro sweep, balance snapshots) ship as follow-ups on top of this scaffold.

### Scheduler foundation
- `SchedulerDO` Durable Object — single instance (`idFromName("v1")`) coordinating periodic background work via `alarm()`. RPC surface: `status()` / `refreshNow(task)` / `pauseUntil(ts)` / `resume()`.
- Class defined **inline in `worker.ts`** (not behind a separate-file import) so OpenNext's bundling preserves it cleanly. Verified via local `wrangler deploy --dry-run` + `wrangler.jsonc` migration history (`v1: new → v2: deleted → v3: new` — see `wrangler.jsonc` comment for rationale).
- Per-task try/catch inside `alarm()`. Failures log structured events (`tenero.*`, `scheduler.*`) via `env.LOGS` → `logs.aibtc.com`. Adaptive backoff: if Tenero returns `minute-remaining <= 0`, sets `nextRunAfter.tenero` so the next tick skips Tenero until that timestamp passes.
- `consecutiveFailures.tenero` tracked from day one; circuit-breaker behavior deferred until observability shows we need it (#768 alarm-failure policy section).

### Tenero task (first consumer)
- `lib/external/tenero-fetch.ts` — typed wrapper modeled on `lib/stacks-api-fetch.ts`. Small retry budget on purpose (`DEFAULT_429_RETRIES = 2`) because Tenero's shared-CF-egress-IP web-ui-ip tier (100/min) makes aggressive retry counterproductive — the next alarm tick is the recovery path.
- `lib/external/tenero/prices.ts` — `fetchTokenPriceUsd(tokenId, logger, apiKey)`. Never throws on non-2xx; surfaces rate-limit headers up to the scheduler.
- `lib/external/tenero/kv-cache.ts` — `tenero:price:{tokenId}` keys, 24h TTL ceiling as a safety net. SchedulerDO writes, SSR / `/api/prices` reads.
- Active token set locked to `STATIC_TOKEN_IDS` (mirrors `TOKEN_DECIMALS` in `app/leaderboard/page.tsx`) — adding a new priceable token is a deliberate two-step edit: decimals table + static list + a Tenero probe.

### Leaderboard SSR rewire
- `app/leaderboard/page.tsx` — D1 query for swap aggregates + `getCachedTokenPrices(kv, tokenIds)` for prices, then computes `volumeUsd` + `allPriced` per row **server-side**. Renders `—` for unpriced legs (honest under-report).
- `app/leaderboard/LeaderboardClient.tsx` — pure presentational now. No `useEffect`, no `localStorage`, no browser-side Tenero fetch. Partial totals render with `*` + tooltip.
- Opportunistic SchedulerDO kick on first SSR — `env.SCHEDULER.idFromName("v1").get().status()` fire-and-forget via `ctx.waitUntil`. The DO instantiation is what arms the first alarm.

### `/api/prices` route (item 6 of @whoabuddy's review)
- `GET /api/prices` — returns `{ prices: { [tokenId]: { priceUsd, fetchedAt } }, supportedTokens: STATIC_TOKEN_IDS }` for AX consumers.
- `GET /api/prices?token=stx` — single-token lookup.
- `GET /api/prices` with no `Accept: application/json` — self-doc, matches the pattern used by other AX routes.
- Rate-limited via the existing `RATE_LIMIT_READ` binding (300/min). No new namespaces needed.

## Constraints to know before / when merging

1. **Workers with DOs don't generate preview URLs** ([CF doc limitation](https://developers.cloudflare.com/workers/configuration/previews/#limitations)). No branch-preview testing for this PR — first observable deploy is the merge-to-main `wrangler deploy`.
2. **CI on this branch will keep failing 10211.** `wrangler versions upload` (the non-prod deploy command) cannot apply DO migrations — only `wrangler deploy` (main's deploy command) can. The Workers Build ❌ is expected, not a code defect. Lint / Test / CodeQL are green.
3. **Rollback path documented**: if anything's wrong post-merge, a `v4: deleted_classes` hotfix follows the same shape as #772.

See the [deploy mechanics comment](https://github.com/aibtcdev/landing-page/pull/743#issuecomment-4428791097) for the long version + lessons from the #772 incident.

## What this PR deliberately doesn't do

- **Competition Hiro sweep task** — deferred per #768. Lands on the same SchedulerDO scaffold in a follow-up.
- **Balance snapshots task** (the third task floated in [#768 comment](https://github.com/aibtcdev/landing-page/issues/768#issuecomment-4427891234)) — same scaffold, separate PR.
- **Tenero API key** — `TENERO_API_KEY` env hook is wired but unset. Unauthenticated 100/min/IP is fine for the static token set; raise to a real key if traffic grows.
- **`/agents` MCP-trade columns** (the original PR scope) — abandoned. `/leaderboard` is the surface.

## Test plan

- [x] Local `npm run build` passes
- [x] Local `wrangler deploy --dry-run --outdir=…` confirms `SchedulerDO` is exported from the bundled worker
- [x] Unit tests for `runTenero` happy / 5xx / 429-backoff paths + `tokenIdToTeneroAddress` + `getCachedTokenPrice` shape-mismatch (see review item 5)
- [ ] On merge: main's CI deploy runs `wrangler deploy`, applies `v3`, registers `SchedulerDO` on the `landing-page` worker
- [ ] First `/leaderboard` visit on `aibtc.com` triggers the opportunistic kick; alarm fires ~5 min later
- [ ] `logs.aibtc.com`: `tenero.refresh_started` → `tenero.refresh_completed` events with `succeeded > 0`
- [ ] `wrangler kv:key list --binding=VERIFIED_AGENTS --prefix='tenero:price:'` shows entries for `stx`, `SM…sbtc-token::sbtc`, `SP…ststx-token::ststx`
- [ ] `/leaderboard` Volume column populates with USD numbers (or `—` + `*` for the unpriced legs)
- [ ] `GET /api/prices` returns the cached price map; `?token=stx` returns the single entry; no-JSON-Accept returns the self-doc


arc0btc (approved):
Adds MCP trade count, USD volume, and latest-trade columns to `/agents` — well-scoped feature with clean architecture.

**What works well:**
- Single D1 `GROUP BY` query with parallel Tenero price calls is the right pattern; no N+1, no separate cron
- Graceful degradation via empty `Map` on D1 unavailability — the page still renders if the trades query fails
- `AbortSignal.timeout` on each Tenero fetch is clean and avoids hanging SSR
- Sort tiebreak (volume as secondary for equal trade counts) is a nice touch
- D1 migration work (Phase 2.5 Step 3.5): the `from_btc_address` SQL gate as tenant-discriminator is the right fix for the cross-agent duplicate-reply bug; 503 + `Retry-After: 5` pattern is consistent with the rest of the codebase

---

**[suggestion] `decimalsFor` fallback will silently miscalculate volume for unknown tokens** (`lib/competition/volume.ts:38`)

The code comment even warns about this — "defaults to 6 below if unset, which is wrong for sBTC and friends." If a new SIP-10 token enters the DB that Tenero can price but isn't in `TOKEN_DECIMALS`, the volume math uses 6 decimals regardless of the actual precision. For an 8-decimal token like sBTC, that's a 100× over-report.

Since you already have the "skip unpriceable tokens" path working, the simplest fix is to extend the same logic to tokens with unknown decimals:

```suggestion
function decimalsFor(assetId: string): number | null {
  return TOKEN_DECIMALS[assetId] ?? null;
}
```

Then in the aggregator:

```ts
const dec = decimalsFor(r.token_in);
if (price != null && dec != null) {
  const human = r.sum_in / 10 ** dec;
  existing.volumeUsd += human * price;
}
```

This way unknown tokens behave identically to unpriceable ones: counted, not volume-attributed. Honest under-report, same as the existing "token_in='unknown'" case.

---

**[suggestion] Volume formatting is duplicated** (`app/agents/AgentList.tsx`)

The `toLocaleString` block for USD formatting appears verbatim in the desktop table cell (~L508) and the mobile chip (~L631). Extract to a one-liner:

```ts
function formatUsdVolume(usd: number): string {
  return usd < 10_000
    ? usd.toLocaleString("en-US", { minimumFractionDigits: 2, maximumFractionDigits: 2 })
    : usd.toLocaleString("en-US", { maximumFractionDigits: 0 });
}
```

Also: the desktop cell uses `minimumFractionDigits: 2` but the mobile chip omits it — minor inconsistency for values like `$1.00` vs `$1`.

---

**[suggestion] MCP Trades and Volume columns should use tabular numbers** (`app/agents/AgentList.tsx`)

Number columns with variable-width digits look misaligned under sort. Add Tailwind's `tabular-nums` class to the count and volume `<span>` elements so digits occupy the same width.

```suggestion
<span className="text-[13px] font-medium tabular-nums text-[#F7931A]">
  {agent.mcpTradeCount}
</span>
```

---

**[nit] Unnecessary `as number` cast** (`app/agents/AgentList.tsx`)

`agent.mcpLatestTradeAt` is already typed as `number | undefined` and the `?? 0` guard makes it `number` — the `as number` cast on the next line is redundant.

---

**[question] PR scope**

This PR bundles the competition columns feature with Phase 2.5 Step 3.5 inbox/outbox D1 read-flip (~500 lines of auth-path migration). Is the base `feat/competition-read-routes` the right landing target for both, or is the Step 3.5 work intended to merge to main separately? The diff reads cleanly but the two areas are independent enough that separate PRs would make CI failures easier to attribute.

---

**Code quality notes:**

- `toTeneroAddress` handles empty string input gracefully (returns `''`) — Tenero call would 404 and price silently becomes `null`. Works but a one-line guard (`if (!assetId) return null`) would be cleaner.
- `result.results ?? []` is defensive-correct; D1's `.all()` spec says `results` is always present on success, but the `??` costs nothing and guards against edge-case undefined.
- 3 live Tenero fetches per `/agents` render is fine at current scale; the existing code comment about adding a cache is the right call-to-action.

secret-mars (commented):
**Substantive review — leading with the CI BLOCKER (concrete, fixable), then soft observations on the volume helper architecture.**

## [BLOCKER] Build is failing on ESLint `no-console` — three lines, one fix

Cloudflare Workers preview build (and the `Workers Builds: landing-page` check that's still pending on this PR — currently `mergeStateStatus: UNSTABLE`) fails at the lint stage:

```
./lib/competition/volume.ts
87:7   Error: Unexpected console statement.  no-console
100:5  Error: Unexpected console statement.  no-console
107:5  Error: Unexpected console statement.  no-console
```

Those map to the three `console.warn(...)` calls in `fetchTokenPriceUsd`:

- `competition.volume.tenero_non_ok` (non-2xx response)
- `competition.volume.tenero_no_price` (unparseable / zero / null price)
- `competition.volume.tenero_threw` (timeout / fetch rejected)

The project's canonical replacement is [`createConsoleLogger` from `@/lib/logging`](https://github.com/aibtcdev/landing-page/blob/3a99da4/lib/logging.ts#L95) — the standard fallback when `env.LOGS` RPC binding isn't available. Two viable shapes:

**(A) Thread a logger into the helper signature** — preferred, lets the page.tsx caller pass `env.LOGS`-backed logger when available, `createConsoleLogger` fallback otherwise:

```ts
// lib/competition/volume.ts
import { createConsoleLogger, type Logger } from "@/lib/logging";

async function fetchTokenPriceUsd(
  assetId: string,
  logger: Logger = createConsoleLogger({})
): Promise<number | null> {
  // ... replace console.warn(...) with logger.warn(...) at lines 87, 100, 107
}

export async function getAgentSubmittedTradeSummary(
  db: D1Database,
  logger: Logger = createConsoleLogger({})
): Promise<Map<string, AgentTradeSummary>> {
  // ... pass logger through to fetchTokenPriceUsd ...
}

// app/agents/page.tsx (caller)
const logger = env.LOGS
  ? createLogger(env.LOGS, ctx, { path: "/agents", scope: "competition.volume" })
  : createConsoleLogger({ path: "/agents", scope: "competition.volume" });
const tradeSummary = env.DB
  ? await getAgentSubmittedTradeSummary(env.DB, logger)
  : new Map();
```

**(B) Module-level logger** — simpler but loses the SSR-context rayId/path correlation:

```ts
import { createConsoleLogger } from "@/lib/logging";
const logger = createConsoleLogger({ scope: "competition.volume" });
// replace all three console.warn → logger.warn
```

(A) is the right shape per the [`createLogger` docstring pattern at lib/logging.ts:10-22](https://github.com/aibtcdev/landing-page/blob/3a99da4/lib/logging.ts#L10) — when `env.LOGS` is bound the warns land in the worker-logs service rather than just the Cloudflare tail console. ~6 LOC added, 3 LOC changed.

Once this is fixed, CI should go green and the soft observations below become the conversation.

---

## What's clean (architecture wins worth keeping)

- **Single D1 GROUP BY query** ([`volume.ts:154-163`](https://github.com/aibtcdev/landing-page/blob/feat/agents-mcp-trades-volume/lib/competition/volume.ts#L154)) — `SELECT sender, token_in, COUNT, SUM, MAX FROM swaps WHERE source='agent' GROUP BY sender, token_in`. One round-trip, all aggregation in JS. Matches the v183 framing I argued for on #651 ("JIT-computable at current scale, no need to amortize scoring across cron").

- **Parallel Tenero pricing deduplicated by token_in** ([`volume.ts:174-178`](https://github.com/aibtcdev/landing-page/blob/feat/agents-mcp-trades-volume/lib/competition/volume.ts#L174)) — one call per distinct token regardless of sender count. `Array.from(new Set(...))` + `Promise.all` is the right shape.

- **`AbortSignal.timeout(5_000)`** ([`volume.ts:78`](https://github.com/aibtcdev/landing-page/blob/feat/agents-mcp-trades-volume/lib/competition/volume.ts#L78)) — bounded per-token latency. Three parallel 5s timeouts is acceptable SSR cost.

- **`token_in='unknown' counts but doesn't price`** — semantic discipline. Honest under-report rather than imputing zero or fake-USD. The docstring at lines 17-20 makes this explicit; the UI at AgentList.tsx hides empty values via `(agent.mcpVolumeUsd ?? 0) > 0` ternaries.

- **`TOKEN_DECIMALS` pinned with provenance comment** ([`volume.ts:34-44`](https://github.com/aibtcdev/landing-page/blob/feat/agents-mcp-trades-volume/lib/competition/volume.ts#L34)) — adding a new token requires probing Tenero first. Same shape as the v167 scout-pre-position pattern from Phase 2.5.

## Non-blocking observations

### 1. No unit tests for `lib/competition/volume.ts` — v137 family (claim-without-test)

The helper makes **at least 7 specific behavioral claims** in its docstrings:

- (a) Only `swaps.source = 'agent'` rows are counted (excludes cron / chainhook)
- (b) Volume is input-side only, no double-count of out-leg
- (c) Native STX passes through as literal `"stx"` (not the `wstx` synthetic)
- (d) Unpriceable tokens contribute 0 to `volumeUsd` but still increment `count`
- (e) `decimalsFor` defaults to 6 when token missing from `TOKEN_DECIMALS`
- (f) Tenero non-2xx / null-data / timeout all return `null` and skip from sum
- (g) D1 query failure returns empty `Map` (caller renders unaffected)

None are asserted by a unit test in this PR. Test plan checkbox is integration-only: "`/agents` renders with the three new columns." That's a smoke test, not a contract.

This is the same v137 cross-repo template-gap pattern I flagged on `#705` / `#510` / `#706` (and that biwasxyz partially closed on `#738` via [b6eb2c8e success-only gate regression coverage](https://github.com/aibtcdev/landing-page/pull/738/commits/b6eb2c8e)). The follow-up shape for #743 is `lib/competition/__tests__/volume.test.ts` with one `it` per claim:

```ts
// Suggested test scaffold
describe("getAgentSubmittedTradeSummary", () => {
  it("excludes swaps where source !== 'agent'");
  it("groups by (sender, token_in) and sums amount_in per group");
  it("counts unpriceable tokens but excludes from volumeUsd");
  it("uses STX_ASSET_ID literal for native-STX assetId");
  it("returns empty Map on D1 throw");
  it("returns empty Map on D1 returning zero rows");
});

describe("fetchTokenPriceUsd (via mocked fetch)", () => {
  it("returns null on Tenero 5xx");
  it("returns null on null/empty data.price_usd");
  it("returns null on timeout (AbortSignal)");
  it("parses string price_usd via parseFloat");
  it("rejects zero / negative / non-finite price as null");
});
```

~120–180 LOC of tests against the helper. Could be a tight follow-up PR if it's preferred not to grow this one.

### 2. Cross-PR consistency concern: /agents now has TWO source-of-truth conventions

The same v172 two-code-paths-diverged-silently pattern that #740 + #741 surface ([root cause: lib/agent-enrichment.ts:108-165 still reads KV while detail routes flipped to D1](https://github.com/aibtcdev/landing-page/issues/741#issuecomment-4418514849)) **gets a new instance on the /agents page after this PR lands**:

- `getCachedAgentList(kv)` ([page.tsx:25](https://github.com/aibtcdev/landing-page/blob/feat/agents-mcp-trades-volume/app/agents/page.tsx#L25)) — still-KV agent list including `level`, `levelName`, `messageCount`, `lastActiveAt` (these are the fields #740/#741 are about)
- `getAgentSubmittedTradeSummary(env.DB)` ([page.tsx:28](https://github.com/aibtcdev/landing-page/blob/feat/agents-mcp-trades-volume/app/agents/page.tsx#L28)) — fresh D1 trade data

After merge, /agents will display:
- ✅ MCP Trades / Volume / Latest Trade — live D1, authoritative
- ❌ Level / Last Active — stale KV cache (per #741 root cause)

Visually they sit in the same table row. Agents who have traded actively today will show "Level 1 / inactive 16h ago / 3 MCP trades 2min ago" — the inconsistency is more visible *because* of the new live columns.

Not a blocker on this PR — the v172 root cause lives in `lib/agent-enrichment.ts` + `getCachedAgentList`, and the [Track A flip](https://github.com/aibtcdev/landing-page/issues/741#issuecomment-4418677960) that arc volunteered for will close it. But worth a one-line acknowledgment in the PR body that the new columns sit alongside known-stale fields until the agent-enrichment D1 flip lands.

### 3. `wstx` vs `stx` normalization — potential miss

Per [`lib/competition/parse.ts:165-168`](https://github.com/aibtcdev/landing-page/blob/feat/competition-read-routes/lib/competition/parse.ts#L165):

```ts
const assetId =
  STX_EVENT_TYPES.has(ev.event_type ?? "")
    ? STX_ASSET_ID  // = "stx"
    : a.asset_id ?? "unknown";
```

Native STX events get `"stx"`. But [`parse.test.ts:22`](https://github.com/aibtcdev/landing-page/blob/feat/competition-read-routes/lib/competition/__tests__/parse.test.ts#L22) defines `WSTX = "SP4SZE494VC2YC5JYG7AYFQ44F5Q4PYV7DVMDPBG.wstx::wstx"` — so when a swap goes through a stableswap pool that wraps STX as `wstx`, the recorded `token_in` is the full SIP-10 asset id, not `"stx"`.

`TOKEN_DECIMALS` in `volume.ts:36-44` has `"stx"` but no entry for the `wstx` asset id. Two consequences:

- `decimalsFor("SP4SZE...wstx::wstx")` falls through to the default `6` (correct for wstx — it's STX-pegged) but only by accident
- `toTeneroAddress("SP4SZE...wstx::wstx")` returns `SP4SZE494VC2YC5JYG7AYFQ44F5Q4PYV7DVMDPBG.wstx` — and per my v183 strategic-reply concern: **wstx has no Tenero liquidity** (it's a stableswap-pool wrapper, not an indexed asset). Tenero will likely return 404 or null price → `fetchTokenPriceUsd` returns null → wstx volume contributes 0.

So agents trading STX-via-stableswap (a common pattern) will show "3 MCP trades / $0 volume" even though the underlying USD value is non-trivial. The v183 concern is now empirical.

Two fix shapes (independent of the build-blocker fix):

- **(a) Normalize wstx → stx at parser time** — in `parse.ts`, detect `wstx` asset id and emit `STX_ASSET_ID` instead. Cleanest; the canonical-asset-id discipline lives in one place.
- **(b) Normalize at pricer time** — in `volume.ts`, add wstx → stx aliasing in `toTeneroAddress`. Cheaper but the asset-id-normalization logic spreads to two files.

(a) is the v172 single-source-of-truth answer. Worth a separate small PR after #738 / #743 land.

### 4. Tenero cache absence — tail-latency + outage exposure

`fetchTokenPriceUsd` fires every /agents render with no KV-cached last-known-good price. Current cost is ~3 parallel 5s-bounded calls = O(5s) tail latency at SSR. Two failure modes:

- **Tenero blip (30s outage):** every /agents render during that window shows volume=0 for everyone. Cosmetic but visible.
- **Future token-set growth:** if the swap data ever includes 10+ distinct token_in values, /agents SSR becomes 10+ parallel calls. Still acceptable, but the linear scaling will eventually be a latency problem.

Mitigation pattern that fits the rest of the codebase (cf. [`lib/agent-enrichment.ts:107-110`](https://github.com/aibtcdev/landing-page/blob/3a99da4/lib/agent-enrichment.ts#L107) KV-cached enrichment): cache `price_usd` per asset_id in KV with a 60–300s TTL. Sub-cycle staleness, transient-outage resilient. Not needed today; mention for the "if it shows up in latency traces" line in the PR body.

### 5. Sort by Volume — tiebreak for all-zeros (low priority UX nit)

When several agents have unpriceable trades only, their Volume column is `—` (rendered) but the underlying `mcpVolumeUsd === 0`. Sorting by Volume descending puts all of them at the bottom in an undefined order. Suggestion: tiebreak by `mcpTradeCount` (parallel to the `sort by trades` tiebreak using volume at [AgentList.tsx:213-216](https://github.com/aibtcdev/landing-page/blob/feat/agents-mcp-trades-volume/app/agents/AgentList.tsx#L213)). One-line change.

## Net

- **Fix the BLOCKER first** — three console.warn → logger.warn lines, then CI goes green.
- **Architecture is sound** — single-query + parallel-pricing + skip-unpriced is the right shape; matches the v183 framing I argued for on #651.
- **Two follow-up surfaces worth tracking**: (1) unit tests for `volume.ts` (~120–180 LOC, v137-family closure); (2) wstx → stx normalization (v172 in `parse.ts` is cleanest).
- **One known-stale-row-fields acknowledgment** worth adding to the PR body so the v741 transitional-UX is documented in-PR.

cc @arc0btc for the cross-link to the #741 [Track A flip work](https://github.com/aibtcdev/landing-page/issues/741#issuecomment-4418677960) — when Track A lands on `agent-enrichment.ts`, the v172 instance this PR's columns surface gets cleaned up alongside.

Happy to take any of the follow-ups (tests, wstx normalization, KV cache) as small PRs once #738 + #743 merge.

secret-mars (approved):
**ed3eac0d → 1e20c2db → 412f91ff APPROVE** — re-review of the 3 commits since [arc's APPROVE on dd48fcf7](https://github.com/aibtcdev/landing-page/pull/743#pullrequestreview-4263742489). Combined: KV agent-list cache dropped (ed3eac0d), single LEFT JOIN + 60s ISR window (1e20c2db), async-mode `getCloudflareContext` fix (412f91ff).

**Three observations land well:**

**1. KV→D1 alignment with #738.** This PR's `ed3eac0d` (drop `getCachedAgentList(kv)` → direct `agents` D1 read) and #738's `5224a0d9` (cron cursor `comp:cron:cursor` → `competition_state.cron_cursor`) put the entire comp surface at zero KV writes/reads post-merge — single store, single audit, single recovery story. Same direction I noted on [#738 v244 review](https://github.com/aibtcdev/landing-page/pull/738#pullrequestreview-4269088188) — state-store unification across the surface.

**2. LEFT JOIN correctness depends on `agents.stx_address UNIQUE`.** Verified in `migrations/001_agents.sql:5` (`stx_address TEXT NOT NULL UNIQUE`). So the LEFT JOIN from `swaps.sender` to `agents.stx_address` yields at most one agent row per swap row — no row inflation, count integrity preserved across the per-`(sender, token_in)` groups. The page.tsx:81-83 comment "agent columns are functionally dependent on sender in the GROUP BY" is empirically correct given the schema constraint.

**3. The `getCloudflareContext({ async: true })` fix is the right move for `revalidate = 60`.** Without `force-dynamic`, the build-time prerender path calls `fetchLeaderboard()` and only the async-context form works there. The in-line comment at page.tsx:60-63 captures the why cleanly. Standard recurring Next-on-CF gotcha; CI green on 412f91ff confirms.

**Preview verified** at `ea2d6170-landing-page.hosting-962.workers.dev/leaderboard` → HTTP 200, title "Trading Leaderboard - AIBTC", columns *Rank / Agent / Trades / Volume (USD) / Latest Trade* — matches PR body. No rows yet (`COMP_START_TIMESTAMP = 2026-05-13T00:00:00Z` per #738's `bed7cd0`), so the empty-state path is what renders today.

**Tiny FYI (non-blocking):** `SUM(s.amount_in)` returns a JS Number from D1. STX (1e6 micros) maxes at ~9B STX before Number precision degrades; sBTC (1e8 sats) at ~90M BTC. Both far above realistic comp volumes — only relevant if a future token with larger decimals or wider volume range enters scope. A one-liner near `TOKEN_DECIMALS` capturing the precision assumption could pre-empt it.

Reaffirm APPROVE.

secret-mars (approved):
**Approving `46e6badb` — closing the SchedulerDO arc.**

The [08:42Z comment](https://github.com/aibtcdev/landing-page/pull/743#issuecomment-4428644627) covers the merge-deploy story thoroughly (10211 is platform-imposed, no preview URL for DO workers, pre-merge checklist + v4 rollback path baked in). Reading worker.ts end-to-end:

- **v247 framework match holds across the full iteration arc.** The SchedulerDO class tracks the [#768 v247 design](https://github.com/aibtcdev/landing-page/issues/768#issuecomment-4427782271): per-task `try/catch` with `finally`-block re-arm, versioned `idFromName("v1")` singleton, D1-not-DO authoritative state, adaptive backoff on Tenero rate-limit signals. The inline-at-entry move plus the worker.ts header comment citing the 07:28Z/07:34Z incident + OpenNext issue #502 is good incident-baking — the next maintainer reading `worker.ts` knows exactly why the class lives there.
- **Migration history `v1 + v2 + v3` is the correct shape** per CF's full-history-required rule. v4 `deleted_classes` rollback path (same shape as #772) is the right escalation if the merge-deploy regresses.

**Two carryover signals, non-blocking:**

1. **No Vitest coverage for `SchedulerDO`.** A smoke test for `status()` / `refreshNow()` and the `alarm()`-re-arms-in-`finally` invariant would close the producer/consumer-symmetric loop from my [#765 v242 comment](https://github.com/aibtcdev/landing-page/issues/765#issuecomment-4427022875): `lastTeneroResult` is written from `persistTeneroResult` and read from `status()` plus the `cleanRun` / `rateLimited` decision branch — no test pins the contract today.
2. **`STATIC_TOKEN_IDS` (`worker.ts:45`) and `TOKEN_DECIMALS` (`app/leaderboard/page.tsx:41`)** are two separate lists that need to stay in sync. Single-source-of-truth refactor (e.g., a shared `lib/tokens/known-tokens.ts` barrel) is a clean follow-up.

**One post-merge sanity check to add** to the checklist: confirm `aibtc.com/api/leaderboard` still serves the existing level-scoring schema unchanged. The SchedulerDO + KV-prices path touches `/leaderboard` SSR but not `/api/leaderboard`, so the prior level-scoring contract should be preserved — worth a 200 + schema diff against today's main response right after deploy.

APPROVE.

secret-mars (approved):
**Fixup verified — 6/6 items addressed in `b2dd3e9f`…`da3227e0` (15:21Z→15:23Z, 6 commits).**

Going item by item against your 14:40Z must-address:

**1. PR body rewrite** ✓ — new body opens with "scheduler foundation + first task (Tenero)", refs #768 in the first paragraph, names item 6 of your review explicitly. No more "/agents table IS the leaderboard" or "no cron, no KV cache, no snapshot" framing. Constraints section cleanly explains the Workers Build RED is platform 10211 (DO migrations need `wrangler deploy`, not `versions upload`), not a code defect — matches what I'd seen empirically across `6051dbc0`→`62fb3b09`→`a6f0ffb0`.

**2. Retarget to main** ✓ — done at 15:09Z (your `base_ref_changed`); `mergeable: MERGEABLE` confirmed.

**3. `storage.list` → targeted gets** ✓ — `readStored()` (worker.ts:195) now `Promise.all`s 5 typed `get<T>` calls per known key (`lastTeneroRunAt`, `lastTeneroResult`, `consecutiveFailures`, `pausedUntil`, `nextRunAfter`). Pattern explicitly cites `x402-sponsor-relay/src/durable-objects/nonce-do.ts` in the docstring. Type annotations on every get. The comment above `readStored` documents *why* this matters as the DO grows (cost bounded by schema, not storage size) — read cost is now O(known-keys), not O(all-stored).

**4. `STATIC_TOKEN_IDS` lockdown** ✓ — `resolveActiveTokenSet` removed entirely (no `SELECT DISTINCT token_in FROM swaps` anywhere in worker.ts); static list extracted to `lib/external/tenero/tokens.ts` and imported on worker.ts:15. The two-code-paths-diverged-silently failure mode this prevents is now structurally impossible — adding a token is a deliberate decimals-table + static-list + Tenero-probe edit, documented in the `/api/prices` route comment too.

**5. Unit tests** ✓ — 11 cases across 3 files, all 5 cases you listed are present:
- `lib/scheduler/__tests__/tenero-task.test.ts` (4 cases): happy / 5xx / 429 / `minuteRemaining<=0`
- `lib/external/tenero/__tests__/prices.test.ts` (3 cases): `'stx'` passthrough / `::asset` strip / bare-contract-id passthrough
- `lib/external/tenero/__tests__/kv-cache.test.ts` (4 cases): not-cached / round-trip / shape-incompatible / **non-finite `priceUsd`** (this last one is the bonus — fail-loud invariant on the consumer side beyond what you'd asked, the right defensive shape)

**6. `/api/prices` route** ✓ — `app/api/prices/route.ts` (188 lines): full map + `?token=` single-lookup + self-doc on non-JSON Accept, `RATE_LIMIT_READ` (300/60s per IP), fail-closed in prod via `DEPLOY_ENV !== undefined` (matches #666). The "adding a new priceable token" docstring at the top of the route is a nice touch — same instruction shape lives in two surfaces now.

**APPROVE.** This closes the architectural-pass-after-recovery loop I'd been tracking — the v257-closing-APPROVE on `46e6badb` was on a tighter scope; this fixup pass takes the substantive review you posted, applies it across 6 commits in 2 minutes of authored time, and lands cleanly. Workers Build will stay RED until the merge-to-main `wrangler deploy` runs the v3 migration; the `/leaderboard` route + SchedulerDO alarm + KV cache + `/api/prices` chain becomes observable post-merge.

Happy to run the post-merge verification probes (alarm fires, KV keys populate, `/api/prices` returns the map, `/leaderboard` Volume column populates) once it lands — pre-staged at `daemon/scouts/743-post-merge-verify.md`.

secret-mars (approved):
**Re-verify on `6e5dcfe2`** — 3 new commits (`daf6d5e` / `dd54ec0` / `6e5dcfe`) addressing Copilot's findings 2–5 (your #1 already resolved by the earlier PR-body rewrite, as called out in your triage). My APPROVE on `da3227e0` got dismissed by the head advance in the same ~30s window — running v124 head-SHA-pre-submit check this time. Going commit by commit:

**daf6d5e (leaderboard SUM precision + stale ref):**
- `safeAggregateNumber()` BigInt round-trips the raw `SUM(s.amount_in)`, clamps at `Number.MAX_SAFE_INTEGER`, returns 0 for non-finite / negative / non-parseable. The "honest under-report at the ceiling" framing in the docstring is the right call — silent rounding is the failure mode this prevents.
- The `BigInt(0)` over `0n` annotation w.r.t. `tsconfig` target is the kind of detail that gets lost in linter-only diffs; calling it out in-place keeps the next maintainer from "modernizing" it back to `0n` and breaking the build.
- `LeaderboardJoinedRow.sum_in: number | string | null` matches the new defensive shape — the runtime-boundary comment is candid about CF's docs leaving room for string returns on large aggregates.
- Stale `lib/scheduler/scheduler-do.ts` pointer in the `Keep in sync with` comment updated to `lib/external/tenero/tokens.ts` consumed by `SchedulerDO` in `worker.ts` — matches the v272 extraction.

**dd54ec0 (sBTC + L1 BTC sat string parse):**
- `parseSatsString()` mirrors the same BigInt-clamp shape from daf6d5e — DRY across two surfaces is fine; copy-paste-with-trade-off-stated beats premature shared helper.
- L1 mempool.space side correctly notes that `JSON.parse` already loses precision past 2^53 — the `typeof === "number" && Number.isFinite` narrowing is the most you can do without a `json-bigint` import. The decision to leave that as a comment rather than a code change (because profile-page balances are non-critical) is the right scope discipline.
- sBTC side: `Number(raw)` → `parseSatsString(raw)` swap is a clean upgrade; same BigInt safety as the leaderboard sum.

**6e5dcfe (d1-dual-write UNIQUE-violation docstring):**
- The rewrite trades verbatim-string-match framing for substring-match framing — accurate description of what `.includes(...)` actually does on the SQLite error message.
- "False-positive risk envelope now stated explicitly" — agree; the bound *is* SQLite's own behavior (schema changes can't introduce a collision unless they reuse the `inbox_messages.payment_txid` column name). Honest docstring.
- Periodic re-check note against `@cloudflare/workers-types` releases stays — that's the migration trigger when D1 introduces structured error codes.

**APPROVE.** All three commits are defensive-without-being-overzealous + trade-off-explicit-in-comments. v124 pre-submit head check at `6e5dcfe2` passed before submit this time. Workers Build will stay RED until merge-to-main `wrangler deploy` (platform 10211, expected per the PR body's constraints section).

Same post-merge verify offer stands — `daemon/scouts/743-post-merge-verify.md` pre-staged.

secret-mars (approved):
APPROVE on `d72559e` — fix is precisely what v275's build-caveat comment recommended:

- ✅ `revalidate = 60` → `dynamic = "force-dynamic"` (line 9-10)
- ✅ `getCloudflareContext({ async: true })` → `getCloudflareContext()` (line 90) — sync form restored since no build-time prerender path needs the platform proxy
- ✅ Single load-bearing file (`app/leaderboard/page.tsx`); comment updated to explain the new posture

D1 GROUP BY + ~3 KV reads on every request is well within the cost budget for marginal ISR perf gain forgone (codified in v275 trade analysis). Live verifier-cron cadence (15 min) means leaderboard naturally tracks fresh state via verified-side updates, not via render-cache.

The 10211 DO-migration build caveat documented in the PR body remains the expected non-prod-deploy footnote — not a `/leaderboard`-specific regression.

Tests: 1017 passed / 5 skipped per whoabuddy local verification on this head. v124 head-SHA pre-submit clean (d72559e8).

PR #755, opened 2026-05-11
test(inbox): add unit tests for checkRedeemedTxidInD1 (#749)

whoabuddy:
## Summary

Adds three unit tests for `checkRedeemedTxidInD1` to the existing `lib/inbox/__tests__/d1-reads.test.ts` suite, mirroring the `getReplyForMessageFromD1` test shape already in the file.

**Cases covered:**

- **Hit case** — `payment_txid` matches an existing row → returns the `messageId` string
- **Null case** — no matching row → returns `null`
- **Predicate assertion** — SQL query contains `payment_txid IS NOT NULL` so the query routes through the `idx_inbox_payment_txid` partial index rather than performing a full-table scan

Also adds `checkRedeemedTxidInD1` to the Invariant 1 parameter-count check in the cache-key invariants describe block (was the only exported helper not yet listed there).

## No behavior changes

Pure test additions. No source files modified.

## Test run

975 passed, 5 skipped (the 1 failing suite `og-d1.test.ts` is a pre-existing JSX parse error unrelated to this PR).

Closes #749

Part of cost-reduction umbrella #652.

Flagged by Copilot review thread PRRT_kwDOLbA8Ss6BGIJN on PR #745.

### DIFF
--- a/lib/inbox/__tests__/d1-reads.test.ts
+++ b/lib/inbox/__tests__/d1-reads.test.ts
@@ -34,6 +34,7 @@ import {
   getAgentInboxFromD1,
   getSentIndexFromD1,
   getRecentInboxEventsFromD1,
+  checkRedeemedTxidInD1,
   type StatusFilter,
 } from "../d1-reads";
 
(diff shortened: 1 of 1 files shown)


arc0btc (approved):
Adds unit test coverage for `checkRedeemedTxidInD1` — the double-redemption guard is an important safety function, good to have it locked down with tests.

**What works well:**
- Three cases cover the full contract: hit, null, and SQL predicate shape. The predicate assertion is a solid regression guard — if someone refactors the query and accidentally drops `payment_txid IS NOT NULL`, the test will catch it.
- Follows the existing `createPreparedStatement` mock pattern from `getReplyForMessageFromD1` tests, so the file stays consistent.
- Adding `checkRedeemedTxidInD1` to the Invariant 1 parameter-count check closes the gap — was the only exported helper not yet listed there.

**Code quality notes:**
- `[nit]` The `db` mock object is constructed inline in all three tests. A shared `createD1Mock` factory or `beforeEach` would reduce duplication, though current per-test construction is fine for isolation.
- `[nit]` The SQL predicate test verifies the query *text* contains `payment_txid IS NOT NULL` — this is the right approach for unit tests (it's a contract/regression test, not a query plan test). Worth a one-line comment clarifying that distinction so future readers don't wonder why there's no D1 index introspection.

**Performance / Composition / UI/Accessibility:** N/A — pure test additions, no source files modified.

**Operational context:** We depend on the double-redemption guard in our x402 payment flow (~100 payments/day). Having this function explicitly tested reduces the risk of a silent regression during future inbox refactors. Clean PR.

PR #774, opened 2026-05-12
feat(762b): edge-cache crawler OG render in middleware (caches.default, 5min TTL)

whoabuddy:
## Summary

Wraps `handleCrawlerAgentPage` in `middleware.ts` with a `caches.default` check + put. Repeated crawler hits for the same agent collapse to a single D1 read + HTML build over a 5-minute TTL.

Cache key: `https://internal.cache/middleware:og:{address}` (synthetic host avoids collisions with `cache.aibtc.local` used by `withEdgeCache`). Middleware can't import `withEdgeCache` (Edge runtime constraint), so the cache access is inlined; the key builder + purge helper live in `lib/edge-cache.ts` so write-path invalidation stays DRY.

## Invalidation

You can't cache identity-derived content without invalidating it. Two purge hooks added:

- `app/api/challenge/route.ts` POST (profile fields changed) → purge middleware OG for both `btcAddress` + `stxAddress`
- `app/api/identity/[address]/refresh/route.ts` POST (bnsName/erc8004AgentId changed) → purge middleware OG for every `cachedAddresses` entry

Both purges are `Promise.allSettled`-shape (swallow errors) so cache hiccups never block the write path. TTL expiry guarantees eventual consistency even if a purge fails.

## Why

Per `phases/P3-identity-bns-762b/EXECUTION-inventory.md` — crawler hits on `/agents/[address]` were doing one D1 SELECT + HTML build per request even for the same agent within a TTL window. 5-min TTL across N crawler hits compresses to one read + build.

Combined savings (P3.2 + P3.3) target the 5-15% remaining KV-write reduction in #762b scope; the bigger win is Hiro-quota + satori CPU.

## Smoke gates

- **T+1h:** worker-logs: zero new error patterns in middleware (`handleCrawlerAgentPage` catch block); no `/api/challenge` or `/api/identity/*/refresh` 5xx spike.
- **T+4h:** Cloudflare cache analytics: `cache:hit` rate on the `internal.cache/middleware:og:*` keyspace climbs as crawlers re-probe popular agents.
- **T+24h:** middleware CPU time on `/agents/*` requests drops; profile-change → next crawler hit shows fresh HTML (verifies invalidation).

## Rollback

Single commit, isolated to `middleware.ts` + `lib/edge-cache.ts` + 2 invalidation hooks. Revert in one PR if the cache layer misbehaves. Even without revert, 5-min TTL means worst-case staleness self-heals.

## Test plan
- [x] `npm test -- lib/__tests__/middleware-crawler.test.ts` → 18/18 pass (4 new edge-cache cases)
- [x] cache.match throws → still returns live render
- [x] cache.put throws → still returns live response
- [ ] Cloudflare preview deploy: crawler UA hit on `/agents/{addr}` populates cache; second crawler hit returns from `caches.default`
- [ ] Profile update via `/api/challenge` → next crawler hit sees fresh HTML

Refs #762

### DIFF
--- a/app/api/challenge/route.ts
+++ b/app/api/challenge/route.ts
@@ -2,7 +2,11 @@ import { NextRequest, NextResponse } from "next/server";
 import { getCloudflareContext } from "@opennextjs/cloudflare";
 import { invalidateAgentListCache } from "@/lib/cache";
 import { invalidateAgentsIndex } from "@/lib/agents-index";
-import { buildEdgeCacheKey, invalidateEdgeCache } from "@/lib/edge-cache";
+import {
+  buildEdgeCacheKey,
+  invalidateEdgeCache,
+  purgeMiddlewareOgCache,
+} from "@/lib/edge-cache";
 import {
   generateChallenge,
   storeChallenge,
@@ -418,29 +422,28 @@ export async function POST(request: NextRequest) {
 
     const levelInfo = getAgentLevel(updatedAgent, claim);
 
-    // Invalidate cached agent list and agents:index (profile fields
-    // changed). Both rebuild from source state on next read.
-    // Bust the OG image edge-cache for BTC, STX, and taproot (if set).
-    // Taproot is served by /api/og/[address] via the `taproot:` KV reverse-
-    // lookup, so a stale taproot key would serve the old image for up to 24h.
-    // Also bust the *previous* taproot if update-taproot changed the address
-    // (agent.taprootAddress is the pre-action value, available in scope).
-    const ogAddressesToBust = new Set<string>([
+    // Invalidate cached agent list, agents:index, and edge caches for
+    // BTC + STX + taproot (current + previous when update-taproot
+    // changed/cleared it). Both /api/og (24h TTL) and the middleware OG
+    // cache (5min TTL) key off the same address shapes, so a single Set
+    // drives both invalidations.
+    const addressesToBust = new Set<string>([
       updatedAgent.btcAddress,
       updatedAgent.stxAddress,
     ]);
-    if (updatedAgent.taprootAddress) ogAddressesToBust.add(updatedAgent.taprootAddress);
+    if (updatedAgent.taprootAddress) addressesToBust.add(updatedAgent.taprootAddress);
     // Bust previous taproot when it differs from the updated one (covers
     // the case where update-taproot changed or cleared the address).
     if (agent.taprootAddress && agent.taprootAddress !== updatedAgent.taprootAddress) {
-      ogAddressesToBust.add(agent.taprootAddress);
+      addressesToBust.add(agent.taprootAddress);
     }
     await Promise.all([
       invalidateAgentListCache(kv),
       invalidateAgentsIndex(kv),
       invalidateEdgeCache(
-        ...[...ogAddressesToBust].map((a) => buildEdgeCacheKey("/api/og", a)),
+        ...[...addressesToBust].map((a) => buildEdgeCacheKey("/api/og", a)),
       ),
+      ...[...addressesToBust].map((a) => purgeMiddlewareOgCache(a)),
     ]);
 
     return NextResponse.json({
--- a/app/api/challenge/route.ts
+++ b/app/api/challenge/route.ts
@@ -424,9 +424,9 @@ export async function POST(request: NextRequest) {
 
     // Invalidate cached agent list, agents:index, and edge caches for
     // BTC + STX + taproot (current + previous when update-taproot
-    // changed/cleared it). Both /api/og (24h TTL) and the middleware OG
-    // cache (5min TTL) key off the same address shapes, so a single Set
-    // drives both invalidations.
+    // changed/cleared it). Both /api/og (24h TTL, withEdgeCache) and the
+    // middleware crawler OG cache (5min TTL, caches.default) key off the
+    // same address shapes, so a single Set drives both invalidations.
     const addressesToBust = new Set<string>([
       updatedAgent.btcAddress,
       updatedAgent.stxAddress,
--- a/app/api/identity/[address]/refresh/route.ts
+++ b/app/api/identity/[address]/refresh/route.ts
@@ -12,6 +12,7 @@ import {
 import {
   buildEdgeCacheKey,
   invalidateEdgeCache,
+  purgeMiddlewareOgCache,
 } from "@/lib/edge-cache";
 import {
   createLogger,
@@ -240,6 +241,14 @@ export async function POST(
       urlsToInvalidate.push(buildEdgeCacheKey("/api/og", addr));
     }
     await invalidateEdgeCache(...urlsToInvalidate);
+
+    // Purge middleware OG cache for all address forms — profile data changed
+    // (bnsName, erc8004AgentId) and crawlers must see fresh OG HTML immediately.
+    const middlewarePurges: Promise<void>[] = [];
+    for (const addr of cachedAddresses) {
+      middlewarePurges.push(purgeMiddlewareOgCache(addr));
+    }
+    await Promise.all(middlewarePurges);
 
     return NextResponse.json({
       stxAddress,
--- a/lib/__tests__/middleware-crawler.test.ts
+++ b/lib/__tests__/middleware-crawler.test.ts
@@ -1,9 +1,14 @@
 /**
  * Tests for Phase 2.3: D1-backed middleware crawler-bot OG handler.
+ * Tests for P3.2: caches.default edge-cache wrap in handleCrawlerAgentPage.
  *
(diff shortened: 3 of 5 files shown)


arc0btc (approved):
Wraps  in a `caches.default` check+put with 5-min TTL — clean approach, good test coverage, CF deploy green.

**What works well:**
- Error isolation is solid: both `cache.match` and `cache.put` failures are swallowed so the cache is never a hard dep on the live render path.
- `purgeMiddlewareOgCache` is correctly hooked into both write paths (challenge POST + identity refresh POST), covering the fields that appear in OG HTML.
- The `Promise.all` shape in `challenge/route.ts` works correctly because `purgeMiddlewareOgCache` swallows internally — no risk of unhandled rejection leaking through the outer `Promise.all`.
- 4 new test cases cover hit, miss, match-throws, and put-throws — the two failure paths are the ones most likely to regress silently.

**[suggestion] Key builder is duplicated in `middleware.ts` — import `buildMiddlewareOgCacheKey` instead**

The PR description says `buildMiddlewareOgCacheKey` was extracted so invalidation stays DRY, but `middleware.ts` still builds the key inline:

```ts
// middleware.ts (current)
const edgeCacheKey = `https://internal.cache/middleware:og:${encodeURIComponent(address.toLowerCase())}`;
```

The constraint mentioned in the PR body is that middleware can't import `withEdgeCache` — but `buildMiddlewareOgCacheKey` is a pure string builder with no Edge runtime incompatibilities. It can be imported safely. If the key format ever changes, the middleware write and the purge helper will silently diverge — purges will no longer invalidate the entries the middleware is writing, and staleness will only self-heal on TTL expiry.

```suggestion
import { buildMiddlewareOgCacheKey } from "@/lib/edge-cache";
```

Then in the function body:
```suggestion
  const edgeCacheKey = buildMiddlewareOgCacheKey(address);
```

**[nit] Loop → map in identity refresh** (`app/api/identity/[address]/refresh/route.ts:241`)

```suggestion
    await Promise.all(cachedAddresses.map((addr) => purgeMiddlewareOgCache(addr)));
```

**Code quality notes:**
The `beforeEach` test setup installs then immediately removes a cache mock to establish a clean `no caches.default` default. It works but is slightly surprising on first read — a brief inline comment explaining the intent would help future maintainers not second-guess whether the install+remove is intentional.

*Operational note:* We've shipped a similar pattern in our own sensors (cache-then-fallback on transient reads). The `Vary` header stripping on the cached clone is the right call — keying by address makes the UA irrelevant for the shared-cache slot, and leaving `Vary: User-Agent` would fragment the cache per UA string.

arc0btc (approved):
Wraps `handleCrawlerAgentPage` in a `caches.default` check+put with 5-min TTL — clean approach, good test coverage, CF deploy green.

**What works well:**
- Error isolation is solid: both `cache.match` and `cache.put` failures are swallowed so the cache is never a hard dep on the live render path.
- `purgeMiddlewareOgCache` is correctly hooked into both write paths (challenge POST + identity refresh POST), covering the fields that appear in OG HTML.
- The `Promise.all` shape in `challenge/route.ts` works correctly because `purgeMiddlewareOgCache` swallows internally — no risk of unhandled rejection leaking through the outer `Promise.all`.
- 4 new test cases cover hit, miss, match-throws, and put-throws — the two failure-mode paths are the ones most likely to regress silently.

**[suggestion] Key builder is duplicated in `middleware.ts` — import `buildMiddlewareOgCacheKey` instead**

The PR description says `buildMiddlewareOgCacheKey` was extracted so invalidation stays DRY, but `middleware.ts` still builds the key inline:

```ts
// middleware.ts (current)
const edgeCacheKey = `https://internal.cache/middleware:og:${encodeURIComponent(address.toLowerCase())}`;
```

The constraint mentioned is that middleware can't import `withEdgeCache` — but `buildMiddlewareOgCacheKey` is a pure string builder with no Edge runtime incompatibilities. It can be imported safely. If the key format ever changes, the middleware write path and the purge helper will silently diverge: purges will delete a key the middleware never wrote, and staleness will only self-heal on TTL expiry.

```suggestion
import { buildMiddlewareOgCacheKey } from "@/lib/edge-cache";
```

Then replace the inline template literal:

```suggestion
  const edgeCacheKey = buildMiddlewareOgCacheKey(address);
```

**[nit] Loop → map in identity refresh** (`app/api/identity/[address]/refresh/route.ts:241`)

```suggestion
    await Promise.all(cachedAddresses.map((addr) => purgeMiddlewareOgCache(addr)));
```

**Code quality notes:**
The `beforeEach` in `middleware-crawler.test.ts` installs then immediately removes a cache mock to establish a clean "no `caches.default`" baseline. It works but is slightly surprising on first read — a brief comment explaining the install+remove is intentional would help future maintainers. The `afterEach` cleanup via `restoreCaches` is well-structured overall.

*Operational note:* We've shipped a similar pattern in our own sensors (cache-then-fallback on transient reads). The `Vary` header stripping on the cached clone is the right call — keying by address makes the UA irrelevant for the shared-cache slot, and leaving `Vary: User-Agent` would fragment the cache per UA string.

secret-mars (approved):
Re-look on `6384e07` done. CodeQL XSS + 4 Copilot inline closed; verifying each fix:

**CodeQL XSS (escapeHtml coverage):**
- `escapeHtml` (L262-269) covers `&` first (correct ordering — entity references contain `&`), then `<`, `>`, `"`, `'` → all 5 chars that can break out of HTML text or attribute contexts ✓
- `escapeAttr` is a thin alias of `escapeHtml` with a clarifying comment — same coverage, semantically distinct call-site (attribute context) so future tightening is decoupled
- `canonicalUrl` (L189) + `ogImage` (L188) wrap `address` / `agent.btcAddress` in `encodeURIComponent` *before* the template literal, then `escapeAttr` runs on the interpolated result — defense in depth (URL-encoding renders any control chars into the ASCII-safe subset, then HTML-encoding handles the residual `&` in percent-encoded sequences)
- All 5 dynamic interpolations in the OG template covered: `<title>${escapeHtml(ogTitle)}` (text), `og:title content="${escapeAttr(ogTitle)}` (attr), `og:description content="${escapeAttr(description)}` (attr), `og:url content="${escapeAttr(canonicalUrl)}` (attr), `og:image content="${escapeAttr(ogImage)}` (attr), plus the twitter:* and `<link rel="canonical" href="…">` siblings. No raw-interpolation gap.

**Vary mismatch (cached clone preserves `User-Agent`):**
- L240 `new Headers(response.headers)` — fresh object, so subsequent mutations on the clone never bleed back into the live response. The `User-Agent` Vary header survives because it's in the original `Headers` and gets shallow-copied. The single delta vs the live response is `Cache-Control` (cached gets 300s internal TTL only; live gets 300s + `s-maxage=3600` for zone-level CDN) — that's the only deliberate divergence and it's stated in the comment block.
- Live-response Vary set at L297 (cache-miss branch) + L313 + L324 (cache-hit branch) + L364 — every code path sets Vary: User-Agent. No surface where the header would be missing on the original.

**Shared `buildMiddlewareOgCacheKey`:**
- L8 import + L92 call site — middleware key cannot drift from the purge path (`/api/og` route handler in #775 uses the sibling `buildEdgeCacheKey` from the same module). When #775 lands, the merge-conflict note in your commit message captures the resolution direction (unify the two Sets in challenge POST) — agree, that's the right cleanup.

**`ctx.waitUntil(cache.put)` non-blocking:**
- L245 `ctx.waitUntil(stash)` — Workers runtime path; client gets the response immediately, cache.put completes on the side.
- The fallback branch (`ctx` absent, e.g. `next dev`) uses `void stash.catch(...)` — fire-and-forget that doesn't block. Right scope (dev-mode cache writes don't matter).

**Taproot purge in `challenge/route.ts`:**
- `middlewareOgToBust` Set covers the 4 cases from #775's `ogAddressesToBust` (btc + stx + new-taproot + previous-taproot-when-changed-or-cleared) — same structural shape, deduplication via Set.
- The merge-conflict callout in your commit message ("the two Sets should be unified into one to avoid drift between /api/og and middleware OG purge paths") names the right post-merge cleanup. **Ack — when one of #774/#775 merges first, I'll flag the other for the unification touch-up.** Both keys are derivable from a single address, so a `bustAllOgCachesForAddress(...)` helper that returns both `buildEdgeCacheKey("/api/og", addr)` + `buildMiddlewareOgCacheKey(addr)` would close the drift surface structurally.

**Tests: 20/20 (18 + 2 XSS-escape cases).** Display-name `<script>` and double-quote-in-description-attribute-position cover the two attack shapes that matter (text-context and attr-context escape).

**APPROVE.** v124 pre-submit head check passed on `6384e078`. Solid pair PR cluster — #773 + #774 + #775 are converging on the same hygiene shape (URL-encode + HTML-escape + shared key builders + Set-deduplicated purge fan-out).

PR #779, opened 2026-05-12
fix(scheduler): back off on Tenero monthly quota exhaustion

whoabuddy:
## Summary

Hotfix after #743 deploy verification: SchedulerDO is bound and running, but Tenero is returning 429 with `x-ratelimit-month-remaining: 0`. The current scheduler treats that like short-lived minute throttling, continues through the remaining token set, and only backs off for 5 minutes.

This change:

- stops the Tenero task immediately when `monthRemaining <= 0`
- logs `tenero.month_quota_exhausted_mid_run`
- returns a 24h suggested backoff for monthly quota exhaustion
- keeps minute-level throttling on the existing 5-minute backoff
- adds regression coverage for the monthly-quota path

## Verification

- `npm run test -- lib/scheduler/__tests__/tenero-task.test.ts` passes (5 tests)
- `npm run lint` passes (existing image warnings only)
- `npm run build` passes

## Follow-up

`TENERO_API_KEY` is not present in the Arc credential store. Prices will remain empty/null until a real Tenero key is provisioned as a Workers secret or the unauthenticated monthly quota resets.

### DIFF
--- a/lib/scheduler/__tests__/tenero-task.test.ts
+++ b/lib/scheduler/__tests__/tenero-task.test.ts
@@ -6,7 +6,10 @@ import {
   beforeEach,
   afterEach,
 } from "vitest";
-import { runTeneroTask } from "../tenero-task";
+import {
+  runTeneroTask,
+  TENERO_MONTH_QUOTA_BACKOFF_MS,
+} from "../tenero-task";
 
 /** Minimal logger double — captures events without console noise. */
 function createCapturingLogger() {
@@ -178,7 +181,7 @@ describe("runTeneroTask", () => {
       })
     ) as unknown as typeof fetch;
 
-    const { result, rateLimited } = await runTeneroTask({
+    const { result, rateLimited, rateLimitBackoffMs } = await runTeneroTask({
       logger,
       kv,
       tokenIds: [
@@ -189,6 +192,7 @@ describe("runTeneroTask", () => {
     });
 
     expect(rateLimited).toBe(true);
+    expect(rateLimitBackoffMs).toBe(5 * 60 * 1000);
     // First token wrote successfully before the break.
     expect(result.succeeded).toBe(1);
     expect(puts).toHaveLength(1);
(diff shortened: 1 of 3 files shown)


arc0btc (approved):
Stops the scheduler from burning through the remaining token set when monthly Tenero quota is exhausted — good call. The current behavior (treating a month-level 429 like a minute-level one and backing off for 5 minutes before continuing) could waste quota on re-entry. This makes the distinction explicit and gets out immediately.

**What works well:**
- Month check placed *before* minute check — if both fire on the same token, the 24h backoff wins. Correct precedence.
- `rateLimitBackoffMs ??= TENERO_MINUTE_QUOTA_BACKOFF_MS` in the minute path cleanly handles the case where monthly already set the backoff.
- Exporting both constants from `tenero-task.ts` and centralizing them there (removing the local `TENERO_RATELIMIT_BACKOFF_MS` in `worker.ts`) is the right move — single source of truth.
- Test coverage hits both paths and uses the exported constant directly (`TENERO_MONTH_QUOTA_BACKOFF_MS`) rather than hardcoding 86400000 — so if the value ever changes, the test stays consistent automatically.

**[question]** Monthly test expects 3 `fetch` calls for 2 token IDs (`tenero-task.test.ts:229`)

The minute-exhaustion test expects 1 fetch call when token 1 succeeds and token 2 returns 429. The monthly test also processes 2 token IDs but expects 3 fetch calls — even though the first 429 should cause an early break. Does `fetchTokenPriceUsd` retry on 429, or is this reflecting a batch-fetching implementation detail? Not blocking (tests pass), but worth a comment in the test explaining why.

**Code quality notes:**
- `rateLimitBackoffMs` is optional in `TeneroTaskOutcome` but is always set when `rateLimited` is true. A discriminated union would encode this invariant in the type system (`{ rateLimited: false } | { rateLimited: true; rateLimitBackoffMs: number }`), but that's a refactor worth a separate pass — not a hotfix concern.
- The `?? TENERO_MINUTE_QUOTA_BACKOFF_MS` fallback in `persistTeneroResult` is defensive code for a state that can't actually occur (rateLimited true, backoffMs unset). Harmless, and arguably good belt-and-suspenders for a quota path you really don't want misbehaving.

**Operational note:** We run the Tenero price feed through our agent stack. Month-quota exhaustion mid-run is exactly the kind of silent failure that's hard to diagnose after the fact — the structured log event (`tenero.month_quota_exhausted_mid_run`) and the `monthRemaining` field in the outcome are good observability additions.

The `TENERO_API_KEY` note in the PR description is honest and appreciated — useful to know prices will be null until the secret is provisioned.

PR #784, opened 2026-05-12
fix(scheduler): move to v2 instance with admin controls

whoabuddy:
## Summary

Production verification showed the original `v1` SchedulerDO instance stayed warm after deploys and continued firing the older token set (`tokenCount=4`, including `unknown`). This moves the active scheduler singleton to `v2` and adds an admin-only control route so operators can inspect/pause/resume/refresh named scheduler instances.

## What changed

- `/leaderboard` now kicks `SCHEDULER.idFromName("v2")` instead of `v1`.
- Adds `GET /api/admin/scheduler?name=v2` for status.
- Adds `POST /api/admin/scheduler?name={name}&action=pause|resume|refresh` protected by `X-Admin-Key`.

## Operational actions already performed

- Deployed this patch directly to production to recover the scheduler.
- Paused old `v1` until a far-future timestamp via the new admin route.
- Forced `v2`/`v3` refreshes to verify current behavior.

## Verification

- `npm run lint` passes (existing image warnings only)
- `npm run test -- lib/scheduler/__tests__/tenero-task.test.ts` passes
- `npm run build` passes
- `npx opennextjs-cloudflare build` passes
- Production route `/api/admin/scheduler` is live and authenticated

## Note on Tenero key

`TENERO_API_KEY` is present as a Worker secret, but the supplied value does not raise Worker-side Tenero quota. Local tests without a key also return quota, so the earlier local key test was a false positive. Worker requests still return `x-ratelimit-month-remaining: 0`. Prices will remain empty until we get a Tenero key/header combination that works from Cloudflare Workers.

### DIFF
--- a/app/api/admin/scheduler/__tests__/route.test.ts
+++ b/app/api/admin/scheduler/__tests__/route.test.ts
@@ -1,0 +1,116 @@
+import { describe, it, expect, vi, beforeEach, type Mock } from "vitest";
+import { NextRequest, NextResponse } from "next/server";
+
+vi.mock("@opennextjs/cloudflare", () => ({
+  getCloudflareContext: vi.fn(),
+}));
+
+vi.mock("@/lib/admin/auth", () => ({
+  requireAdmin: vi.fn().mockResolvedValue(null),
+}));
+
+import { getCloudflareContext } from "@opennextjs/cloudflare";
+import { requireAdmin } from "@/lib/admin/auth";
+import { GET, POST } from "../route";
+
+const schedulerStub = {
+  status: vi.fn().mockResolvedValue({ now: 123 }),
+  refreshNow: vi.fn().mockResolvedValue({ tenero: { succeeded: 1 } }),
+  pauseUntil: vi.fn().mockResolvedValue(undefined),
+  resume: vi.fn().mockResolvedValue(undefined),
+};
+
+const schedulerNamespace = {
+  idFromName: vi.fn((name: string) => `id:${name}`),
+  get: vi.fn(() => schedulerStub),
+};
+
+function request(path: string, method = "GET") {
+  return new NextRequest(`https://aibtc.com${path}`, { method });
+}
+
+function mockCloudflareContext() {
+  (getCloudflareContext as Mock).mockResolvedValue({
+    env: { SCHEDULER: schedulerNamespace },
+    ctx: { waitUntil: vi.fn(), passThroughOnException: vi.fn() },
+  });
+}
+
+beforeEach(() => {
+  vi.clearAllMocks();
+  (requireAdmin as Mock).mockResolvedValue(null);
+  mockCloudflareContext();
+});
+
+describe("GET /api/admin/scheduler", () => {
+  it("requires admin auth", async () => {
+    (requireAdmin as Mock).mockResolvedValueOnce(
+      NextResponse.json({ error: "Unauthorized" }, { status: 401 })
+    );
+
+    const response = await GET(request("/api/admin/scheduler"));
+
+    expect(response.status).toBe(401);
+    expect(getCloudflareContext).not.toHaveBeenCalled();
+  });
+
+  it("returns scheduler status with no-store/noindex headers", async () => {
+    const response = await GET(request("/api/admin/scheduler"));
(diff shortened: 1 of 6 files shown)


arc0btc (approved):
Scheduler v2 migration with admin controls — exactly the right response to a sticky DO instance. Production-first fix then this PR to formalize it is a sensible sequence.

**What works well:**
- Auth check is first in both GET and POST — no accidental unauthenticated access path.
- `pause` validation correctly rejects a missing `until` param (null → 0 → not finite future) with a clear 400.
- `schedulerStub()` helper de-duplicates the env cast between GET and POST.
- Named constant `SCHEDULER_INSTANCE_NAME` in `leaderboard/page.tsx` is a clean improvement over the inline string.
- Deployment CI is green and production verified per the PR description.

**[suggestion] `SchedulerStub` type duplication** (`app/api/admin/scheduler/route.ts:5-10`)
This interface is defined inline here. If the Durable Object codebase already exports a `SchedulerDO` type (or will in the future), this local copy will drift silently. Worth checking if there's a shared type to import — even a comment pointing to the canonical source would help.

**[nit] Error message when `until` param is absent**
The validation correctly rejects a missing `until` (0 ≤ Date.now()), but the error says "Provide a future unix-millis `until` value" — same message as an invalid/past value. Minor UX friction for an operator who just forgot the param. Not blocking.

**[question] v3 instance status?**
PR description mentions v3 was also refreshed to verify behavior. Is there a live v3 that could also stay warm and replicate the v1 problem? The admin route handles arbitrary names so it's manageable — just want to make sure v3 isn't also firing stale token sets.

*Operational note:* The leaderboard page uses `ctx.waitUntil()` to warm the scheduler on page loads. With the constant extracted, future instance migrations are a one-line change — good operational hygiene.

secret-mars (approved):
APPROVE on `aa4e484` — focused hotfix, clean scope, already prod-verified.

**Admin auth pattern (`lib/admin/auth.ts`)** — well-thought-through:

1. HMAC-as-constant-time-comparison: `HMAC(k_public, expectedKey)` vs `HMAC(k_public, providedKey)`, then byte compare the digests. The final `hexA !== hexB` is NOT byte-level constant-time, but it doesn't matter — both inputs are 256-bit HMAC outputs, indistinguishable from random, so a short-circuit on the i-th byte leaks information about the HMAC output (which is randomized) rather than about `expectedKey` (the actual secret). This is the textbook RFC-4226 §6 pattern.
2. HMAC key is the literal string `"admin-auth"` — correct here since security flows from `expectedKey` secrecy, not from HMAC-key secrecy. The HMAC layer is the timing-safety primitive only.
3. Both HMACs computed in parallel via `Promise.all` — same wallclock regardless of input — clean.
4. Header-absent branch returns early before HMAC, which has different timing than the wrong-key branch — that's fine since header presence is not a secret.

**Action-surface blast radius:** pause / resume / refresh are all reversible. No `delete` or `destroy` verb. `pauseUntil` validates `until > Date.now()` so you can't accidentally lock in a past-pause that does nothing visible. Good bounding.

**Two non-blocking questions for a follow-up:**

1. **Rate limit on `/api/admin/*`** — HMAC strength (~2^128 forgery space) makes brute-force impractical, so this is defense-in-depth not strictly required. But the existing rate-limit bindings (`RATE_LIMIT_MUTATING`, etc. per `wrangler.jsonc:36-40`) make adding one cheap. Worth considering before more admin routes land under this prefix.
2. **robots.txt + `noindex` for `/api/admin/*`** — JSON 401 responses are unlikely to get indexed, but adding a directory-prefix rule preemptively avoids surprise indexing if a future route returns HTML on a misroute. Cheap preventive.

**Minor:**
- The `task === "all" ? "all" : "tenero"` coercion at line 49 of the new route silently maps any non-"all" value to "tenero". That's fine for now since only one task exists, but worth a `400` once a third task lands so callers don't get a stealth-fallback.
- DurableObject stub type assertion (`as unknown as {...}`) is pragmatic given the SCHEDULER binding shape — could be cleaned by exporting a `SchedulerStub` interface from `lib/scheduler/`, but not blocking.

Production already running this code per the PR body — substantive review is retroactive verification, no merge gate.

Issue #780, opened 2026-05-12
refactor(762b): unify OG cache invalidation via shared helper + identity-refresh shape parity


whoabuddy:
## Background

Council Cycle 9 (steel-yeti, 3-of-4 convergent) flagged a shape inconsistency introduced during the P3.2/P3.3 edge-cache campaign:

- `app/api/challenge/route.ts` was updated to use a unified `addressesToBust: Set<string>` + grouped `Promise.all` shape
- `app/api/identity/[address]/refresh/route.ts` retained the older `middlewarePurges: Promise<void>[]` array pattern

This inconsistent precedent means the two primary write paths have diverged in shape, making future maintenance harder and increasing the risk of alias coverage drift.

## secret-mars narrowing note (Cairn Q1)

When extracting the shared helper, the middleware-OG purge component should be narrowed to **btc/stx forms only**. Per Phase 2.3 crawler-handler scope, middleware OG cache is exclusively keyed on btc:/stx: prefixes — `purgeMiddlewareOgCache(taproot)` and `purgeMiddlewareOgCache(bnsName)` are currently no-op purges (harmless but wasteful). By contrast, the `/api/og` purge via `buildEdgeCacheKey("/api/og", addr)` correctly keeps full enumeration since `/api/og/[address]` does support taproot/bnsName lookups.

## Proposed implementation

Extract `invalidateAllOgCaches(kv: KVNamespace, addresses: Set<string>, opts: { includeMiddleware: boolean })` in `lib/edge-cache.ts`. The helper would:

1. Always invalidate `/api/og` cache for all addresses in the set
2. When `includeMiddleware: true`, also run middleware-OG purge but **filtered to btc/stx forms only**

Adopt this helper in both:
- `app/api/challenge/route.ts` (already has the Set shape; wire up the helper)
- `app/api/identity/[address]/refresh/route.ts` (migrate from `middlewarePurges: Promise<void>[]` array to Set + helper)

## Alias-parity gap (host loop finding)

identity-refresh's `cachedAddresses` Set includes `{btc, stx, taproot?, previousBnsName?, nextBnsName?}` but does NOT include `previousTaprootAddress` for the `update-taproot` action case. This is a minor alias-parity gap relative to challenge route's full previous-alias coverage. Folding this fix into the same cleanup pass is natural — extend `cachedAddresses` to include `previousTaprootAddress` when the refresh action changed it.

## Scope

- `lib/edge-cache.ts` — add shared helper
[...]

PR #785, opened 2026-05-12
fix(771): KV claim fallback when D1 LEFT JOIN misses for erc8004 agents

secret-mars:
## Summary

Resolves #771. Adds a KV claim-fallback at `/api/agents/[address]` for agents whose claim records predate the Phase 2.5 dual-write (#705) and live in KV but not yet in D1's `claims` table (#691 backfill scope).

## Root cause

`/api/verify` reads claim from KV; `/api/agents` reads claim from D1 LEFT JOIN claims since #672 Phase 2.2. For the 708-ish KV-only-claim agents, the D1 LEFT JOIN misses → `mapRowToClaimRecord` returns `null` → `enrichAgentProfile` interprets as "caller confirmed no claim" → `computeLevel` returns 1 → `agent-news/src/services/identity-gate.ts` (hardcoded to `https://aibtc.com/api/agents`) sees Level 1 → `IDENTITY_REQUIRED` on POST `/api/signals`.

Confirmed live on `bc1q73ffx0fwtdvxhs6cfr5hguxsa3pasyg0txyae8` (Opal Gorilla, `erc8004AgentId=319`): `/api/verify` returns Level 2 Genesis; `/api/agents` returns Level 1. Full triage at https://github.com/aibtcdev/landing-page/issues/771#issuecomment-4433226783.

## Fix

```ts
if (shouldFallBackToKVClaim(agent, d1Claim)) {
  // agent narrowed to AgentRecord by type-guard predicate
  logger.info("profile.d1_claim_miss_with_erc8004", {
    btc: agent.btcAddress,
    erc8004AgentId: agent.erc8004AgentId,
  });
  d1Claim = undefined;  // signal enrichAgentProfile to fall back to KV
}
```

Heuristic: `agent.erc8004AgentId !== null` signals on-chain Genesis registration, a strong prior that a claim SHOULD exist. When a D1 LEFT JOIN miss coincides with an erc8004 agent-id, the most likely cause is the backfill gap rather than a true claim absence. Bounded to the Genesis subset (~123 of 951 registered agents) — the Level-1 majority (no `erc8004AgentId`) sees no extra KV read on the cold-miss-read path, preserving Phase 2.2's fan-out elimination for hot paths.

## Files

- **`lib/cache/agent-profile.ts`** — adds `shouldFallBackToKVClaim(agent, d1Claim)` pure helper with TypeScript type-guard predicate (`agent is AgentRecord`), so the route call site can use `agent.btcAddress` without a non-null assertion.
- **`app/api/agents/[address]/route.ts`** — invokes the helper between D1 lookup and `enrichAgentProfile` call; logs `profile.d1_claim_miss_with_erc8004` on the recovery path for operational visibility (helps measure how often the backfill gap fires, can drive #691 prioritization).
- **`app/api/agents/[address]/__tests__/profile-d1.test.ts`** — 6 unit tests for the helper:
  - null agent (no D1 hit) → false
  - undefined d1Claim (already KV-fallback path) → false
  - set d1Claim (D1 hit) → false
  - no `erc8004AgentId` (true Level-1) → false
  - `erc8004AgentId` set + `d1Claim=null` (dual-write gap) → true
  - `erc8004AgentId=0` edge case (valid uint, defensive null/undefined checks)

## Verification path

After deploy, the v279 repro should flip:

```
$ curl https://aibtc.com/api/agents/bc1q73ffx0fwtdvxhs6cfr5hguxsa3pasyg0txyae8
→ level: 2, levelName: "Genesis"   (was Level 1)
```

This unblocks `POST /api/signals` end-to-end via `agent-news`'s identity-gate. The 1h cache in `identity-gate.ts` means the fix takes effect as cache entries naturally expire; operators can force immediate recovery by clearing the affected `agent-level:{address}` KV keys in agent-news.

## Why this approach over alternatives

- **Always fall back on `d1Claim === null`** — adds a KV read per cold-miss for the ~828 Level-1 majority. Defeats Phase 2.2's fan-out elimination.
- **Cross-repo: switch agent-news identity-gate to /api/verify** — works but riskier (cross-repo coordination + cache invalidation across agent-news deploys). Better landed AFTER landing-page-side mitigation.
- **Wait for #691 backfill** — root fix but unknown timeline. This PR is the bridge until #691 completes; once it does, this helper becomes dead code and gets removed.

## Test plan

- [ ] CI green (unit tests + lint + type-check)
- [ ] After merge + deploy, manually verify `/api/agents/bc1q73ffx0fwtdvxhs6cfr5hguxsa3pasyg0txyae8` returns Level 2
- [ ] Spot-check a non-Genesis agent (no `erc8004AgentId`) still returns Level 1 without an extra KV read (logs should NOT show `profile.d1_claim_miss_with_erc8004` for those addresses)
- [ ] Confirm `logger.info("profile.d1_claim_miss_with_erc8004", ...)` appears in production logs for affected agents — gives #691 a measurable signal

## Related

- #691 — root cleanup (backfill KV-only claims into D1); once closed, this helper can be retired
- #783 — sibling gated-retirement pattern (`MIN_EXPECTED_D1_ROWS` after #691); same retirement-gate
- #780 — shape unification (challenge + identity-refresh OG cache); unrelated but adjacent in the OG-cache campaign

### DIFF
--- a/app/api/agents/[address]/__tests__/profile-d1.test.ts
+++ b/app/api/agents/[address]/__tests__/profile-d1.test.ts
@@ -18,8 +18,10 @@ import {
   mapRowToAgentRecord,
   mapRowToClaimRecord,
   computeProfileLevel,
+  shouldFallBackToKVClaim,
 } from "@/lib/cache/agent-profile";
 import type { AgentProfileRow } from "@/lib/cache/agent-profile";
+import type { AgentRecord, ClaimRecord } from "@/lib/types";
 
 // ---------------------------------------------------------------------------
 // Fixtures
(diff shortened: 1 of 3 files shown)


arc0btc (approved):
KV claim fallback gated by `erc8004AgentId` presence — clean bridge fix for the Phase 2.5 dual-write gap until #691 backfill lands.

**What works well:**
- The `shouldFallBackToKVClaim` type-guard predicate is the right abstraction. Making the fallback decision pure and testable (rather than inline in the route) means the heuristic is easy to reason about and retire cleanly once #691 ships.
- Test coverage is thorough — especially the `erc8004AgentId === 0` edge case. The defensive `!== null && !== undefined` is justified given that 0 is a valid registry id.
- Scoping the KV read to the Genesis subset (~123 agents) correctly preserves Phase 2.2's fan-out elimination for the Level-1 majority. Good cost awareness.
- The `logger.info("profile.d1_claim_miss_with_erc8004", ...)` gives #691 a concrete operational signal — measuring backfill gap frequency from production logs is exactly the right way to prioritize that work.

**[question] KV claim validation on the fallback path**
When `enrichAgentProfile` falls back to `kv.get('claim:{btcAddress}')`, does it validate the claim contents (status, signatures, expiry) before computing level? If KV holds a stale or malformed claim for any of the ~123 affected agents, this bridge could surface incorrect Level-2 data. Not a blocker — just want to confirm the KV read path already validates before trusting the result.

**[suggestion] Simplify the null/undefined guard in `shouldFallBackToKVClaim`**
The `!== null && !== undefined` double-check on `erc8004AgentId` is defensive for good reason (test comment explains it), but `!= null` (loose inequality) is a direct equivalent that covers both cases in a single expression. Either is fine — just noting it exists if you want to tighten it in a follow-up.
```suggestion
    agent.erc8004AgentId != null
```

**[nit] JSDoc length**
The 26-line JSDoc on `shouldFallBackToKVClaim` is thorough but would transfer better to the #771 issue comment or PR description than to the function itself. The background on Phase 2.5, #705, and #691 is valuable for someone reading the PR; a future reader of the file mostly needs: "fallback heuristic for Genesis agents during the backfill gap — see #691 for retirement." Take it or leave it.

**Performance:** No concerns. KV read only added for the Genesis subset; unchanged for the Level-1 majority.

**Composition:** N/A — API route + utility, no UI components.

**UI/Accessibility:** N/A — no JSX or component changes.

Solid fix. The operational observability built in (`logger.info` keyed off `profile.d1_claim_miss_with_erc8004`) makes this self-documenting in production. Unblocks signal submission for the affected erc8004 agents immediately.

PR #786, opened 2026-05-12
docs(762): add edge-cache PR checklist (closes #782)

secret-mars:
## Summary

Closes #782. Adds `docs/edge-cache-pr-checklist.md` — a campaign-level reference distilled from the steel-yeti Council Cycle 3/4/7/8 advisories during the P3.2/P3.3 campaign (PRs #774, #775).

## Six required fields + one optional

1. **Cache namespace** — `caches.default` (global) vs middleware OG vs KV
2. **Canonical key builder** — read + write paths MUST share (Cycle 3/4 finding on #774)
3. **Purge helper** — best-effort vs throws; helper signature inputs
4. **Affected write paths** — exhaustive list + intentional exclusions
5. **Vary behavior** — preserved on cached clones (Cycle 7 finding on #774)
6. **waitUntil / non-blocking** — Workers `ctx.waitUntil` + Node `void+.catch` fallback (Cycle 3 Cairn finding)
7. **(Optional) Zone-CDN behavior** — `s-maxage` gap per #781

## Why this shape

The fields above were each surfaced multiple times across PRs in the P3 campaign before the campaign converged on a shared shape. Each field corresponds to a recurring finding pattern:

- **Field 2** addresses the cache-key duplication finding from Cycle 3 / 4 (PR #774) where the read path used an inline template literal and the write path used the helper — they drifted, and the purge missed.
- **Field 5** addresses the Vary mismatch from Cycle 7 (PR #774) — without `new Headers(response.headers)` on the cached clone, the cached entry could mutate the live response or vice versa.
- **Field 6** addresses the unhandled-rejection concern from Cycle 3 Cairn — both branches need explicit handling.

The doc also includes a filled-out example checklist (P4.2 heartbeat dual-write) so a PR author has a concrete template, and reference implementations pointing to #774's fixup pattern + #780's unification target + #781's zone-CDN gap.

## Timing

Per #782: lands before the next P3 sibling (B6.2 BNS reverse-index D1 or P4.2 heartbeat dual-write). The doc is reference material — no behavioral code change — so review burden is light.

## Maintenance

Doc has a "Maintenance" section with explicit triggers for when to update it:
- New shared helper in `lib/edge-cache.ts` → expand field-2/3 helper lists
- New cache namespace used (e.g. KV-as-cache) → expand field 1
- Council retrospective surfacing a 7th repeated finding → promote to required
- Zone-CDN purge gap closes (#781 lands) → field 7 moves to required + obsoletes manual-purge mitigation note

## Test plan

- [ ] CI green (no code change — just a markdown doc; only Build / Lint really apply)
- [ ] Spot-check that the example checklist (P4.2 heartbeat) is structurally valid — the field labels match the headings, the helper references match `lib/edge-cache.ts` exports
- [ ] (Author-side) confirm the doc renders cleanly on github.com after merge

## Refs

- Closes #782
- Distilled from steel-yeti Cycle 3/4/7/8 advisories
- Cited from #774 v276 review (issuecomment-4432812896 — Forge +1)
- Related: #780 (unification target), #781 (zone-CDN gap), #774 + #775 (reference patterns)

### DIFF
--- a/docs/edge-cache-pr-checklist.md
+++ b/docs/edge-cache-pr-checklist.md
@@ -1,0 +1,126 @@
+# Edge-Cache PR Checklist
+
+Closes #782. Land this doc reference alongside any PR that touches the edge-cache layer (Workers `caches.default` API, `lib/edge-cache.ts` helpers, or write-path invalidation).
+
+The fields below were distilled from the steel-yeti Council Cycle 3/4/7/8 advisories during the P3.2/P3.3 campaign (PRs #774, #775). Each field corresponds to a finding that recurred across sibling PRs until the campaign converged on a shared shape. Filling out this checklist in the PR description means the next P3 sibling doesn't re-discover the same six concerns ad-hoc.
+
+## When this checklist applies
+
+Fill it out in the PR body for any change that:
+
+- Adds a new `withEdgeCache(...)` wrapper around a route handler
+- Adds or modifies a `buildEdgeCacheKey(...)` / `buildMiddlewareOgCacheKey(...)` call site
+- Adds or modifies cache invalidation (`invalidateEdgeCache`, `purgeMiddlewareOgCache`) on a write-path route
+- Reads from or writes to `caches.default` directly (rare — prefer `lib/edge-cache.ts` wrappers)
+
+If you're only changing rate-limit bindings, KV-only paths, or D1-only paths with no edge-cache surface, skip the checklist.
+
+## Required fields
+
+### 1. Cache namespace
+
+Which Workers cache namespace does this PR touch? In practice this is one of:
+
+- `caches.default` — global per-zone Workers cache (see `lib/edge-cache.ts:withEdgeCache`)
+- `caches.default` — middleware OG cache, keyed under the `middleware:og:` prefix (see `lib/edge-cache.ts:buildMiddlewareOgCacheKey`)
+- A KV binding (`VERIFIED_AGENTS`, etc.) — note this is NOT the Workers cache layer; KV reads/writes have different semantics
+
+State the binding name explicitly. If the PR is multi-layer (cache + KV), list each.
+
+### 2. Canonical key builder
+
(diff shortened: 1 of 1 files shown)


arc0btc (approved):
Adds an edge-cache PR checklist distilled from the P3.2/P3.3 Council Cycle advisories — good call surfacing this before the next sibling lands.

**What works well:**
- Each field traces back to a specific recurring finding (Cycle 3/4 key duplication, Cycle 7 Vary mismatch, Cycle 3 Cairn waitUntil). That provenance makes the fields feel load-bearing rather than aspirational.
- The filled-out P4.2 example is exactly the right format — a PR author can copy it and fill in blanks rather than guessing at the structure.
- Maintenance triggers are explicit and tied to concrete code events (`lib/edge-cache.ts` helper additions, #781 landing). These don't go stale the way "keep this up to date" instructions do.
- The intentional exclusion note in Field 4 (heartbeat doesn't invalidate `/api/og` because it doesn't render `lastActiveAt`) is the kind of discipline that prevents the checklist from becoming a rubber-stamp.

**[nit]** `docs/edge-cache-pr-checklist.md`, line 3
The file body opens with `"Closes #782."` — that's a GitHub issue-close directive. It belongs in the PR description (where it already is) rather than in the doc file itself. In the rendered file on GitHub it reads as stray content rather than documentation.

```suggestion
Land this doc reference alongside any PR that touches the edge-cache layer (Workers `caches.default` API, `lib/edge-cache.ts` helpers, or write-path invalidation).
```

**Code quality notes:**
Field 1 lists `caches.default` twice as separate bullet points. The distinction (global per-zone vs. middleware OG with `middleware:og:` prefix) is real, but the duplication could trip a reader who expects each bullet to be a different binding. A sub-bullet or inline parenthetical (`caches.default — global` / `caches.default — middleware OG, keyed under middleware:og: prefix`) would make the split clearer without adding new content.

**Performance / Composition / UI-Accessibility:**
Not applicable — markdown-only change.

**Operational note:** The `edge-cache-auth-gate-leak` pattern we captured (edgeCacheMatch before BIP-322 auth = author-only data leak) is orthogonal to this checklist's scope, but worth knowing: if a future PR using this checklist adds a new cached route that returns auth-gated data, Field 4's write-path list should also flag whether *reads* need auth gating before the cache lookup. The checklist as written focuses on invalidation, not on cache-before-auth risks. Minor gap, worth a follow-up doc note when it becomes relevant.

PR #792, opened 2026-05-13
refactor(leaderboard): move Tenero price read from SSR to client

biwasxyz:
## Summary

- Leaderboard SSR no longer reads `tenero:price:*` from KV. Server returns per-token aggregates (`tokenId`, `sumAmountIn`, `decimals`); the client fetches `/api/prices` and computes USD volume in-browser.
- `TOKEN_DECIMALS` moved out of `app/leaderboard/page.tsx` into `app/leaderboard/tokens.ts` so the server passthrough and the client computation share one source of truth.
- Backend untouched: the `SchedulerDO` Tenero refresh, `lib/external/tenero/*`, and `/api/prices` endpoint are unchanged. Only the read site moves.

While prices are in flight the volume cell shows `…`; once `/api/prices` resolves the cells fill in. Fetch failure falls back to the existing "—" state.

## Test plan

- [ ] Visit `/leaderboard` — rows render immediately, volume column briefly shows `…`, then resolves to USD values.
- [ ] Network tab: SSR HTML does not include precomputed USD volume; one client-side `GET /api/prices` request is issued.
- [ ] Block `/api/prices` (devtools) — rows still render, volumes render as "—" (no crash, no infinite spinner).
- [ ] Confirm the partial-priced asterisk still appears when a token in an agent's mix has no cached price.
- [ ] Mobile breakpoint shows the same `…` → value transition.

### DIFF
--- a/app/leaderboard/LeaderboardClient.tsx
+++ b/app/leaderboard/LeaderboardClient.tsx
@@ -1,8 +1,15 @@
 "use client";
 
+import { useEffect, useState } from "react";
 import Link from "next/link";
 import { generateName } from "@/lib/name-generator";
 import { truncateAddress, formatRelativeTime } from "@/lib/utils";
+
+export interface LeaderboardTokenAggregate {
+  tokenId: string;
+  sumAmountIn: number;
+  decimals: number;
+}
 
 export interface LeaderboardRow {
   stxAddress: string;
@@ -13,16 +20,18 @@ export interface LeaderboardRow {
   tradeCount: number;
   latestTradeAt: number;
   /**
-   * USD volume computed server-side from KV-cached Tenero prices. Comes in
-   * as a plain number so this component stays presentational — no fetch,
-   * no localStorage, no useEffect.
+   * Per-token aggregates from D1. USD volume is computed client-side
+   * from `/api/prices` so SSR doesn't block on the KV price cache.
    */
+  tokens: LeaderboardTokenAggregate[];
+}
+
+interface PricesResponse {
+  prices?: Record<string, { priceUsd: number | null; fetchedAt: number }>;
+}
+
+interface RowVolume {
   volumeUsd: number;
-  /**
-   * False if any token in the agent's volume couldn't be priced from KV
-   * (cold scheduler, paused, or token has no published price). Lets the UI
-   * footnote the row instead of misleadingly under-reporting.
-   */
   allPriced: boolean;
 }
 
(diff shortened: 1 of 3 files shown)


arc0btc (approved):
Moves Tenero price computation off the SSR render path — the KV read was the main latency contributor on the leaderboard, and this eliminates it cleanly.

**What works well:**
- The cancellation flag in `useEffect` is correct — prevents stale state updates after unmount/route change.
- Loading state uses `aria-label="Loading USD volume"` on the placeholder span — good accessibility touch.
- `computeVolumes` extracted as a pure function — easy to test and reason about in isolation.
- Empty-rows early exit skips the network round-trip entirely.
- The ellipsis loading placeholder uses the proper U+2026 character, not three dots.
- Sort order in SSR is by `tradeCount` (not `volumeUsd`), so client-side price computation landing async doesn't cause any layout shift in row order. Clean.

**[nit]** `app/leaderboard/page.tsx` has a `see #TBD for the rationale` placeholder in a comment. Either link a real issue or drop the reference.

**[question]** `useEffect([rows])` — since `rows` is serialized from a Server Component, the reference is stable across renders and this fires effectively once on mount. Was the intent to also re-fetch prices if rows ever update (e.g., if this component is later used in a context with dynamic row updates), or is `[]` semantically more honest here? Either is defensible — just confirming the intent.

**[suggestion]** The initial HTML payload now carries per-token aggregates (tokenId, sumAmountIn, decimals) for every row rather than just the precomputed `volumeUsd`. For a leaderboard with many traders each holding multiple tokens, this grows the serialized payload. Probably fine at current scale, but worth a comment noting the trade-off — future you will appreciate it when the payload grows.

---

## Performance

No CRITICAL violations. The move from SSR KV reads to a client `/api/prices` fetch is the right call for TTFB — KV reads on the render path add latency that users feel. The client now sees the page structure immediately, then prices fill in, which is a better progressive rendering story.

No waterfall issues — single independent fetch, no sequential awaits.

## Composition

Clean extraction of `TOKEN_DECIMALS` and `DEFAULT_TOKEN_DECIMALS` into `tokens.ts` as the single source of truth. The `computeVolumes` function is appropriately kept local (not exported) since it's an implementation detail of this component.

No boolean prop issues, no forwardRef usage, no composition anti-patterns.

## UI/Accessibility

- ✅ `aria-label="Loading USD volume"` on the loading span — explicitly labels the loading state for screen readers.
- ✅ Proper U+2026 ellipsis character in loading state.
- ✅ Mobile list handles the `volumes === null` loading state with "…" matching the desktop behavior.
- No BLOCKING violations found.

**Code quality notes:**
The `PricesResponse` and `RowVolume` interfaces are file-local (not exported), which is correct — they're implementation details of the price-fetch logic. `LeaderboardTokenAggregate` is exported because `page.tsx` needs it to construct `LeaderboardRow` values — the export scope is appropriate.

PR #800, opened 2026-05-13
feat(tenero): derive scheduler refresh set from D1 swaps table

biwasxyz:
## Summary

The SchedulerDO has been refreshing prices for a hand-curated 3-token list (`STATIC_TOKEN_IDS`). Now that the `swaps` table is the source of truth for what agents actually trade, the refresh schedule should follow the data instead of a hardcoded list someone has to remember to update.

## What changed

| File | Change |
|---|---|
| `lib/external/tenero/tokens.ts` | New `getActiveTokenIds(db)` + `MAX_TRACKED_TOKENS=50` + `isValidTokenId(id)` regex filter. Unions the static core with D1-discovered tokens, ranked by trade count, junk-filtered. |
| `lib/external/tenero/index.ts` | Barrel re-exports for the new symbols. |
| `worker.ts` (`SchedulerDO.runTenero`) | Calls `getActiveTokenIds(this.env.DB)` instead of importing `STATIC_TOKEN_IDS` directly. Logs the resolved count. |
| `app/api/competition/tracked-tokens/route.ts` | New public read endpoint exposing the active set for debug visibility. 1-min cache. |
| `lib/external/tenero/__tests__/tokens.test.ts` | 14 unit tests: regex shape filter, union semantics, bounds, fallback paths. |

## Behavior

- **Happy path**: D1 query returns distinct (token_in ∪ token_out) from successful agent/cron swaps, ranked by trade count, capped at `MAX_TRACKED_TOKENS`, filtered by `isValidTokenId` (matches `stx` literal OR `S[PMTN]…contract::asset` shape — drops the `\"unknown\"` parser sentinel and any other malformed ids). Result is `[...STATIC_TOKEN_IDS, ...dynamic]` deduplicated.
- **Empty swaps table**: returns `STATIC_TOKEN_IDS` only (cold-start safe — never an empty refresh set).
- **Missing D1 binding** (local dev, preview without DB): returns `STATIC_TOKEN_IDS` only.
- **D1 query throws**: caught, returns `STATIC_TOKEN_IDS` only — a transient DB blip doesn't drop all price refreshes.

## Quota note

`STATIC_TOKEN_IDS` × 5min refresh = 3 × 8,640 ticks/month ≈ 26k Tenero calls/month — well under the 50k free-tier ceiling.

If dynamic discovery actually saturated to `MAX_TRACKED_TOKENS=50`, the math is 50 × 8,640 ≈ 432k/month — would burst free-tier and trigger the existing `TENERO_MONTH_QUOTA_BACKOFF_MS = 24h` adaptive backoff in `runTeneroTask`. The scheduler won't get banned (it backs off cleanly) but prices would go stale for the remainder of the month.

In practice today the swaps table has 1 row, so dynamic discovery adds 0-1 tokens. Once `env.TENERO_API_KEY` is set (already plumbed end-to-end via `SchedulerDO.lookupTeneroApiKey` → `runTeneroTask` → `teneroFetch`), the authenticated tier removes the cap concern.

## What doesn't change

- `STATIC_TOKEN_IDS` is still exported and is the always-include core. Dynamic discovery layers on top.
- `/api/prices` still uses `STATIC_TOKEN_IDS` directly. That endpoint's contract is \"the 3 baseline tokens\" and its aggressive Cache-Control (`s-maxage=86400`) would make a dynamic list confusing. Decoupled by design.
- `runTeneroTask`'s signature is unchanged (still takes `tokenIds: readonly string[]`). Only the caller's source changes.
- No DB migration, no DO storage shape change.

## Why SchedulerDO, not a Worker cron trigger

(Came up in design conversation.) The DO + alarms pattern was chosen because Tenero rate-limit semantics need state across runs — when a tick hits `x-ratelimit-month-remaining: 0`, the scheduler writes `nextRunAfter.tenero = now + 24h` to DO storage and the next alarm checks that. With a cron-only setup, that state would have to live in KV with manual coordination. Same pattern as `x402-sponsor-relay`. Switching to dynamic token discovery doesn't change the DO choice — just the input list.

## Test plan

- [x] `npm test -- --run lib/external/tenero lib/scheduler app/api/admin/scheduler` — 26+ tests pass.
- [x] `npm run build` — clean.
- [ ] After deploy: `curl https://aibtc.com/api/competition/tracked-tokens` should return the static core (3 tokens) since the swaps table has only the one `\"unknown\"`-tagged trade today (which gets filtered out by `isValidTokenId`).
- [ ] After more trades land: the same endpoint should reflect new tokens. The next SchedulerDO tick after a new token first appears will start refreshing its price into KV.

## Related

- The `\"unknown\"` filter is what made me confident dropping the junk-safety reason for keeping the list static. The parser bug that produced `\"unknown\"` was fixed in commit `21f16f4`, but the filter belt-and-suspenders is cheap.
- This is foundational for the (forthcoming) `GET /api/competition/pnl?address=` endpoint — server-side P&L will read swaps from D1, get distinct tokens, and lean on KV prices populated by this exact refresh path.

### DIFF
--- a/app/api/competition/tracked-tokens/route.ts
+++ b/app/api/competition/tracked-tokens/route.ts
@@ -1,0 +1,87 @@
+// CACHE_INVARIANTS:POSTURE=public-only-get
+// Public read of the dynamic token set the SchedulerDO refreshes Tenero
+// prices for. Derived from the `swaps` table on each request — see
+// `lib/external/tenero/tokens.ts:getActiveTokenIds`. Read-only, no auth.
+
+import { NextRequest, NextResponse } from "next/server";
+import { getCloudflareContext } from "@opennextjs/cloudflare";
+import {
+  STATIC_TOKEN_IDS,
+  MAX_TRACKED_TOKENS,
+  getActiveTokenIds,
+} from "@/lib/external/tenero";
+
+function selfDocResponse() {
+  return NextResponse.json(
+    {
+      endpoint: "/api/competition/tracked-tokens",
+      method: "GET",
+      description:
+        "The Tenero refresh set the SchedulerDO uses on each tick. Union of the always-include static core and distinct token_in / token_out from successful agent/cron swaps in D1, junk-filtered for shape, ranked by trade count, capped at MAX_TRACKED_TOKENS. Falls back to the static core on D1 query failure.",
+      queryParameters: {
+        docs: {
+          type: "string",
+          description: "Pass ?docs=1 to return this documentation payload instead of data",
+          example: "?docs=1",
+        },
+      },
+      responseFormat: {
+        tokens: "string[] — the active set (static core ∪ dynamic discovery)",
+        count: "number — length of tokens",
+        source:
+          "'dynamic' when D1 derivation succeeded, 'static-fallback' on failure or missing binding",
+        staticCore: "string[] — the always-include core that backs the fallback",
+        maxTracked: "number — upper bound on dynamic discovery per tick (MAX_TRACKED_TOKENS)",
+      },
+      relatedEndpoints: {
+        allowlist:
+          "GET /api/competition/allowlist — contracts + functions the verifier will accept",
+        prices: "GET /api/prices — cached USD prices for the static core only",
+      },
+      notes: [
+        "Dynamic discovery falls back to STATIC_TOKEN_IDS when there's no DB binding or the query throws.",
+        "Tokens are ordered with the static core first, then dynamic entries by descending trade count.",
+        "The set is recomputed on each SchedulerDO tick (5 min cadence); this endpoint reports what the *next* tick would use.",
+      ],
+    },
+    {
+      headers: {
+        // 1 min browser, 1 min shared — the set can shift as agents trade.
+        "Cache-Control": "public, max-age=60, s-maxage=60",
+      },
+    }
+  );
+}
+
+export async function GET(request: NextRequest) {
+  const { searchParams } = new URL(request.url);
+  if (searchParams.get("docs") === "1") return selfDocResponse();
+
+  const { env } = await getCloudflareContext({ async: true });
+  const db = env.DB as D1Database | undefined;
+  const tokens = await getActiveTokenIds(db);
+
+  // `getActiveTokenIds` falls back to STATIC_TOKEN_IDS on any failure; we
+  // can't distinguish "fallback" from "static core happened to win" without
+  // a side-channel, so the heuristic is: if the result has more tokens than
+  // the static core, dynamic discovery added something; if it equals the
+  // core exactly we couldn't add anything (could be empty table OR DB error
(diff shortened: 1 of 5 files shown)


arc0btc (approved):
Derives the SchedulerDO refresh set from D1 swaps instead of a hardcoded list — good architectural move, and the fallback chain (missing binding → empty table → query throws → static core) is exactly right.

**What looks good:**
- The three-layer fallback in `getActiveTokenIds` is production-hardened: no code path leaves the scheduler with an empty token set
- Regex compiled once at module scope — no per-call allocation
- 14 unit tests cover the full fallback/filter/dedup surface, including the exact bind-argument assertion
- Route correctly caches at 1 min given the 5-min tick cadence
- SQL-level `!= 'unknown'` pre-filter + JS-level `isValidTokenId` post-filter is belt-and-suspenders; cheap and correct

**[suggestion] Index coverage for the D1 query** (`lib/external/tenero/tokens.ts`)
The UNION ALL + GROUP BY + ORDER BY across `swaps` will be a full table scan if there's no index on `(source, tx_status, token_in)` and `(source, tx_status, token_out)`. Today with 1 row it's invisible; once the swaps table reaches 10k+ rows this fires on every 5-min SchedulerDO tick. Consider a follow-up migration that adds composite indexes before the swaps volume makes it observable:

```sql
CREATE INDEX IF NOT EXISTS idx_swaps_token_in  ON swaps (source, tx_status, token_in);
CREATE INDEX IF NOT EXISTS idx_swaps_token_out ON swaps (source, tx_status, token_out);
```

**[nit] Merged set can slightly exceed MAX_TRACKED_TOKENS**
`LIMIT MAX_TRACKED_TOKENS` applies to the *dynamic* results, but `[...STATIC_TOKEN_IDS, ...dynamic]` always adds the 3-token static core on top. If all 50 dynamic slots are filled by tokens not already in the static core, the scheduler sees 53 tokens. The PR's quota math (50 × 8,640) uses 50 as the ceiling — it's conservative enough that 53 doesn't matter, but worth noting in the quota comment for accuracy.

**[question] `source` field accuracy in the route**
The heuristic `tokens.length > STATIC_TOKEN_IDS.length ? 'dynamic' : 'static-fallback'` will report `'static-fallback'` if the D1 query succeeded but all discovered tokens happened to already be in the static core. The code comment already acknowledges this — just confirming it's intentional and not worth a schema change at this scale.

**Code quality notes:**
The `cnt: number | string` type on `TrackedTokenRow` handles D1's type coercion defensively, but `cnt` is never read after filtering (only `r.id` is used). The type is harmless and the defensive intent is clear — leaving as-is is fine.

**Operational context:**
We run SchedulerDO patterns ourselves for the x402-sponsor-relay equivalent — the DO + alarms state-carrying approach for rate-limit tracking is solid and the design note in the PR body is accurate. Dynamic discovery aligns with what we'd expect once agent trading volume picks up.

PR #829, opened 2026-05-13
feat(leaderboard): add competition countdown

whoabuddy:
## Summary
- add a compact countdown panel to the trading leaderboard
- use COMP_START_TIMESTAMP as the single source of truth for the start time
- switch the panel to a live state once the start time has passed

## Verification
- npm run build
- npm run typecheck currently fails on existing test typing errors unrelated to this change

## Notes
- Plain next dev cannot render /leaderboard in this repo because getCloudflareContext requires OpenNext Cloudflare dev initialization; npm run build passed.

### DIFF
--- a/app/leaderboard/CompetitionCountdown.tsx
+++ b/app/leaderboard/CompetitionCountdown.tsx
@@ -47,12 +47,12 @@ export default function CompetitionCountdown({
     () => formatStartTime(startTimestamp),
     [startTimestamp]
   );
-  const parts = useMemo(() => countdownParts(msRemaining), [msRemaining]);
+  const parts = countdownParts(msRemaining);
 
   useEffect(() => {
     if (isLive) return;
-    const id = window.setInterval(() => setNowMs(Date.now()), SECOND_MS);
-    return () => window.clearInterval(id);
+    const id = setInterval(() => setNowMs(Date.now()), SECOND_MS);
+    return () => clearInterval(id);
   }, [isLive]);
 
   return (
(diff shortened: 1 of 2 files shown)


arc0btc (approved):
Adds a compact competition countdown panel to the leaderboard — clean implementation with a solid pattern for avoiding hydration mismatches.

**What works well:**
- `initialNowMs` passed from the Server Component is the right approach: `Date.now()` runs server-side, `useState(initialNowMs)` initializes from the prop, so hydration is always consistent. No client-side `new Date()` in the render path.
- `aria-label="Competition start countdown"` on the `<section>` and `aria-hidden` on the decorative dot are both correct.
- `tabular-nums` on the digit cells — numbers jump without it.
- The interval cleanup tied to `[isLive]` is correct: when the countdown hits zero, the effect re-runs, takes the early-return path, and cleans up the previous interval. No leak.
- `COMP_START_TIMESTAMP` as single source of truth eliminates the classic "start time is different in two places" class of bug.

**[suggestion] Drop the `useMemo` on `countdownParts`** (`CompetitionCountdown.tsx:55`)

`countdownParts` does four `Math.floor` calls on scalars — O(1), negligible cost. Since `msRemaining` changes every second, the memo never serves a cache hit in steady state. Just inline the call:

```suggestion
  const parts = countdownParts(msRemaining);
```

**[nit]** `window.setInterval` / `window.clearInterval` — the `window.` prefix is redundant inside a `"use client"` component; the globals are in scope and read more cleanly.

## Performance

No issues. No sequential awaits, no barrel imports, no large sync loads. Effect cleanup is correct.

## Composition

No issues. Self-contained component with a clean props interface.

## UI/Accessibility

No blocking findings. `<section aria-label>` provides the landmark, status copy is screen-reader exposed, decorative dot is `aria-hidden`, no interactive elements missing keyboard handling, `tabular-nums` applied correctly.

PR #834, opened 2026-05-13
perf(d1): cache heartbeat unread-count behind 30s edge cache

whoabuddy:
## Summary

- Adds `lib/inbox/cached-counts.ts` with `cachedUnreadCount()` — a thin `caches.default` wrapper around `countInboxMessagesFromD1` for the heartbeat path
- Swaps `app/api/heartbeat/route.ts` to call `cachedUnreadCount` instead of the live D1 query directly
- Updates heartbeat test mock to target the new module

## Context

Part of quest **d1-count-bill-stop** (P2 Bandage B). Baseline: ~26–34B D1 rows/day at $30/day overage.

P1 (#833) cut `/api/inbox/[address]` COUNTs from 4 to 2 (derived counts from fetched rows). This PR (P2) wraps the remaining hot-path COUNT in heartbeat.

The heartbeat route fires a live `SELECT COUNT(*) WHERE read_at IS NULL` on **every** GET orientation and POST check-in with no caching. At ~1,000 agents checking in every 5 minutes, that is ~288 D1 rows-read/agent/day from heartbeat alone — a primary overage driver.

## Cache design

| Decision | Choice | Reason |
|----------|--------|--------|
| Cache layer | `caches.default` | Zero KV ops; in-process per-colo store; preserves 86% KV-op cut from predecessor quest |
| TTL | 30 seconds | Heartbeat cadence = 5 min; 30s drift invisible to agents. Conservative to keep unread count fresh for orientation. |
| Wrapper shape | Thin numeric cache (`lib/inbox/cached-counts.ts`) | `withEdgeCache` wraps full HTTP Responses; a single `number` fits a minimal Request/Response pair. No `ctx.waitUntil` complexity needed — `caches.default` put is in-process with negligible latency. |
| Error handling | Fail-open (returns 0) | Mirrors existing `fetchUnreadCount` try/catch; heartbeat orientation must never be blocked by a transient D1 error |

## Verify

- [x] `npm run build` passes
- [x] `npx vitest run app/api/heartbeat/` — 7/7 tests pass
- [x] `caches.default` path: second heartbeat request within 30s window for same address skips D1
- [x] Fallback path: `caches.default` absent (Node/next dev) → live D1 query, no error

## Post-merge smoke checklist

The operator handles T+1h/T+4h/T+24h D1 rows-read verification via Cloudflare analytics.

- [ ] T+0 — merge time recorded in quest artifact
- [ ] T+1h — D1 rows-read trending down vs. P1 baseline
- [ ] T+4h — sustained reduction visible
- [ ] T+24h — full-day delta vs. 26–34B baseline

**Note:** `caches.default` scope is per Cloudflare colo. Multiple colos each miss once per 30s window — acceptable for this use case.

### DIFF
--- a/app/api/heartbeat/__tests__/stx-db-threading.test.ts
+++ b/app/api/heartbeat/__tests__/stx-db-threading.test.ts
@@ -22,8 +22,8 @@ vi.mock("@/lib/agent-lookup", () => ({
   lookupAgentWithLevel: vi.fn(),
 }));
 
-vi.mock("@/lib/inbox/d1-reads", () => ({
-  countInboxMessagesFromD1: vi.fn().mockResolvedValue(0),
+vi.mock("@/lib/inbox/cached-counts", () => ({
+  cachedUnreadCount: vi.fn().mockResolvedValue(0),
 }));
 
 vi.mock("@/lib/levels", () => ({
--- a/app/api/heartbeat/route.ts
+++ b/app/api/heartbeat/route.ts
@@ -7,7 +7,7 @@ import { X_HANDLE } from "@/lib/constants";
 import { ACTIVE_BEATS_LIST } from "@/lib/news-beats";
 import type { AgentRecord, ClaimStatus } from "@/lib/types";
 import { generateName } from "@/lib/name-generator";
-import { countInboxMessagesFromD1 } from "@/lib/inbox/d1-reads";
+import { cachedUnreadCount } from "@/lib/inbox/cached-counts";
 import {
   CHECK_IN_MESSAGE_FORMAT,
   buildCheckInMessage,
@@ -116,7 +116,7 @@ function getNextAction(
 async function fetchUnreadCount(db: D1Database | undefined, btcAddress: string): Promise<number> {
   if (!db) return 0;
   try {
-    return await countInboxMessagesFromD1(db, btcAddress, "unread");
+    return await cachedUnreadCount(db, btcAddress);
   } catch {
     return 0;
   }
--- a/lib/inbox/cached-counts.ts
+++ b/lib/inbox/cached-counts.ts
@@ -1,0 +1,134 @@
+/**
+ * Edge-cache wrappers for inbox COUNT(*) queries that are called on every
+ * heartbeat request.
+ *
+ * ## Why this file exists
+ *
+ * countInboxMessagesFromD1 fires a live SELECT COUNT(*) on every heartbeat GET
+ * and POST. Agents heartbeat every 5 minutes; at ~1,000 registered agents the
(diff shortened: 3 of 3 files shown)


arc0btc (approved):
Wraps the heartbeat unread count query behind a 30s `caches.default` layer to eliminate the live D1 COUNT on every orientation/check-in. The design choices (caches.default over KV, 30s TTL, per-address key, fail-open) are all well-reasoned and correctly implemented.

**What works well:**
- The cache key lowercases the BTC address for canonical matching — same pattern as `buildEdgeCacheKey` in `lib/edge-cache.ts`, which is exactly right.
- The `typeof parsed === "number"` guard handles malformed cached values safely and falls through to a live query rather than throwing.
- The `caches.default` availability guard is correct and keeps local dev working without a cache mock.
- Test mock update is minimal and targets the right module.

## Performance

Nothing to flag — this PR IS the performance fix.

## Composition

No component changes.

## UI/Accessibility

No UI changes.

---

**[suggestion] `cache.put` failure silently discards the fetched count** (`lib/inbox/cached-counts.ts:119-127`)

In the cache-miss path, `cache.put` runs inside the outer `try/catch`. If `cache.put` throws unexpectedly, the entire catch block returns `0` — even though the D1 query already succeeded and `count` is correct. The cache write is best-effort; its failure shouldn't affect the return value.

```suggestion
    // Cache miss: fetch from D1 and store.
    const count = await countInboxMessagesFromD1(db, btcAddress, "unread");

    // Best-effort cache write — failure must not discard a successful D1 result.
    try {
      const storedResponse = new Response(JSON.stringify(count), {
        headers: {
          "Content-Type": "application/json",
          "Cache-Control": `public, max-age=${UNREAD_COUNT_CACHE_TTL_SECONDS}`,
        },
      });
      await cache.put(cacheKey, storedResponse);
    } catch {
      // Cache write failure is non-fatal.
    }

    return count;
```

**[suggestion] `getDefaultCache()` is duplicated from `lib/edge-cache.ts`** (`lib/inbox/cached-counts.ts:68-74`)

`lib/edge-cache.ts` has an identical unexported implementation. Exporting it from `edge-cache.ts` and importing it here would eliminate the duplication and keep the caching utilities in sync if the globalThis guard ever needs updating.

**[nit]** `CACHE_HOST = "https://internal.cache"` vs `"https://cache.aibtc.local"` in `edge-cache.ts`. Both isolate fine; just noting the inconsistency in case you want to consolidate to one synthetic host pattern.

*Operational note:* Heartbeat is one of the highest-frequency paths in the agent network — we see Arc's own heartbeat calls every 5 minutes. A 30s edge cache is exactly the right lever here; even per-colo miss-once behavior is fine for orientation-level counts.

PR #843, opened 2026-05-14
feat(bounty): native bounty system — replaces bounty.drx4.xyz proxy

biwasxyz:
## Summary

Replaces the external `bounty.drx4.xyz` proxy with a first-party bounty system. Genesis-level (L2+) agents post bounties, any Registered (L1+) agent submits, the poster accepts a winner and proves payment with a **confirmed on-chain sBTC transaction** that the platform verifies on Hiro — including a required memo (`BNTY:{bountyId}`) so the same transfer cannot be passed off as a payment for a different bounty.

Phases delivered:
- **Phase 1** — backend: D1 migration, lib, 9 API routes, 39 unit tests
- **Phase 2** — discovery: agent.json / llms.txt / llms-full.txt / heartbeat / skill.md / docs / openapi.json / CLAUDE.md
- **Phase 3** — UX: /bounty pages rewired from bounty.drx4.xyz to /api/bounties; new /bounty/new instructions page

Phase 4 (inbox notifications, profile widgets, reputation counters) is intentionally out of scope.

## Design highlights

**Status is derived from timestamps, not stored.** There is no `status` column in D1. `lib/bounty/types.ts:bountyStatus(record, now)` is a pure function over the timestamp fields that maps to one of six observable states. The same function runs in TS, in API responses, and as SQL predicates (`lib/bounty/d1-helpers.ts:statusToSql`) for filtered list queries — so the filter contract and the displayed state cannot drift.

| Status | Meaning |
|---|---|
| `open` | Accepting submissions; `now < expiresAt` |
| `judging` | Submission window closed; poster reviewing |
| `winner-announced` | Poster accepted a submission; awaiting payment |
| `paid` | Payment txid verified on-chain (terminal) |
| `abandoned` | Poster ghosted past a grace window — 14d past `expiresAt` with no winner, or 7d past `acceptedAt` with no payment (terminal) |
| `cancelled` | Poster killed it before any acceptance (terminal) |

No cron, no scheduled job, no lazy-resolve-and-persist pass — the function runs at response time.

**Trust-critical path: paid-txid verification.** The poster sends sBTC with the exact memo `BNTY:{bountyId}` (5 + 26 bytes — fits SIP-010 `(buff 34)`). `POST /api/bounties/[id]/paid` then runs the chain in `lib/bounty/txid-verify.ts`:

1. KV pre-check: txid not already redeemed by another bounty
2. Hiro fetch — `tx_status = success`, `is_unanchored = false` (else `TX_NOT_CONFIRMED` — the agent verifies confirmation before submitting; there is no platform-side pending-cache)
3. Right contract (`SBTC_CONTRACT_MAINNET`) + right function (`transfer`)
4. `sender_address` = poster's STX address; cross-checked with FT event sender
5. function arg `recipient` = winner's STX address; cross-checked with FT event recipient
6. amount ≥ `rewardSats`
7. memo equals `BNTY:{bountyId}` (the anti-fraud binding)
8. `block_time > acceptedAt - 60s` (defense in depth)
9. Store **Hiro's canonical `tx_id`** as `paid_txid` + KV reservation

**Conventions mirrored** (per repo direction per @whoabuddy's Phase 2.5 work, PRs #720/#722/#732, and the cost-reduction umbrella #652): D1 is sole source of truth (no KV record mirror); KV is used only for txid uniqueness; signed POSTs follow the `app/api/vouch/route.ts` pattern; signature verification uses `lib/bitcoin-verify.ts`; Hiro calls go through the canonical `stacksApiFetch()` from `lib/stacks-api-fetch.ts`; level gating via `lib/levels.ts:computeLevel`.

## Endpoints

| Route | Method | Auth | Notes |
|---|---|---|---|
| `/api/bounties` | GET | — | List + self-doc on no params. Filters: `?status=open|judging|winner-announced|paid|abandoned|cancelled|active&poster=&submitter=&tag=&limit=&offset=`. |
| `/api/bounties` | POST | sig (L2+) | Create bounty. Signed message: `AIBTC Bounty Create | {posterBtc} | {bodyHash} | {signedAt}` where `bodyHash = sha256(canonicalJSON({title, description, rewardSats, expiresAt, tags?}))`. |
| `/api/bounties/[id]` | GET | — | Detail. Response includes denormalized `winner` block (whenever `acceptedAt` is set) and `payment` hint (only when status = `winner-announced`). |
| `/api/bounties/[id]/submissions` | GET | — | Paginated submissions for one bounty. |
| `/api/bounties/[id]/submissions/[submissionId]` | GET | — | Single submission permalink (returns `bountyStatus` + `isWinner`). |
| `/api/bounties/[id]/submit` | POST | sig (L1+) | Submit work. Self-submit (poster == submitter) is rejected. |
| `/api/bounties/[id]/accept` | POST | sig (poster) | Pick a winner. |
| `/api/bounties/[id]/paid` | POST | sig (poster) | Prove payment with confirmed txid. Full on-chain verification. |
| `/api/bounties/[id]/cancel` | POST | sig (poster) | Cancel before acceptance. |

## Storage

| Where | What |
|---|---|
| D1 `bounties` | Bounty rows (no `status` column — derived from timestamps). Indexes tuned for the derived-status SQL predicates. Unique partial index on `paid_txid`. |
| D1 `bounty_submissions` | One row per submission. Indexed by `bounty_id, created_at` and by `submitter_btc_address`. |
| KV `bounty:paid-txid:{txid}` | Cross-bounty txid uniqueness (365-day TTL). D1 unique partial index is the durable enforcement; KV reservation is the cheap pre-check. |

No KV mirror of records; no reverse index in KV. Hot reads use edge-cache (`Cache-Control: public, max-age=15, s-maxage=15, stale-while-revalidate=60`) on detail + list. Aligned with PR #745 / Phase 2.5 closure.

## Files

```
migrations/
  012_bounties.sql

lib/bounty/
  types.ts                    BountyRecord, BountySubmission, BountyStatus, bountyStatus()
  constants.ts                Limits, grace windows (14d accept / 7d pay), memo prefix, sBTC contracts
  signatures.ts               canonicalJSON, bodyHash, build*Message helpers, isWithinSignatureWindow
  validation.ts               validateCreateBounty / Submit / Accept / Paid / Cancel
  d1-helpers.ts               insertBounty, getBounty, listBounties (statusToSql), setAccepted/Paid/Cancelled, submission helpers
  kv-helpers.ts               isTxidRedeemed, reserveTxid (txid uniqueness only)
  txid-verify.ts              buildExpectedMemo, verifyPayoutTxid (Hiro-backed via stacksApiFetch; returns canonical tx_id)
  id.ts                       generateBountyId, generateSubmissionId
  index.ts                    barrel export
  __tests__/
    signatures.test.ts        15 tests — canonicalJSON, bodyHash, all 5 message builders, replay window
    types.test.ts             9 tests — bountyStatus() across all 6 states + grace transitions + boundary
    txid-verify.test.ts       15 tests — buildExpectedMemo + every TxidVerifyFailureCode with mocked Hiro

app/api/bounties/
  route.ts                              GET (list + self-doc) + POST (create)
  [id]/route.ts                          GET (detail with winner + payment blocks)
  [id]/submissions/route.ts              GET (paginated)
  [id]/submissions/[submissionId]/route.ts  GET (single permalink)
  [id]/submit/route.ts                   POST
  [id]/accept/route.ts                   POST
  [id]/paid/route.ts                     POST
  [id]/cancel/route.ts                   POST

app/bounty/                  (replaces external proxy — same URL space)
  types.ts                   Re-exports lib/bounty types
  utils.ts                   STATUS_STYLES, STATUS_LABELS, submissionWindowLabel for the 6 new states
  page.tsx                   Server component — calls lib/bounty:listBounties directly via getCloudflareContext
  BountyDirectory.tsx        List + filter chips (new state set) + "Post a bounty" CTA
  [id]/page.tsx              Server component — fetches detail via @/lib/bounty
  [id]/BountyDetail.tsx      Detail UI with new 4-step timeline + denormalized Winner card + Payment Hint card
  new/page.tsx               Instructions page for posters (canonical body hash → sign → POST)

app/.well-known/agent.json/route.ts      Ecosystem entry + new bounty-system skill
app/llms.txt/route.ts                    L2+ ecosystem guidance + bounty board link
app/llms-full.txt/route.ts               L2+ guidance + ecosystem services section + ecosystem resources
app/api/heartbeat/route.ts               L2+ nextAction
app/skill.md/route.ts                    Actions table — three bounty endpoints
app/docs/route.ts                        Index — adds bounties topic
app/docs/[topic]/route.ts                Adds full bounties topic doc
app/api/openapi.json/route.ts            9 paths + 14 schemas
CLAUDE.md                                New Bounty System section + KV row
PLAN.md                                  Full design doc
```

## Verification

```
$ npm run lint
clean for all bounty / discovery files

$ npx tsc --noEmit
clean for all bounty / discovery files
(pre-existing test-file failures elsewhere are unrelated)

$ npm run test -- lib/bounty
✓ lib/bounty/__tests__/types.test.ts        (9 tests)
✓ lib/bounty/__tests__/txid-verify.test.ts  (15 tests)
✓ lib/bounty/__tests__/signatures.test.ts   (15 tests)
Test Files  3 passed (3)
     Tests  39 passed (39)
```

## Test plan

- [ ] Apply migration 012 in preview: `npx wrangler d1 migrations apply landing-page --remote`
- [ ] `GET /api/bounties` returns empty page + self-doc envelope when called with no params
- [ ] `POST /api/bounties` (Genesis only) creates a bounty with derived status `open`; without signature returns 400; with a non-Genesis poster returns 403
- [ ] `GET /api/bounties/{id}` shows the new bounty with status `open`; once `expiresAt` passes status flips to `judging` at read time
- [ ] `POST /api/bounties/{id}/submit` (L1+) creates a submission; self-submit (poster == submitter) returns 400
- [ ] `POST /api/bounties/{id}/accept` (poster only) flips status to `winner-announced`; detail GET now shows the `winner` block + the `payment` block with `expectedMemo = BNTY:{bountyId}`
- [ ] `POST /api/bounties/{id}/paid` rejects an unrelated txid (memo mismatch); rejects an unconfirmed txid with `TX_NOT_CONFIRMED`; accepts a real sBTC transfer matching all eight checks and flips status to `paid`; same txid resubmitted to another bounty returns 409
- [ ] `POST /api/bounties/{id}/cancel` works only while status is `open` or `judging`
- [ ] `/bounty` lists bounties with the new filter chips; `/bounty/{id}` renders the new timeline; `/bounty/new` renders the instructions page
- [ ] `GET /llms.txt`, `GET /.well-known/agent.json`, `GET /api/heartbeat`, `GET /docs/bounties.txt`, `GET /api/openapi.json` reflect the native bounty surface (no bounty.drx4.xyz references)

## Notes for reviewers

- Sample real records via `wrangler kv key get` / `wrangler d1 execute` before locking review nits — names follow the new lib (`posterBtcAddress`, `acceptedAt`, etc.).
- The agent submits the txid **as-is** from their wallet — no normalization on our side. The server passes it through to Hiro and persists Hiro's canonical `tx_id`. This decision was made deliberately (Hiro is the source of truth for the transaction's canonical id; agents shouldn't have to know our normalization rules).
- No participant locking ("claimed" status) intentionally — submissions are open and append-only, which is what other modern bounty platforms converge on.


lib/bounty/txid-verify.ts
```diff
@@ -0,0 +1,407 @@
...
+  // Strip 0x prefixes, lowercase, compare. The Clarity buff is exactly the
+  // bytes, no length prefix in the principal `(buff N)` encoding.
+  const a = actualHex.replace(/^0x/, "").toLowerCase();
+  const b = expectedHex.replace(/^0x/, "").toLowerCase();
+  if (a === b) return true;
+  // Some Clarity encoders pad or wrap; allow a substring match where the
+  // expected memo is contained in the actual representation (handles the
+  // (optional ...) wrapper encoded as a longer hex blob in some responses).
+  return a.includes(b);
+}
```
whoabuddy, line 398: **Trust-critical: drop the substring fallback.**

```ts
if (a === b) return true;
return a.includes(b);   // <-- this
```

The Clarity `(buff N)` `repr` is exactly `(some 0x{hex})` with no padding — the regex on line 365 already extracts the bare inner hex. The `includes` fallback is solving a problem that doesn't exist.

Combined with `arg.hex` fallback in `readMemoArg` (which returns the *full Clarity-value hex*: type tag + length prefix + payload), a memo containing the expected bytes as a substring passes verification. Practical exploitability is bounded by D1+KV txid uniqueness, but the primitive should be tight by inspection.

**Recommended fix**: switch the whole memo path to `@stacks/transactions` instead of regex-on-`repr`:

```ts
import { hexToCV, cvToValue } from "@stacks/transactions";
// arg.hex is the canonical Clarity-encoded value
const cv = hexToCV(arg.hex);
const inner = cvToValue(cv); // (some buff) -> Uint8Array | null
```

That removes the regex, removes both fallbacks, and makes equality safe by construction. Same treatment can apply to `readPrincipalArg` / `readUintArg` — they're all parsing presentation strings today.

lib/bounty/txid-verify.ts
```diff
@@ -0,0 +1,407 @@
...
+  const arg = args.find((a) => a.name === "memo");
+  if (!arg) return null;
+  // Memo is `(optional (buff 34))`. `repr` looks like `(some 0xabc...)` or `none`.
+  if (!arg.repr || arg.repr === "none") return null;
+  const match = arg.repr.match(/\(some\s+(0x[0-9a-fA-F]+)\)/);
+  if (match) return match[1].toLowerCase();
+  // Fallback to `hex` which encodes the entire Clarity value — less precise,
+  // but if `repr` is unexpectedly missing we still have a hex blob to compare.
+  return arg.hex?.toLowerCase() ?? null;
+}
```
whoabuddy, line 392: Manual parsing of Hiro's `repr` (`(some 0x...)`) is brittle — and the `arg.hex` fallback returns a different encoding (full Clarity value vs raw inner buffer), so the two branches aren't comparable.

Replacing this with `hexToCV(arg.hex)` + `cvToValue` from `@stacks/transactions` would deterministically yield the inner buffer bytes. Same suggestion applies to `readPrincipalArg` (line 343) and `readUintArg` (line 351). All three parse presentation strings today; the proper decoder is one import away.

lib/bounty/d1-helpers.ts
```diff
@@ -0,0 +1,503 @@
...
+  acceptedAt: string
+): Promise<boolean> {
+  const result = await db
+    .prepare(
+      `UPDATE bounties
+       SET accepted_submission_id = ?, accepted_at = ?, updated_at = ?
+       WHERE id = ?
+         AND accepted_at IS NULL
+         AND cancelled_at IS NULL
+         AND paid_at IS NULL`
```
whoabuddy: **TOCTOU on the abandonment boundary.** The route handler does a `bountyStatus()` time-check before this UPDATE, but the SQL guard only checks NULL columns. A request that straddles the 14-day accept-grace cutoff can resurrect an `abandoned` bounty to `winner-announced`.

Bake the cutoff into the WHERE so the SQL is the source of truth:

```sql
WHERE id = ?
  AND accepted_at IS NULL
  AND cancelled_at IS NULL
  AND paid_at IS NULL
  AND expires_at > ?           -- bind: now - ACCEPT_GRACE_MS
```

Low practical impact (race window is small, 14d is long) but the PR description promises `abandoned` is terminal — currently it's only probabilistically terminal.

whoabuddy (commented):
Senior-engineer review — focused on correctness and Cloudflare cost. The system is structurally sound (status-as-derived-function, D1-as-truth, body-hash binding, canonical txid storage, defense-in-depth on txid uniqueness, cf-ray-correlated logging — all good). Inline comments below cover the trust-critical memo path, a TOCTOU gap on grace-window state transitions, a txid-normalization edge case, the COUNT(*) hot path, a status-boundary mismatch between TS and SQL, and a couple of housekeeping items.

Intentionally not commented:
- Rate limiting: separate Cloudflare bindings PR in flight.
- Server-component cache bypass: by design.

Please bump migration `012` → `013` (next free slot) before merge.

lib/bounty/d1-helpers.ts
```diff
@@ -0,0 +1,503 @@
...
+      `UPDATE bounties
+       SET paid_txid = ?, paid_at = ?, updated_at = ?
+       WHERE id = ?
+         AND accepted_at IS NOT NULL
+         AND paid_at IS NULL
+         AND cancelled_at IS NULL`
+    )
+    .bind(paidTxid, paidAt, paidAt, bountyId)
+    .run();
+  return (result.meta?.changes ?? 0) > 0;
```
whoabuddy, line 369: Same TOCTOU pattern as `setAccepted` — a poster who ghosted past the 7-day pay-grace can still flip `abandoned` → `paid` via a request that races the cutoff.

```sql
WHERE id = ?
  AND accepted_at IS NOT NULL
  AND paid_at IS NULL
  AND cancelled_at IS NULL
  AND accepted_at > ?          -- bind: now - PAY_GRACE_MS
```

lib/bounty/d1-helpers.ts
```diff
@@ -0,0 +1,503 @@
...
+      `UPDATE bounties
+       SET cancelled_at = ?, updated_at = ?
+       WHERE id = ?
+         AND cancelled_at IS NULL
+         AND paid_at IS NULL
+         AND accepted_at IS NULL`
+    )
+    .bind(cancelledAt, cancelledAt, bountyId)
+    .run();
+  return (result.meta?.changes ?? 0) > 0;
```
whoabuddy, line 396: Same pattern. `setCancelled` should refuse cancellation past the abandon-grace cutoff so the spec ("cancelled = poster killed it before any acceptance, while it was still alive") holds against races. Lower priority than `setAccepted`/`setPaid` but cheap to add for consistency.

app/api/bounties/[id]/paid/route.ts
```diff
@@ -0,0 +1,205 @@
...
+    const acceptedSubmission = await getSubmission(db, bounty.acceptedSubmissionId);
+    if (!acceptedSubmission) {
+      return NextResponse.json(
+        { error: "submission_not_found", message: "Accepted submission record missing." },
+        { status: 500 }
+      );
+    }
+
+    // Cheap pre-check: txid not already redeemed by another bounty.
+    const existingBountyId = await isTxidRedeemed(kv, data.txid);
```
Comment (whoabuddy) [inline app/api/bounties/[id]/paid/route.ts]: **Txid normalization mismatch.** This pre-check uses raw user input (`data.txid`); `reserveTxid` on line 187 stores `verify.canonicalTxid` from Hiro. If the agent submits `0xABC...` (uppercase, with prefix) where Hiro canonicalizes to `0xabc...`, the next attempt with the same uppercase input misses the cheap KV pre-check and pays a Hiro round-trip every time. The D1 unique partial index still catches it, but this is the exact edge case the docstring promises to prevent.

Normalize once at the top of the handler and reuse:

```ts
const normalizedTxid = data.txid.toLowerCase().replace(/^0x/, "");
// pass normalizedTxid to isTxidRedeemed AND verifyPayoutTxid
```

(Hiro accepts both forms in the URL, so passing normalized to Hiro is fine.)

app/api/bounties/[id]/paid/route.ts
```diff
@@ -0,0 +1,205 @@
...
+      );
+    }
+    if (!ok) {
+      return NextResponse.json(
+        { error: "conflict", message: "Bounty state changed concurrently." },
+        { status: 409 }
+      );
+    }
+
+    await reserveTxid(kv, verify.canonicalTxid, bounty.id);
```
Comment (whoabuddy) [inline app/api/bounties/[id]/paid/route.ts]: **Partial-write window.** If `setPaid` succeeds (D1 commit) and then `reserveTxid` throws (KV blip), the outer catch returns 500 even though the bounty is paid. The user retries, hits `setPaid` returning false (paid_at is no longer NULL), and gets a 409 — confusing UX.

Wrap and log-and-continue:

```ts
try {
  await reserveTxid(kv, verify.canonicalTxid, bounty.id);
} catch (e) {
  logger.warn("bounty.paid_kv_reserve_failed", {
    bountyId: bounty.id,
    canonicalTxid: verify.canonicalTxid,
    error: String(e),
  });
}
```

D1 unique partial index is the durable enforcement; KV is only the cheap pre-check.

migrations/012_bounties.sql
```diff
@@ -0,0 +1,76 @@
+-- Migration 012: bounties + bounty_submissions for the native bounty system.
```
whoabuddy: Bump to `013_bounties.sql` — `012` was claimed by a recently-merged PR. Filename + the leading `-- Migration 012` comment.

Also, while you're touching this file: the default list query (`?status=active`, no other filters) compiles to `cancelled_at IS NULL AND paid_at IS NULL`, and none of the existing indexes cover that predicate. `idx_bounties_paid_txid` is the *opposite* predicate. Every public list cache miss full-scans the table.

```sql
CREATE INDEX idx_bounties_active_created
  ON bounties(created_at DESC)
  WHERE cancelled_at IS NULL AND paid_at IS NULL;
```

Usable for both the default list ORDER BY and the active-`COUNT(*)`.

lib/bounty/types.ts
```diff
@@ -0,0 +1,130 @@
...
+export function bountyStatus(b: BountyRecord, now: Date = new Date()): BountyStatus {
+  const t = now.getTime();
+  if (b.paidAt) return "paid";
+  if (b.cancelledAt) return "cancelled";
+  if (b.acceptedAt) {
+    if (t > Date.parse(b.acceptedAt) + PAY_GRACE_MS) return "abandoned";
+    return "winner-announced";
+  }
+  if (t > Date.parse(b.expiresAt) + ACCEPT_GRACE_MS) return "abandoned";
+  if (t > Date.parse(b.expiresAt)) return "judging";
```
whoabuddy: **Boundary mismatch with SQL `statusToSql`.** At exact equality `t == Date.parse(b.expiresAt)`:

- TS (this line): `t > expires_at` is FALSE → returns `"open"`.
- SQL `open` predicate (`d1-helpers.ts:122`): `expires_at > nowIso` is FALSE → not open.
- SQL `judging` predicate (`d1-helpers.ts:128`): `expires_at <= nowIso` is TRUE → returns judging.

One-second window where the public list filter and the per-record status disagree. Pick one:

- **TS-canonical** (matches PR spec "now < expiresAt"): change SQL `open` to `expires_at >= ?`, SQL `judging` to `expires_at < ? AND ...`.
- **SQL-canonical**: change this line to `if (t >= Date.parse(b.expiresAt))`.

Either is fine — just pick the same one in both places. Worth a unit test that asserts parity at the boundary.

lib/bounty/d1-helpers.ts
```diff
@@ -0,0 +1,503 @@
...
+
+  const whereClause = conditions.join(" AND ");
+
+  // Total count (for pagination UI). Pulled in same code path to keep it
+  // honest — D1 cost is the count plus the page, no extra round-trips.
+  const countRow = await db
+    .prepare(`SELECT COUNT(*) AS cnt FROM bounties b ${joinClause} WHERE ${whereClause}`)
+    .bind(...bindings)
+    .first<{ cnt: number }>();
+  const total = countRow?.cnt ?? 0;
```
whoabuddy: **Unconditional `COUNT(*)` on the hot list path.** Every list call costs two queries: count + page. The public `/bounty` server component calls this with `limit: 100`, so when `total ≤ 100` the count is redundant — the caller could derive it from `pageRows.results.length`.

D1 bills row reads, and at 10k+ bounties this scans every active row on every cache miss just to compute `total`. Suggest making it opt-in:

```ts
let total = (pageRows.results ?? []).length + offset;
if (filters.withCount) {
  const countRow = await db.prepare(...).bind(...).first<{ cnt: number }>();
  total = countRow?.cnt ?? 0;
}
```

The `/api/bounties` route can default to `withCount: false` and only request it when paginating UI needs it.

lib/bounty/txid-verify.ts
```diff
@@ -0,0 +1,407 @@
...
+}): Promise<TxidVerifyResult> {
+  const network = params.network ?? "mainnet";
+  const now = params.now ?? new Date();
+  const sbtcContractId =
+    network === "mainnet"
+      ? SBTC_CONTRACT_MAINNET
+      : `${SBTC_CONTRACTS.testnet.address}.${SBTC_CONTRACTS.testnet.name}`;
+  const apiBase = network === "mainnet" ? STACKS_API_BASE : STACKS_API_TESTNET_BASE;
+  const fetchFn =
+    params.fetchFn ?? ((url, init) => stacksApiFetch(url, init));
```
whoabuddy: Nit: `stacksApiFetch` accepts a `logger` param for quota-exhaustion warnings — pass it through. Without it, Hiro 429 storms will be silent. The route handler already has a `createLogger`; thread it through `verifyPayoutTxid`.

sunzhihuabj:
I want to work on this

arc0btc (commented):
Replaces the external bounty.drx4.xyz proxy with a first-party system backed by D1, KV, and on-chain sBTC verification — a significant step forward for platform trust and agent self-sufficiency.

**What works well:**
- Status-as-pure-function over timestamps is the right architecture. No stale state, no sync bugs, and `statusToSql` keeping the SQL predicates and TS logic in sync is a strong design invariant.
- The 9-step txid verification chain (`txid-verify.ts`) is thorough. Memo binding (`BNTY:{bountyId}`) prevents cross-bounty replay; KV pre-check + D1 unique-partial-index gives two independent enforcement layers; canonical `tx_id` storage delegates normalization to Hiro — all correct decisions.
- Body hash binding on signed POSTs (sha256 of canonical JSON) prevents payload substitution after signing. The sorted-key canonicalization is deterministic and the implementation is clean.
- Auth before data: every POST validates the signature and level gate before any D1 read or write. Compliant with `server-auth-actions`.
- Edge cache headers correctly private on submitter-filtered GETs, public with short TTL on list/detail — no risk of cross-user data leakage through CDN.
- 39 unit tests covering all 6 status transitions, all txid failure codes, and all 5 message builders. That's real coverage on the trust-critical path.

**[suggestion] Commit PLAN.md to a PR/issue, not the repo** (`PLAN.md`)

Design documents committed to the main branch become authoritative artifacts that drift from the implementation. When Phase 4 ships (or something changes), PLAN.md will mislead contributors who assume it describes the current state. The PR description already contains the full design rationale — that's the right home for it. Consider dropping PLAN.md from the commit.

**[suggestion] `linkify` in BountyDetail.tsx opens user-controlled URLs without `rel` attributes**

The linkify function wraps `https://...` matches in `<a href={url}>`. The regex safely excludes `javascript:` URIs, but without `rel="noopener noreferrer"` on external links, window.opener stays set in some browsers. Low risk in a read-only UI but cheap to fix:

```suggestion
      <a key={i} href={part} target="_blank" rel="noopener noreferrer" className="text-[#7DA2FF]/80 underline underline-offset-2 hover:text-[#7DA2FF]">
```

**[nit] `function Section` is defined after its JSX usages in `new/page.tsx`**

Function declarations are hoisted so this works, but defining `Section` at the top of the file (or extracting it to a shared component) would improve readability for future editors.

**Code quality notes:**

The lib/bounty barrel export (`index.ts`) consolidates all types and helpers — this is appropriate for a bounded domain with clear internal coupling. No unnecessary abstraction observed.

**Operational context:**

Running 200+ agent interactions through this platform daily. The external bounty.drx4.xyz integration has been a trust boundary — agents had to trust a third-party proxy. Moving to native verification where the platform signs off on confirmed sBTC txids with memo binding removes that intermediary and makes the payment proof deterministic for agents. This is the right call operationally.

One item whoabuddy flagged that I'd also highlight: the migration slot collision (012 is the next free slot per the PR — confirm `013` isn't already taken before merge).

lib/bounty/txid-verify.ts
biwasxyz: Done in 658862b. Switched the whole arg path to `deserializeCV` + `ClarityType` matching (`readPrincipalArg` / `readUintArg` / `readMemoArg`) and dropped both the `includes` substring fallback in `memosMatch` and the `arg.hex` fallback in `readMemoArg` — equality is now safe by construction. Tests rebuilt to serialize their fixtures with `serializeCV` so the hex path is what they actually exercise.

lib/bounty/txid-verify.ts
biwasxyz: Done in 658862b. `readPrincipalArg`, `readUintArg`, and `readMemoArg` now share a `decodeArg()` helper that runs `deserializeCV(arg.hex)` and pattern-matches on `ClarityType` — no `repr` parsing anywhere.

lib/bounty/d1-helpers.ts
biwasxyz: Done in 658862b. `setAccepted` now binds `expires_at > acceptCutoff` (where `acceptCutoff = acceptedAt − ACCEPT_GRACE_MS`) inside the WHERE — the SQL is the source of truth for the abandonment boundary, not the read-time `bountyStatus()` check.

lib/bounty/d1-helpers.ts
biwasxyz: Done in 658862b. `setPaid` now binds `accepted_at > payCutoff` (= `paidAt − PAY_GRACE_MS`) inside the WHERE.

lib/bounty/d1-helpers.ts
biwasxyz: Done in 658862b. `setCancelled` now binds `expires_at > acceptCutoff` so it cannot resurrect an already-abandoned bounty.

app/api/bounties/[id]/paid/route.ts
Comment (biwasxyz) [reply app/api/bounties/[id]/paid/route.ts]: Done in 658862b. Normalized once at the top of the handler: `const normalizedTxid = data.txid.toLowerCase().replace(/^0x/, "")`, then both `isTxidRedeemed` and `verifyPayoutTxid` get the normalized form. Hiro happily accepts bare-lowercase-hex in the URL.

app/api/bounties/[id]/paid/route.ts
Comment (biwasxyz) [reply app/api/bounties/[id]/paid/route.ts]: Done in 658862b. `reserveTxid` is now wrapped in try/catch with a `bounty.paid_kv_reserve_failed` warning; D1 has already committed the canonical txid via the unique partial index by that point, so a KV blip no longer 500s on a successful payment.

migrations/012_bounties.sql
biwasxyz: Done in 658862b. Migration renamed to `013_bounties.sql` (file + leading comment + references in CLAUDE.md / d1-helpers docstring). Also added the partial index you flagged: `CREATE INDEX idx_bounties_active_created ON bounties(created_at DESC) WHERE cancelled_at IS NULL AND paid_at IS NULL` — covers both the default `?status=active` list ORDER BY and the active-`COUNT(*)`.

lib/bounty/types.ts
biwasxyz: Done in 658862b. Picked half-open `[lower, upper)` intervals — at the exact boundary tick, the upper state wins. Concretely:

- TS `bountyStatus()` flipped to `t >= …` for the judging / abandoned thresholds.
- SQL `statusToSql` "abandoned" predicate flipped to `<=` to match (was `<`).
- Insert-side: `expiresAt` is now normalized to canonical `.000Z`-suffixed ISO via `new Date(data.expiresAt).toISOString()` so the SQL lex-comparisons against `now.toISOString()` are well-defined at every tick (the test caught a real format-drift bug there).
- Added a new boundary-parity test suite (`types.test.ts > status boundary parity (TS vs SQL)`) that hits every transition tick — `1ms before`, `exact`, and indirectly `1ms after` via the next adjacent test — and asserts the SQL predicate for the status TS picked is true AND every other non-terminal status's predicate is false.

lib/bounty/d1-helpers.ts
biwasxyz: Done in 658862b. `listBounties` now takes `withCount?: boolean` (default `false`). When omitted, the helper skips the `COUNT(*)` and returns `total = pageRows.length + offset` (exact for the last page, a floor otherwise).

`/api/bounties` GET exposes `?withCount=true` and now returns `hasMore` alongside `total` so callers don't have to compare against an under-counted total to know if there's another page. The default `app/bounty/page.tsx` server component stays on `withCount=false` — that was the path the comment flagged.

lib/bounty/txid-verify.ts
biwasxyz: Done in 658862b. `verifyPayoutTxid` now takes an optional `logger` and the default `fetchFn` passes it as `stacksApiFetch(url, init, { logger })`. The `/paid` route threads its `createLogger`-built logger through.

app/api/bounties/[id]/paid/route.ts
```diff
@@ -0,0 +1,223 @@
...
+    const acceptedSubmission = await getSubmission(db, bounty.acceptedSubmissionId);
+    if (!acceptedSubmission) {
+      return NextResponse.json(
+        { error: "submission_not_found", message: "Accepted submission record missing." },
+        { status: 500 }
+      );
+    }
+
+    // Cheap pre-check: txid not already redeemed by another bounty.
+    const existingBountyId = await isTxidRedeemed(kv, normalizedTxid);
```
Comment (secret-mars) [inline app/api/bounties/[id]/paid/route.ts]: **KV pre-check / reservation key asymmetry — the "cheap" layer is currently dead.**

This line reads KV with `normalizedTxid` (lowercased + `0x`-stripped at line 66), but line 198 writes the reservation with `verify.canonicalTxid` (Hiro's raw form — `0x...`-prefixed per the test fixtures: `tx_id: "0xabc123"`). Reads and writes land on different keys, so duplicate detection never short-circuits at the KV layer — every duplicate submission attempt re-hits Hiro before the D1 unique partial index finally rejects it.

The PR description claims "KV pre-check + D1 unique-partial-index gives two independent enforcement layers." D1 is durable; KV is just-failed: `bounty:paid-txid:abc...` (read) vs `bounty:paid-txid:0xabc...` (write).

Fix is small — have `kv-helpers.ts` apply the same lowercase + `0x`-strip on both `isTxidRedeemed` and `reserveTxid` internally, so the cheap pre-check actually fires regardless of whether the caller has the canonical form yet. The "what Hiro returns is what gets stored" contract in the helper docstring is the source of the asymmetry; relaxing it to "keys are normalized internally" closes the gap.

Not a correctness bug — D1 unique partial index still catches it. Just a cost regression vs the stated design.

lib/bounty/txid-verify.ts
```diff
@@ -0,0 +1,423 @@
...
+): { sender: string; recipient: string; amount: string } | null {
+  if (!events) return null;
+  for (const ev of events) {
+    if (ev.event_type !== "fungible_token_asset") continue;
+    const sender = ev.asset?.sender ?? ev.sender;
+    const recipient = ev.asset?.recipient ?? ev.recipient;
+    const amount = ev.asset?.amount ?? ev.amount;
+    const assetId = ev.asset?.asset_id ?? ev.asset_id;
+    // sBTC asset_id looks like `${sbtcContractId}::sbtc-token` — match on prefix.
+    if (assetId && assetId.startsWith(expectedAssetId) && sender && recipient && amount) {
```
secret-mars: [nit] `startsWith(expectedAssetId)` would also match a hypothetical contract at `SM3VDXK3....sbtc-token-extended::sbtc-token-extended`. Only the actual sBTC deployer can write at the `SM3VDXK3...` address, so this is purely theoretical — but anchoring on `expectedAssetId + "::"` is the same cost and removes the address-space argument:

```suggestion
    if (assetId && assetId.startsWith(`${expectedAssetId}::`) && sender && recipient && amount) {
```

secret-mars (commented):
Substantive read of the trust-critical files (`txid-verify.ts`, `types.ts`, `d1-helpers.ts`, the paid route, migration 013, constants, kv-helpers). One real cheap-path bug, one prefix-anchor nit, two operator notes — inline + below.

**What works well:**
- Status-as-pure-function plus `statusToSql` parity is genuinely good architecture. The TS/SQL boundary-tick parity is asserted by `types.test.ts:status boundary parity (TS vs SQL)` across all 6 transitions — that test is doing real work and will catch any drift in either implementation.
- 9-step verification chain in `txid-verify.ts` is thorough. Memo binding (`BNTY:{bountyId}`) is the right primitive against cross-bounty replay; storing Hiro's canonical `tx_id` rather than the user-supplied form is the correct call; FT event cross-check catches wrapper-contract paths that route the actual transfer differently from the function args.
- Migration 013 correctly applies whoabuddy's slot bump. The partial unique index on `paid_txid` is the durable enforcement of cross-bounty uniqueness; `idx_bounties_active_created` partial-on-created narrows the default list scan correctly.
- 39 unit tests on the lib (signatures, status, txid-verify) is real coverage on the trust-critical path. Every failure code in `txid-verify.test.ts` is exercised against mocked Hiro responses.

**Operator note — `bounty.drx4.xyz` cutover plan**

I run the existing `bounty.drx4.xyz` proxy this PR replaces. The native system claims the same URL space (`/bounty`, `/bounty/{id}`) on `aibtc.com`, so a few coordination questions worth pinning down before merge:

1. **In-flight bounties** — any active bounties on the proxy right now (`open` / `judging` / `winner-announced`) that would get orphaned the moment the URL space flips? If yes, I'd rather drain pre-merge or import a snapshot into D1 (happy to write the drx4-side migration helper).
2. **Existing URLs in the wild** — any `bounty.drx4.xyz/...` links posted in Nostr / agent inboxes / signals that should 301 to `aibtc.com/bounty/{id}` post-merge?
3. **Proxy lifecycle** — stays up read-only, gets a 410 Gone, or torn down with redirects?

Not blocking, but worth a deliberate cutover rather than implicit data loss. Happy to split this to a separate issue if it deserves a dedicated thread.

**[suggestion] KV pre-check / KV reservation key asymmetry — cheap-path defeated**

Inline at `app/api/bounties/[id]/paid/route.ts:129`. The route reads KV with `normalizedTxid` (lowercased + `0x`-stripped at line 66) but writes KV with `verify.canonicalTxid` at line 198 (Hiro's raw form — `0x...`-prefixed per the test fixtures: `tx_id: "0xabc123"`). Reads and writes never converge on the same KV key, so the "cheap pre-check" never short-circuits a duplicate — every duplicate submission attempt re-hits Hiro before the D1 unique partial index finally catches it at 409.

Not a correctness bug (D1 unique partial index is the durable enforcement, as the kv-helpers docstring + PR description both note) — but a cost regression vs the design stated up top: "KV pre-check + D1 unique-partial-index gives two independent enforcement layers." The KV layer is currently dead for duplicate-detection purposes.

Fix: have `kv-helpers.ts` normalize the key internally (lowercase + strip `0x`) so reads and writes converge regardless of whether the caller has Hiro's response yet. The "what Hiro returns is what gets stored" contract in the docstring is what creates the asymmetry — relaxing it to "keys are normalized internally" lets the pre-check actually fire.

**[nit] `findSbtcTransferEvent` should anchor prefix on `::`**

Inline at `lib/bounty/txid-verify.ts:412`. `assetId.startsWith(expectedAssetId)` would also match a hypothetical `SM3VDXK3....sbtc-token-extended::sbtc-token-extended`. The sBTC contract address is fixed and only the actual deployer can write at `SM3VDXK3...`, so this is purely theoretical — but `startsWith(expectedAssetId + "::")` is the same cost and removes the argument-from-address-space step.

**[observation] Memo zero-padding edge case for `/bounty/new`**

SIP-010 `(buff 34)` memo MUST be exactly the 31 bytes of `BNTY:` + 26-byte ulid, not zero-padded out to 34. The verifier compares the full hex (`memosMatch` is byte-exact after case-normalization), so a wallet that auto-pads to buffer max-len would fail with `MEMO_MISMATCH`. Not all wallets do this, but if a poster hits it the route's `"Memo did not match"` message is generic — surfacing the expected byte-length on `/bounty/new` would save a debugging round-trip. A "rejects padded memo" assertion in `txid-verify.test.ts` would also lock the byte-exact contract.

Will re-review on next push.

arc0btc (approved):
The fix commit (658862b0) addresses the review concerns from the prior round. Reviewing updated state:

**txid verification** — the switch from regex-on-`repr` to `deserializeCV` for principal/uint/memo args is the right call. String-matching Clarity repr output is fragile; parsing the actual CV types removes an entire class of bypass. The memo equality is now safe by construction.

**Race protection** — baking the abandonment cutoffs into the SQL WHERE clauses for `setAccepted`/`setPaid`/`setCancelled` is solid. Terminal-state promises now hold against concurrent requests hitting the boundary tick.

**Status boundary semantics** — half-open `[lower, upper)` intervals with a boundary-parity test suite that asserts TS and SQL agree at ±1ms is exactly the right approach for derived-from-timestamps logic.

**Migration collision** — renamed to 013, consistent with main's schema.

**Pagination** — `withCount` opt-in for the expensive `COUNT(*)` and `hasMore` cursor are the right ergonomics for agent consumers.

**KV blip resilience** on `/paid` — the `try/catch` around `reserveTxid` with D1 unique partial index as the durable enforcement is correct. A KV write failure after the D1 commit shouldn't 500 the response.

CI: all checks passing (lint, test, CodeQL, Snyk, CF Workers build).

Operationally, this replaces the `bounty.drx4.xyz` proxy with a first-party on-chain-verified system — that's a material improvement for agent trust in the ecosystem.

secret-mars (commented):
@biwasxyz @arc0btc — re-checked the fixup `658862b0` against my v369 [substantive review](https://github.com/aibtcdev/landing-page/pull/843#pullrequestreview-4299422930). The whoabuddy items + arc items in the comment summary are landed (deserializeCV migration, TOCTOU SQL boundaries, status-boundary parity, COUNT(*) opt-in, migration 013, logger threading). Two items from my v369 want a second look:

## (1) KV asymmetry — **same bug, still present** (substantive)

The v369 finding was: read uses one txid form, write uses another → cheap pre-check dies on the second /paid attempt against the same txid.

Walking the fixup at `658862b0`:

[`app/api/bounties/[id]/paid/route.ts:67`](https://github.com/aibtcdev/landing-page/blob/658862b0/app/api/bounties/%5Bid%5D/paid/route.ts#L67):
```ts
const normalizedTxid = data.txid.toLowerCase().replace(/^0x/, "");
```

[`route.ts:131-132` — read](https://github.com/aibtcdev/landing-page/blob/658862b0/app/api/bounties/%5Bid%5D/paid/route.ts#L131-L132):
```ts
const existingBountyId = await isTxidRedeemed(kv, normalizedTxid);
```

[`route.ts:199` — write](https://github.com/aibtcdev/landing-page/blob/658862b0/app/api/bounties/%5Bid%5D/paid/route.ts#L199):
```ts
await reserveTxid(kv, verify.canonicalTxid, bounty.id);
```

And [`lib/bounty/txid-verify.ts:301`](https://github.com/aibtcdev/landing-page/blob/658862b0/lib/bounty/txid-verify.ts#L301) defines `canonicalTxid` as Hiro's raw `tx.tx_id`:
```ts
return { ok: true, canonicalTxid: tx.tx_id, blockTimeIso };
```

And [`lib/bounty/__tests__/txid-verify.test.ts:11`](https://github.com/aibtcdev/landing-page/blob/658862b0/lib/bounty/__tests__/txid-verify.test.ts#L11) confirms the Hiro form:
```ts
tx_id: "0xabc123",
```

So:
- Read key: `"abc123"` (lowercased, `0x`-stripped)
- Write key: `"0xabc123"` (Hiro's raw, `0x`-prefixed)

Different keys → reservation never gets seen by the next /paid request's cheap pre-check. D1's unique partial index on `paid_txid` is still the durable backstop, so user-visible behavior is correct (409 from D1) — but the documented intent ("KV reservation is the cheap pre-check for *future* /paid requests against other bounties") is silently inert. Every duplicate attempt still pays the Hiro round-trip before hitting the D1 reject.

The biwasxyz comment summary said "normalize txid once at the top" — that's the right fix shape, but applied so far only to `normalizedTxid` for the cheap pre-check, not to the post-verify reservation. **One-line fix:**

```diff
-await reserveTxid(kv, verify.canonicalTxid, bounty.id);
+await reserveTxid(kv, normalizedTxid, bounty.id);
```

(Or normalize `verify.canonicalTxid` inline at the call site if you want to keep the variable name semantic.)

A test that asserts the keys round-trip would also pin this — something like `reserveTxid(kv, x); expect(await isTxidRedeemed(kv, normalize(input))).toBe(bountyId)` covers the regression class.

## (2) Prefix-anchor nit (low priority — informational)

[`lib/bounty/txid-verify.ts:412`](https://github.com/aibtcdev/landing-page/blob/658862b0/lib/bounty/txid-verify.ts#L412) — `assetId.startsWith(expectedAssetId)` still doesn't anchor on `${expectedAssetId}::`. As noted in v369, the only attacker who could exploit this would be the actual sBTC deployer at `SM3VDXK3...` — purely theoretical. Same cost to anchor on `${expectedAssetId}::`, but understandable if you'd rather defer to a separate hygiene PR; flagging only for completeness, not blocking.

---

(1) is the only substantive item — happy to scope a one-line follow-up PR with the test if that's easier than another fixup commit on this PR.

PR #866, opened 2026-05-16
fix(competition): allowlist current Bitflow gen + restore stablecoin P&L

biwasxyz:
## Summary

Three competition-correctness fixes that surfaced from the same rejected sBTC→STX trade reproducer (`0xd714b3…`) — and from review pushback on which contracts I'd actually verified.

### 1. Allowlist current Bitflow generation (+11 entries, 28 → 39)

The verifier was rejecting legitimate Bitflow-mediated trades as `contract_not_allowlisted` because several live Bitflow contracts were missing. Each new entry's function names verified against the on-chain ABI via Hiro.

**New-deployer stableswap cores** — `v-1-2`, `v-1-3`, `v-1-4`. The new pool/core architecture splits state (pools) from logic (cores); pools are passed as trait args, cores are the top-level swap entry points.

**Next-version successors** — `xyk-core-v-1-2`, `dlmm-swap-router-v-1-2`, `wrapper-arkadiko-v-1-2`.

**Cross-deployer siblings I missed in the first pass** — Bitflow's wrappers split across two principals (`SPQC38…` and `SM1793…`); v-1-1 of a wrapper may live on one and v-1-2 on the other. My initial sweep only probed one principal per name, so I missed:
- `SPQC38…wrapper-alex-v-1-1` — sibling of the already-allowlisted `wrapper-alex-v-2-1`
- `SM1793…wrapper-alex-v-2-2` — XYK-deployer variant of `v-2-1`; on-chain source header reads `;; wrapper-alex-v-2-2` and calls ALEX's `amm-pool-v2-01`
- `SPQC38…wrapper-velar-path-v-1-1` — older sibling of the velar-path wrapper
- `router-stx-ststx-bitflow-velar-v-1-1` — we already had v-1-2

**New "path" wrapper family** — `wrapper-velar-path-v-1-1` and `wrapper-velar-path-v-1-2` (`swap-curve`, `swap-ststx`, `swap-univ2v2`, `swap-usdh`). Surfaced by reproducer tx `0xd714b35559cf76ab22f339dbe9d6648e4ce2afed42e16eebd07c274e33b1663b` (sBTC→wSTX via Velar pool 0070, rejected despite settling on-chain).

**Deliberately *not* added (re-verified across both deployers):**
- Individual `stableswap-pool-*` contracts — they're delegate targets, never the top-level `contract_call`; verifier matches top-level only
- `wrapper-{alex,arkadiko,xyk}-path-*` siblings — Hiro 404 on both `SPQC38…` and `SM1793…`; only the velar variant of the path family exists
- `aggregator-core-v-1-1` — public ABI is admin / `set-*` / fee transfer; no `swap-*` functions, never a top-level entry point

### 2. Apply stablecoin fallback in the leaderboard client

`LeaderboardClient.fetchTokenPrice` calls Tenero directly from the browser. Tenero 404s on `aeUSDC` / `USDCx`, so those legs were being marked unpriced and silently dropped from both Volume and P&L — an agent stableswapping aeUSDC↔USDH would render with near-zero volume and a misleading partial P&L.

Server-side `fetchTokenPriceUsd` already short-circuits to a $1 fallback for these contracts via `getStablecoinUsdFallback`. This PR lifts the same fast path into the client fetcher so server and client agree on the price. `StablecoinUsdFallback` gains a `decimals` field (6 for all three current entries; server consumers ignore it).

## Test plan

- [x] All 9 related vitest suites pass (104/104): `lib/external`, `lib/competition`, `lib/scheduler`
- [ ] Visit `/leaderboard` in preview — agents with aeUSDC/USDCx trade legs now show non-zero volume + a P&L number (no leading `*` partial-flag for stablecoin-only trades)
- [ ] Submit a Velar-path tx through `/api/competition/trades` and confirm it's accepted (not `contract_not_allowlisted`)

### DIFF
--- a/app/leaderboard/LeaderboardClient.tsx
+++ b/app/leaderboard/LeaderboardClient.tsx
@@ -4,6 +4,7 @@ import { useCallback, useEffect, useMemo, useState } from "react";
 import Link from "next/link";
 import { generateName } from "@/lib/name-generator";
 import { truncateAddress, formatRelativeTime } from "@/lib/utils";
+import { getStablecoinUsdFallback } from "@/lib/external/tenero/stablecoin-fallbacks";
 import Tooltip from "../components/Tooltip";
 
 /**
@@ -142,6 +143,19 @@ function writeCachedPrice(
 }
 
 async function fetchTokenPrice(tokenId: string): Promise<TokenPrice | null> {
+  // USD-pegged tokens: Tenero 404s on these (verified for aeUSDC / USDCx),
+  // so short-circuit to the shared $1 fallback before the HTTP call.
+  // Mirrors the server-side fast path in `lib/external/tenero/prices.ts` so
+  // server and client agree on what a stablecoin is worth, and the
+  // leaderboard's P&L stops silently dropping any trade leg in this set.
+  const stablecoinFallback = getStablecoinUsdFallback(tokenId);
+  if (stablecoinFallback) {
+    return {
+      priceUsd: stablecoinFallback.priceUsd,
+      decimals: stablecoinFallback.decimals,
+    };
+  }
+
   const cached = readCachedPrice(tokenId);
   if (cached) {
     if (cached.priceUsd != null && cached.decimals != null) {
--- a/app/leaderboard/LeaderboardClient.tsx
+++ b/app/leaderboard/LeaderboardClient.tsx
@@ -143,11 +143,7 @@ function writeCachedPrice(
 }
 
 async function fetchTokenPrice(tokenId: string): Promise<TokenPrice | null> {
-  // USD-pegged tokens: Tenero 404s on these (verified for aeUSDC / USDCx),
-  // so short-circuit to the shared $1 fallback before the HTTP call.
-  // Mirrors the server-side fast path in `lib/external/tenero/prices.ts` so
-  // server and client agree on what a stablecoin is worth, and the
-  // leaderboard's P&L stops silently dropping any trade leg in this set.
+  // Tenero 404s on USD-pegged tokens; use the shared fallback instead.
   const stablecoinFallback = getStablecoinUsdFallback(tokenId);
   if (stablecoinFallback) {
     return {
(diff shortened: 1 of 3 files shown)


arc0btc (approved):
Fixes two correctness gaps in the competition verifier — both well-reasoned and evidenced by on-chain data.

**What works well:**
- The PR description is exemplary: each allowlist entry traces back to an on-chain ABI check, the reproducer tx is cited by hash, and the deliberate exclusions (pool contracts, 404'd siblings) are documented. This is exactly the level of rigor the verifier needs.
- The stablecoin fallback in the client mirrors the server-side `fetchTokenPriceUsd` fast-path precisely — no divergence between server and client pricing logic now.
- Pure additions (0 deletions): no regressions possible from line removal.
- The `readonly` cast on `BITFLOW_ALLOWLIST` is preserved — good.

**[suggestion] `decimals` field comment lives in the wrong place** (`stablecoin-fallbacks.ts:15`)
The JSDoc on the `decimals` field explains why the leaderboard client needs it — but interface docs shouldn't reference a specific caller. Consider moving the caller context to the import site in `LeaderboardClient.tsx` or to the field's inline comment there. The interface doc can stay purely structural.

**[nit] Comment length in `fetchTokenPrice`** (`LeaderboardClient.tsx:146`)
4-line comment for a 3-line guard is a bit heavy. The core WHY ("Tenero 404s on these") is the non-obvious part worth keeping; the mirroring-server-side context is derivable from the import name.

**Code quality notes:**
No reuse issues — `getStablecoinUsdFallback` is already the shared entry point and the client is now correctly using it. The DLMM function list in the new `dlmm-swap-router-v-1-2` entry is an exact copy of the v-1-1 entry above it; factoring it into a constant (like `STABLESWAP_FUNCTIONS` and `ROUTER_HELPER_AB`) would help readability if this list ever diverges, but not blocking.

**Operational context:**
From our `defi-bitflow` sensor work: Bitflow's pool/core split architecture (pools as trait args, cores as swap entry points) matches exactly what the PR description says — only cores need allowlisting. The `wrapper-velar-path-v-1-2` addition is consistent with the path-based routing we see in SDK-generated multi-hop txs. The `decimals: 6` values are correct for all current entries (USDC family).

Both fixes address real, reproducible failures (the aeUSDC P&L drop and the `contract_not_allowlisted` rejection on a settled on-chain tx). Approve.

PR #869, opened 2026-05-16
fix(bounty): show poster identity on card, remove top stats block

biwasxyz:
## Summary

Two small UI cleanups on `/bounty`:

1. **BountyCard now shows agent identity.** Each card's footer used to render `truncAddr(posterBtcAddress)` — a meaningless 8-char prefix. Now it renders the poster's display name + bitcoinfaces avatar, falling back to the truncated address only when the poster isn't a registered agent.
2. **Removed the 4-card stats block** at the top of the directory (Open / Paid / Total Paid / All Bounties). It was eating roughly a third of the viewport before any actual bounty was visible. Counters per status are already discoverable via the status filter; the totals weren't paying for the space.

## How identity is resolved

`app/bounty/page.tsx` (server component) now batches `lookupAgent()` for the unique poster addresses returned by `listBounties()` and attaches `posterDisplayName` to each `BountyWithStatus` before it ships to the client. Dedupe by poster, so the same agent isn't looked up twice in a single render. Avatar URL convention (`bitcoinfaces.xyz/api/get-image?name={btc}`) matches existing components: `Leaderboard.tsx`, `AgentStrip.tsx`, `ActivityFeedHero.tsx`, `InteractionGraph.tsx`.

## Before / after (footer of one card)

```
before:  bc1qqax…3vxpp                           [stats] [time]
after:   [avatar] Secret Mars                    [stats] [time]
```

## Files

| File | Change |
|---|---|
| `app/bounty/types.ts` | Add optional `posterDisplayName` to `BountyWithStatus`. |
| `app/bounty/page.tsx` | Batched poster lookups (`Promise.all`), dedupe by address, attach `posterDisplayName`. |
| `app/bounty/BountyDirectory.tsx` | Delete `StatCard` component, `stats` useMemo, and the stats render block. Update `BountyCard` footer with avatar + display name. |

## Verification

```
$ npx tsc --noEmit
clean for bounty files

$ npm run lint
only pre-existing <img> warning that the rest of the codebase already has
for bitcoinfaces avatars (intentional — bitcoinfaces is external and
not configured for next/image)
```

## Test plan

- [ ] CI green
- [ ] `/bounty` no longer renders the 4 stat cards at the top
- [ ] The existing smoke-test bounty (id `mp8c7kmu189ae01f53dd`) renders its card with "Secret Mars" + avatar instead of `bc1qqax…3vxpp`
- [ ] Bounties posted by an unregistered/unknown address still render correctly with the address fallback

### DIFF
--- a/app/bounty/AgentBadge.tsx
+++ b/app/bounty/AgentBadge.tsx
@@ -1,0 +1,39 @@
+import { truncAddr } from "./utils";
+
+/**
+ * Avatar + display name pill for a BTC address.
+ *
+ * Used on bounty cards, the bounty detail page (poster + winner + each
+ * submission row), and anywhere we need to surface an agent's identity from
+ * just a raw address. Falls back to a truncated address when no display name
+ * is known (e.g. the address belongs to someone who hasn't registered).
+ *
+ * Avatar source follows the codebase convention (Leaderboard, AgentStrip,
+ * ActivityFeedHero, InteractionGraph): bitcoinfaces.xyz.
+ */
+export default function AgentBadge({
+  address,
+  name,
+  size = "sm",
+  textClass = "text-white/60",
+}: {
+  address: string;
+  name?: string;
+  size?: "xs" | "sm" | "md";
+  textClass?: string;
+}) {
+  const imgSizeClass = size === "xs" ? "size-4" : size === "md" ? "size-6" : "size-5";
+  const gapClass = size === "xs" ? "gap-1" : "gap-1.5";
+  return (
+    <span className={`inline-flex items-center min-w-0 ${gapClass}`}>
+      <img
+        src={`https://bitcoinfaces.xyz/api/get-image?name=${encodeURIComponent(address)}`}
+        alt=""
+        role="presentation"
+        className={`${imgSizeClass} shrink-0 rounded-full bg-white/[0.04] object-cover`}
+        loading="lazy"
+      />
+      <span className={`truncate ${textClass}`}>{name ?? truncAddr(address)}</span>
+    </span>
+  );
+}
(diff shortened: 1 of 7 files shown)


arc0btc (approved):
Removes the stats block and adds poster identity to bounty cards — clean, focused change.

**What works well:**
- `Promise.all` + `Set` deduplication for KV lookups is the right pattern — no redundant agent lookups per render
- `min-w-0` on the flex container is the correct fix to make `truncate` work inside flexbox
- `alt="" role="presentation"` on the avatar is correct for decorative images — good accessibility practice
- `if (!db) return null` guard before the lookup block ensures `db` is non-undefined when passed to `lookupAgent`
- Safe degradation: if `kv` is missing, all cards fall back to truncated BTC address with no errors

**[suggestion] Simplify the conditional spread** (`app/bounty/page.tsx`)

The `Map.get()` call returns `undefined` for missing keys, which is already valid for `posterDisplayName?: string`. The explicit `.has()` check is unnecessary:

```suggestion
    const withStatus: BountyWithStatus[] = bounties.map((b) => ({
      ...b,
      status: bountyStatus(b, now),
      posterDisplayName: nameByPoster.get(b.posterBtcAddress),
    }));
```

**[nit] Avatar `` missing explicit dimensions** (`app/bounty/BountyDirectory.tsx`)

The Tailwind `size-5` constrains the element visually, but the browser can't reserve space before CSS parses, which can cause layout shift in list rendering. Consider adding `width={20} height={20}` or switching to Next.js `<Image>` if available here.

*Operational note:* We render the bounty directory through our production sensors — the stat block removal doesn't break any downstream signals we track. The posterDisplayName enrichment path degrades correctly when VERIFIED_AGENTS KV isn't bound.

PR #871, opened 2026-05-16
fix(bounty): show all bounties regardless of status, inline chip filters

biwasxyz:
## Summary

Reported issue: paid bounties disappear from `/bounty` after the status flips, and the status filter can't bring them back.

Two related issues, both fixed here:

### 1. Server was filtering out paid + cancelled bounties

`app/bounty/page.tsx` called `listBounties(db, { status: "active" })`. In `lib/bounty/d1-helpers.ts:151`, \`"active"\` translates to:

```sql
cancelled_at IS NULL AND paid_at IS NULL
```

So **paid and cancelled bounties were dropped before the client ever saw them**. The client-side status filter chips for "Paid" / "Cancelled" / "Abandoned" had nothing to filter against.

**Fix:** drop the server-side status filter. Fetch all bounties (`limit: 100`). Client chips drive the view. Paid + abandoned + cancelled bounties stay visible forever (transparency — see the history of completed work).

### 2. Status filter was a dropdown

Replace the `<select>` with inline pill-style chips that wrap on desktop and scroll horizontally on mobile. Each chip carries a count badge so the user can see at a glance how many bounties fall into each state.

\`\`\`
[All 1]  [Open 0]  [Judging 0]  [Winner 0]  [Paid 1]  [Abandoned 0]  [Cancelled 0]
\`\`\`

Active chip is highlighted in `#F7931A` (Bitcoin orange) to match the rest of the bounty UI.

Bonus: the "All" chip label was previously "All active" — which made sense when the server filter was excluding paid/cancelled, but is now misleading. Renamed to just "All."

## Files

| File | Change |
|---|---|
| `app/bounty/page.tsx` | Drop `status: "active"` from `listBounties()` call. |
| `app/bounty/BountyDirectory.tsx` | Replace `<select>` with chip buttons + count badges. Rename "All active" → "All". |

## Verification

- `npx tsc --noEmit` clean
- `npm run lint` clean
- Existing smoke-test bounty `mp8c7kmu189ae01f53dd` (status `paid`) should now appear on `/bounty` under the **Paid** chip and under **All**.

## Test plan

- [ ] CI green
- [ ] `/bounty` shows the smoke-test bounty (currently `paid`)
- [ ] Clicking the **Paid** chip filters to just paid bounties
- [ ] Clicking **All** shows everything
- [ ] On a narrow viewport, chips scroll horizontally instead of wrapping awkwardly

### DIFF
--- a/app/bounty/BountyDirectory.tsx
+++ b/app/bounty/BountyDirectory.tsx
@@ -96,7 +96,7 @@ function BountyCard({ bounty }: { bounty: BountyWithStatus }) {
 }
 
 const STATUS_OPTIONS: { value: BountyStatus | "all"; label: string }[] = [
-  { value: "all", label: "All active" },
+  { value: "all", label: "All" },
   { value: "open", label: "Open" },
   { value: "judging", label: "Judging" },
   { value: "winner-announced", label: "Winner" },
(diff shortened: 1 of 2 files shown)


arc0btc (approved):
Replaces the bounty status `<select>` filter with inline chip buttons and removes the server-side `status: "active"` filter so all bounties are returned and filtered client-side. Good UX improvement.

**What works well:**
- The `role="tablist"` + `role="tab"` + `aria-selected` ARIA pattern is appropriate here — matches the semantics of mutually-exclusive filter states. `aria-label` on the container is a nice touch.
- Count badges per chip are a clear affordance — hiding the badge when count is 0 (`{count > 0 && ...}`) avoids zero-clutter.
- The inline comment in `page.tsx` explaining *why* the status filter was removed is exactly the kind of non-obvious context that should live in the code.
- `?? 0` guards on count derivation prevent NaN renders if `initialBounties` is undefined.

**[question] Focus-visible ring on chip buttons** (`BountyDirectory.tsx`)
The chip `<button>` elements have no explicit `focus-visible:ring-*` classes. If global Tailwind base styles reset `outline` (common in `@layer base { button { outline: none } }`), these chips would be invisible to keyboard users navigating with Tab. Could you confirm a focus ring is rendered either globally or add one here?

```suggestion
              className={`inline-flex items-center gap-1.5 whitespace-nowrap rounded-full border px-3 py-1.5 text-sm transition-colors focus-visible:outline-none focus-visible:ring-2 focus-visible:ring-[#F7931A]/60 focus-visible:ring-offset-1 focus-visible:ring-offset-black ${
```

**[suggestion] Precompute counts outside the render loop** (`BountyDirectory.tsx`)
`initialBounties?.filter(...)` runs for each chip on every render. At 100 bounties it's negligible, but a `useMemo` makes the intent clearer and guards against future list growth:

```suggestion
  const statusCounts = useMemo(() => {
    const counts: Record<string, number> = { all: initialBounties?.length ?? 0 };
    for (const b of initialBounties ?? []) {
      counts[b.status] = (counts[b.status] ?? 0) + 1;
    }
    return counts;
  }, [initialBounties]);
```

Then `count = statusCounts[opt.value] ?? 0` inside the map.

**[nit] `aria-controls` pairing**
Strict tablist/tabpanel semantics expect each `role="tab"` to have `aria-controls` pointing to the panel it reveals. Since the bounty list isn't wrapped in a `role="tabpanel"`, screen readers may not expose the relationship. Not blocking — many filter chip patterns skip this — but worth noting for completeness.

**Code quality notes:**
- Logic is clean and self-contained. No dead imports visible in the diff.
- The restructured JSX separates the chip row and the search row into two distinct `<div>` blocks, which is easier to scan than the original mixed layout.

**Operational note:** No impact on our sensors — we read the bounty API directly, not the UI page. The server-side change (removing `status: "active"`) means the query now returns more rows up to the `limit: 100` cap. Still safe given the cap, but if the bounty count grows significantly the limit should be revisited.

PR #877, opened 2026-05-17
Swr/15min cache dry

biwasxyz:


arc0btc (approved):
Centralizes SWR key building into `lib/swr-keys.ts` and adds a global 15-minute cache via `app/providers.tsx` — solid refactor that eliminates scattered inline URL strings and makes cache invalidation predictable.

**What works well:**
- `swrKeys` as a single source of truth for all endpoint URLs means `swrMatchers` can reliably cover every cached variant — the `anyInbox`, `anyReputation`, etc. matchers are a clean solution to wildcard invalidation
- The `dedupingInterval` overrides in `ActivityFeed`, `ActivityFeedHero`, and `RelayStatus` are the right fix — the global 15-min dedupe would silently swallow polling ticks without them, and the comments explain the hazard clearly
- `useReputationData` seeding from the SWR cache before fetching is a nice touch; agents you visited in the last 15 minutes load instantly instead of being re-requested
- `ActivityFeedHero` converting to SWR with a `compare` function is elegant — hash-based dedup prevents the animation from resetting on identical payloads, which was the manual `hashRef` pattern before
- `InboxPage` splitting into first-page SWR + component-state accumulation for pagination is the right architecture. SWR holds the hot path; paginated state doesn't pollute the cache with positional keys

**[suggestion] `baseReplies` not memoized alongside `baseReceived` and `baseSent`** (`app/inbox/[address]/page.tsx`)

`baseReceived` and `baseSent` are both wrapped in `useMemo`, but `baseReplies` is computed inline:
```ts
const baseReplies = inboxData?.inbox.replies ?? {};
```
This creates a new object reference on every render. Any child that receives `replies = { ...baseReplies, ...extraReplies }` will re-render unnecessarily. Should be consistent:
```suggestion
  const baseReplies = useMemo(
    () => inboxData?.inbox.replies ?? {},
    [inboxData]
  );
```
And `receivedMessages`, `sentMessages`, and `replies` themselves are spread into new arrays/objects on every render — wrapping those in `useMemo` too would prevent downstream re-renders when neither the SWR data nor the extra pages changed.

**[suggestion] Inline outbox fetcher should be module-level** (`app/inbox/[address]/page.tsx`)

```ts
const { data: outboxData } = useSWR(..., {
  fetcher: async (url: string) => { ... },
  ...
});
```
SWR captures the fetcher at mount time so this doesn't cause refetches, but it's a new closure every render and an anti-pattern that confuses readers. Extract to module scope:

```ts
async function outboxFetcher(url: string) {
  try {
    return await fetcher<OutboxResponse>(url);
  } catch {
    return { outbox: undefined };
  }
}
```

**[nit] `reputationFeedback` builds its URL inconsistently with other keys** (`lib/swr-keys.ts:57`)

All other keys use `buildQuery`. This one manual-concatenates `?type=feedback&cursor=...`. Low risk since the URL is stable, but if the API ever gains another param this is the one that'll get missed:
```ts
reputationFeedback: (address: string, cursor?: number | null) =>
  `/api/identity/${encodeURIComponent(address)}/reputation${buildQuery({
    type: "feedback",
    cursor: cursor ?? undefined,
  })}`,
```

**[question] Direct `cache.get()` access in `useReputationData`** (`app/agents/AgentList.tsx:67`)

```ts
const cached = cache.get(swrKeys.reputationSummary(agent.btcAddress))?.data as ...
```

`cache` from `useSWRConfig()` is SWR's internal `Cache<any>` interface, and `.data` on its returned value is an internal `State` field. This works today but it's accessing implementation internals rather than the public API. The public alternative is `mutate(key)` with no value to check cache presence, or restructuring to let SWR itself manage the fetch (passing each key to a separate `useSWR` call). Worth a comment noting the internal API dependency so it's visible during future SWR upgrades.

**Code quality notes:**
- `buildQuery` correctly skips `false` values, which handles the `includePartners` boolean elegantly — no risk of `?include=false` leaking into URLs
- The `swrMatchers` using `key.startsWith(prefix)` pattern is correct for SWR's `mutate(matcher)` API
- `DEFAULT_CACHE_MS` exported from `swr-keys.ts` gives polling components a shared constant to override — better than magic numbers
- CF deploy is green per the bot comment

**Operational note:** We use the inbox API in our production sensors and have seen the endpoint return 404 when an agent hasn't registered yet (not a server error). The outbox null-shape tolerance in the new code matches what we see operationally — good defensive handling.

PR #883, opened 2026-05-19
fix(inbox): clamp reconciliation queue to nonceExpiresAt (#375)

whoabuddy:
## Summary

Stops the inbox reconciliation queue from polling a stale `paymentId` past the relay's 10-minute sponsor-nonce TTL, and replaces the stale entry with a fresh sponsor assignment via `submitPayment`.

- **Root cause (issue #375 in aibtcdev/x402-sponsor-relay):** The relay reclaims sponsor nonces after `STALE_THRESHOLD_MS = 10 min`, but the reconciliation queue was polling for up to ~70 min (`RECONCILIATION_DELAYS_SECONDS` × `MAX_RECONCILIATION_ATTEMPTS`). Past the relay's TTL, polling the old `paymentId` is wasteful and later retries using the same sponsored hex produce `ConflictingNonceInMempool` errors.
- **Fix (Option C from the issue):** At staging time, record `nonceExpiresAt = now + 10 min` alongside the original `txHex` and `settleOptions`. In the queue, check `isSponsorNonceExpired()` before each reconcile cycle. Past TTL: call `rpc.submitPayment(txHex, settle)` to obtain a fresh sponsor assignment, update the staged record, and re-enqueue with the new `paymentId`.

## Changes

| File | Change |
|------|--------|
| `lib/inbox/constants.ts` | Add `SPONSOR_NONCE_TTL_MS = 600 000 ms` (mirrors relay `STALE_THRESHOLD_MS`) |
| `lib/inbox/types.ts` | Extend `StagedInboxMessage` with `txHex?`, `nonceExpiresAt?`, `settleOptions?`, `paymentIdentifier?` |
| `lib/inbox/index.ts` | Export new type `StagedSettleOptions` and constant `SPONSOR_NONCE_TTL_MS` |
| `app/api/inbox/[address]/route.ts` | Populate new fields at staging time |
| `lib/inbox/reconciliation-queue.ts` | Add TTL clamp + re-submit path (import helpers; add `isSponsorNonceExpired`; splice TTL check before reconcile) |
| `lib/inbox/__tests__/reconciliation-queue.test.ts` | 4 new integration tests covering all TTL paths |

## Behavior guarantees

- **Pre-TTL:** Normal `checkPayment` poll path unchanged.
- **Post-TTL + txHex stored:** Call `submitPayment(txHex, settle)` → update staged record → re-enqueue with new `paymentId`.
- **Post-TTL + no txHex (pre-Phase-5 records):** Log warn, ack/discard gracefully.
- **`submitPayment` failure post-TTL:** Ack without retrying. Do NOT fall back to stale hex.
- **Backward compatible:** Old staged records without `nonceExpiresAt` fall back to `stagedAt + SPONSOR_NONCE_TTL_MS` for the expiry check.

## Test plan

- [ ] `npm run typecheck` passes (no new errors beyond pre-existing worktree issues)
- [ ] `npm test -- reconciliation-queue` — 8 tests pass in main file (4 existing + 4 new TTL tests)
- [ ] Verify TTL test: expired + txHex → submitPayment called, staged record updated, re-enqueued with new paymentId
- [ ] Verify TTL test: expired + no txHex → ack without submitPayment
- [ ] Verify TTL test: submitPayment rejection → ack without retry
- [ ] Verify TTL test: non-expired → normal checkPayment path (submitPayment not called)

## Cross-links

- Relay wire contract (Phase 3): aibtcdev/x402-sponsor-relay#383 (`nonceExpiresAt` + `sponsorNonceValidForMs` on `/sponsor` responses)
- Relay contract test (Phase 4): aibtcdev/x402-sponsor-relay#384
- Issue: aibtcdev/x402-sponsor-relay#375 (consumer half, Option C)

### DIFF
--- a/app/api/inbox/[address]/route.ts
+++ b/app/api/inbox/[address]/route.ts
@@ -19,8 +19,9 @@ import {
   buildSenderAuthMessage,
   DEFAULT_RELAY_URL,
   enqueueInboxReconciliation,
-  SPONSOR_NONCE_TTL_MS,
 } from "@/lib/inbox";
+// Import directly from constants to avoid test-mock interference on the barrel.
+import { SPONSOR_NONCE_TTL_MS } from "@/lib/inbox/constants";
 import { verifyBitcoinSignature } from "@/lib/bitcoin-verify";
 import { networkToCAIP2, X402_HEADERS } from "x402-stacks";
 import type { PaymentPayloadV2 } from "x402-stacks";
--- a/app/api/inbox/[address]/route.ts
+++ b/app/api/inbox/[address]/route.ts
@@ -19,6 +19,7 @@ import {
   buildSenderAuthMessage,
   DEFAULT_RELAY_URL,
   enqueueInboxReconciliation,
+  SPONSOR_NONCE_TTL_MS,
 } from "@/lib/inbox";
 import { verifyBitcoinSignature } from "@/lib/bitcoin-verify";
 import { networkToCAIP2, X402_HEADERS } from "x402-stacks";
@@ -1535,11 +1536,28 @@ export async function POST(
       const checkStatusUrl =
         paymentResult.checkStatusUrl ?? `/api/payment-status/${paymentResult.paymentId}`;
 
+      // Capture the original sponsored tx hex and settle options so the
+      // reconciliation queue can re-submit after the relay's sponsor nonce
+      // expires (Phase 5 of quest nonce-conflict-attribution, issue #375).
+      const stagedTxHex = paymentPayload?.payload?.transaction ?? undefined;
+      const stagedNonceExpiresAt = new Date(Date.now() + SPONSOR_NONCE_TTL_MS).toISOString();
+      const stagedSettleOptions =
+        agent.stxAddress && paymentRequirements.amount
+          ? {
+              expectedRecipient: agent.stxAddress,
+              minAmount: String(paymentRequirements.amount),
+              tokenType: "sBTC" as const,
+            }
+          : undefined;
+
       await storeStagedInboxPayment(kv, {
         paymentId: paymentResult.paymentId,
         createdAt: now,
         ...(senderAgent?.btcAddress && { senderSentIndexBtcAddress: senderAgent.btcAddress }),
         message,
+        ...(stagedTxHex && { txHex: stagedTxHex }),
+        nonceExpiresAt: stagedNonceExpiresAt,
+        ...(stagedSettleOptions && { settleOptions: stagedSettleOptions }),
       });
 
       await enqueueInboxReconciliation(
--- a/lib/inbox/__tests__/reconciliation-queue.test.ts
+++ b/lib/inbox/__tests__/reconciliation-queue.test.ts
@@ -389,4 +389,340 @@ describe("inbox reconciliation queue", () => {
       })
     );
   });
+
+  // ---- Sponsor-nonce TTL tests (#375 Phase 5) ----
+
+  it("acks when nonce is expired and staged record has no txHex (pre-Phase-5 record)", async () => {
+    const kv = createMockKV();
+    const { db } = createMockD1();
+    // Store a staged record WITHOUT txHex (old-format record)
+    const stagedAt = new Date(Date.now() - 15 * 60 * 1000).toISOString(); // 15 min ago
+
+    await storeStagedInboxPayment(kv, {
+      paymentId: "pay_ttl_no_txhex",
+      createdAt: stagedAt,
+      // no txHex, nonceExpiresAt is implicitly expired (15 min ago)
+      message: {
+        messageId: "msg_ttl_no_txhex",
+        fromAddress: "SP123",
+        toBtcAddress: "bc1recipient",
+        toStxAddress: "SP456",
+        content: "hello",
+        paymentSatoshis: 100,
+        sentAt: stagedAt,
+        paymentStatus: "pending",
+        paymentId: "pay_ttl_no_txhex",
+      },
+    });
+
+    const ack = vi.fn();
+    const retry = vi.fn();
+    const submitPayment = vi.fn();
+
+    await processInboxReconciliationQueue(
+      {
+        queue: "landing-page-inbox-reconciliation",
+        messages: [
+          {
+            id: "queue-msg-ttl-1",
+            timestamp: new Date(),
+            attempts: 1,
+            body: {
+              paymentId: "pay_ttl_no_txhex",
+              stagedAt, // 15 min ago — nonce expired
+              attempt: 0,
+              source: "queue_retry" as const,
+            },
+            ack,
+            retry,
+          },
+        ],
+        ackAll: vi.fn(),
+        retryAll: vi.fn(),
+      },
+      {
+        VERIFIED_AGENTS: kv,
+        DB: db,
+        X402_RELAY: {
+          submitPayment,
+          checkPayment: vi.fn(), // should NOT be called
+        },
+        INBOX_RECONCILIATION_QUEUE: { send: vi.fn(), sendBatch: vi.fn() },
+      },
+      mocks.logger,
+      "test-version"
+    );
+
+    // Should ack (discard) without calling checkPayment or submitPayment
+    expect(ack).toHaveBeenCalledTimes(1);
+    expect(retry).not.toHaveBeenCalled();
+    expect(submitPayment).not.toHaveBeenCalled();
+    expect(mocks.logger.warn).toHaveBeenCalledWith(
+      "payment.retry_decision",
+      expect.objectContaining({
+        paymentId: "pay_ttl_no_txhex",
+        action: "ack_nonce_expired_no_tx_hex",
+      })
+    );
+  });
+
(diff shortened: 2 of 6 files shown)


arc0btc (approved):
Fixes the reconciliation queue polling past the relay's 10-min sponsor-nonce TTL — a real source of `ConflictingNonceInMempool` errors we've seen operationally.

**What looks good:**
- All four TTL paths covered and tested: pre-Phase-5 no-txHex discard, expired+txHex resubmit, submitPayment failure ack, and non-expired passthrough.
- Backward compat is solid — old records without `nonceExpiresAt` fall back to `stagedAt + SPONSOR_NONCE_TTL_MS` rather than crashing.
- The hard rule ("Do NOT fall back to stale hex on submitPayment failure") is exactly right. Acking on failure is the safest call.
- KV write failure after successful resubmit is handled — logs `ack_resubmit_kv_write_failed` and acks to avoid orphaned state.
- The `NONCE_EXPIRY_BUFFER_MS = 5_000` clock-skew guard is a nice touch.
- The `constants.ts` comment flags future `sponsorNonceValidForMs` forward-compat work clearly.

**[suggestion] Double KV read** (`lib/inbox/reconciliation-queue.ts`)
`getStagedInboxPayment` is called here for the TTL check, and `reconcileStagedInboxPayment` below likely reads the same key again for the non-expired path. Consider threading the already-fetched `stagedRecord` into `reconcileStagedInboxPayment` as an optional parameter to avoid the duplicate read.

**[suggestion] Old staged record left in KV after resubmit**
After `storeStagedInboxPayment` creates the new `pay_xyz_fresh` key, the old `pay_xyz` key is never deleted. It will expire via `STAGED_PAYMENT_TTL_SECONDS` naturally, but until then it's a stale record that could confuse monitoring. A best-effort delete of the old key after the KV write succeeds would clean this up.

**[nit] `isSponsorNonceExpired` — duplicate fallback paths** (`lib/inbox/reconciliation-queue.ts`)
The unparseable-`nonceExpiresAt` branch and the absent-`nonceExpiresAt` branch both derive `stagedAt + SPONSOR_NONCE_TTL_MS`. They could be collapsed:

```suggestion
function isSponsorNonceExpired(stagedAt: string, nonceExpiresAt?: string): boolean {
  const now = Date.now();
  const expiryMs = nonceExpiresAt ? new Date(nonceExpiresAt).getTime() : NaN;
  const effectiveExpiry = Number.isFinite(expiryMs)
    ? expiryMs
    : new Date(stagedAt).getTime() + SPONSOR_NONCE_TTL_MS;
  if (!Number.isFinite(effectiveExpiry)) return true;
  return now >= effectiveExpiry - NONCE_EXPIRY_BUFFER_MS;
}
```

**[question]** Is reading `sponsorNonceValidForMs` from relay responses (noted in the constants comment) tracked as a follow-up, or does it land in Phase 3 relay PR (#383)?

*Operational note:* We process x402 inbox messages with reconciliation cycles spanning 70+ minutes. ConflictingNonceInMempool failures are silent without explicit TTL modeling — this PR adds the right abstraction layer.

secret-mars (commented):
Substantive review — Phase 5 consumer side of the nonce-conflict-attribution quest. arc-APPROVED already; this review surfaces one load-bearing wire-contract gap + 3 smaller observations. Excellent end-to-end shape and the explicit "do NOT fall back to stale hex" discipline in the resubmit failure branches is exactly right.

## What I checked

- `lib/inbox/reconciliation-queue.ts:57-83` — `isSponsorNonceExpired()` derivation
- `lib/inbox/reconciliation-queue.ts:215-402` — TTL-check + resubmit branch (sponsor-fault recovery)
- `app/api/inbox/[address]/route.ts:1538-1561` — staging-time capture of txHex + nonceExpiresAt + settleOptions
- `lib/inbox/relay-rpc.ts:118-160` (existing) — `parseSubmitPaymentResult` shape
- `lib/inbox/types.ts` — `StagedInboxMessage` extension
- Cross-referenced [x402sr#383](https://github.com/aibtcdev/x402-sponsor-relay/pull/383) (Phase 3) wire contract: `nonceExpiresAt` + `sponsorNonceValidForMs` on `/sponsor` responses

## Findings

### 1. `nonceExpiresAt` never sourced from the relay — wire contract from #383 unused on consumer side

**Where:** Three places combine to make this a no-op.

`app/api/inbox/[address]/route.ts:1541`:
```ts
const stagedNonceExpiresAt = new Date(Date.now() + SPONSOR_NONCE_TTL_MS).toISOString();
```

`lib/inbox/reconciliation-queue.ts:328` (post-resubmit):
```ts
const newNonceExpiresAt = new Date(Date.now() + SPONSOR_NONCE_TTL_MS).toISOString();
```

`lib/inbox/relay-rpc.ts:121-141` (existing — `parseSubmitPaymentResult`) only extracts `paymentId`, `accepted`, `status`, `checkStatusUrl`. The relay's `nonceExpiresAt` field from PR#383's wire contract is never read off the response.

**Why this matters:** The function comment at `reconciliation-queue.ts:60-66` says

> *"Uses `nonceExpiresAt` from the staged record when available (preferred — relay clock is authoritative). Falls back to `stagedAt + SPONSOR_NONCE_TTL_MS` for records created before Phase 5..."*

But the staged record is *always* populated from LP local clock + LP's constant — never from the relay. So:

- The `nonceExpiresAt` branch in `isSponsorNonceExpired` is effectively dead code (semantically identical to the fallback when both compute the same value)
- Relay-side clock skew is invisible to the LP
- The whole reason PR#383 added `nonceExpiresAt` to the response was so consumers could trust the relay's authoritative reclaim deadline rather than guessing; this PR doesn't consume it

**Suggested fix:** Extend `parseSubmitPaymentResult` to extract `nonceExpiresAt` (and optionally `sponsorNonceValidForMs`) from the relay response, surface it on `RelaySubmitResult`, then use it in route.ts staging instead of the local-derived value. That makes #383's wire contract live and the function comment honest. ~15 LOC.

If the relay's RPC binding doesn't yet surface `nonceExpiresAt` (PR#379 wires it through `okWithTx` HTTP-side; RPC-side surfacing should be confirmed against `aibtcdev/x402-sponsor-relay#379` head), call that out as a Phase 5.1 follow-up issue rather than landing this PR with a vestigial branch.

### 2. Test plan checkboxes unchecked

All 6 test-plan checkboxes in the PR body are unchecked `[ ]`. The pattern in this repo is to tick them after local verification. The diff includes 4 new TTL test cases — minor ask: confirm `npm run typecheck` + `npm test -- reconciliation-queue` were run locally and tick the boxes (or call out that CI will do this).

### 3. `getRetryDelaySeconds(body.attempt)` uses pre-increment attempt#

**Where:** `lib/inbox/reconciliation-queue.ts:353` after a successful resubmit:
```ts
const nextAttempt = body.attempt + 1;
const queueDelaySeconds = getRetryDelaySeconds(body.attempt);  // <-- old attempt
```

`getRetryDelaySeconds(body.attempt)` is called with the old attempt number, then `nextAttempt` is what's enqueued. So the new payment's first poll happens at the cadence-slot of the *old* payment's last attempt. If `body.attempt` is already near `MAX_RECONCILIATION_ATTEMPTS`, the next poll for the *fresh* relay-side sponsor nonce will use the longest delay (`RECONCILIATION_DELAYS_SECONDS[4] = 600s`) even though the payment is brand-new on the relay side and would benefit from the shorter early-attempt cadence.

Mild — the fresh-resubmit case is rare, and the staged record carries `body.attempt` semantics for the *staging*, not the relay-side polling. But worth confirming whether the intent is "fresh nonce ⇒ reset attempt counter to 0" (poll quickly) or "stay on the retry curve" (slow down).

### 4. KV-write-failure ack: documented orphan

**Where:** `lib/inbox/reconciliation-queue.ts:336-355`. After a successful relay-side resubmit, if the KV write to update the staged record fails, you ack the message. The comment correctly notes: *"old paymentId is stale, new paymentId is not persisted."*

This means the relay now has a freshly-sponsored payment with a 10-min TTL that the LP will never poll. The relay's NonceDO will reclaim the nonce after TTL but the in-flight broadcast may have succeeded by then. Net effect: a sponsored on-chain settlement with no consumer-side payment record — invisible to the inbox.

The ack-on-KV-failure decision is correct (alternative is double-sponsorship via retry). Just worth surfacing in observability so ops can spot this rare drift — perhaps a structured `payment.orphaned_after_kv_write_failure` log event at WARN level so any nonzero counts are alertable. Cheap addition.

## What I think the merge plan should look like

Finding #1 is the only one that meaningfully affects this PR's design intent. If the relay's RPC binding already surfaces `nonceExpiresAt`, ~15 LOC extension to `parseSubmitPaymentResult` + use in route.ts + queue.ts unblocks consuming it. If it doesn't, leave the local-derivation in place but fix the function comment to admit it, and open a follow-up issue.

Findings #2 / #3 / #4 are non-blocking observability/edge-case notes. The structural shape of the TTL clamp + resubmit branch is sound; the 9-cell behavior matrix in the body matches the implementation; the `stagedRecord.txHex` absence branch (pre-Phase-5 records) is correctly handled with discard-and-log.

Phase 5 of the quest reads complete pending #1; once #381 + #382 land and this rebases on a Phase-1-merged main, the consumer-side recovery is end-to-end provable.

PR #888, opened 2026-05-19
feat(d1): add agents.last_check_in_at for P2 heartbeat ratelimits work

whoabuddy:
## Summary

Migration-only PR: adds a nullable \`TEXT\` column \`last_check_in_at\` to the \`agents\` D1 table. This is the durable last-seen state half of P2 — the rate-limit window itself moves to a Cloudflare \`ratelimits\` binding in the follow-up code PR.

## Why

Retires the \`checkin:{btcAddress}\` KV key family. P0 baseline confirmed the no-\`{ expirationTtl }\` bug at \`lib/heartbeat/kv-helpers.ts:60\` is still live (696 lifetime keys, growing forever — original retro flag). The KV-RMW shape is also the largest remaining write surface after P1 closed (estimated ~8K writes/hr at active-agent steady state vs P0's measured 26.4K writes/24h total).

The new column is semantically distinct from \`last_active_at\`:
- \`last_active_at\` updates on any liveness signal (register, vouch, identity, challenge, ...).
- \`last_check_in_at\` is specifically the most recent successful POST \`/api/heartbeat\` — used by the response payload to compute \`nextCheckInAt\`.

Both nullable. No backfill — column starts NULL and each agent's next check-in populates it (~30 min to ~95% population at the 5-min heartbeat convention).

## Test plan

- [x] Migration parses; \`wrangler d1 migrations list --remote\` shows the new file as pending.
- [ ] **Post-merge operational gate**: I will run \`npm run wrangler -- d1 migrations apply landing-page --env production --remote\` from this repo and verify the column with \`SELECT name, type FROM pragma_table_info('agents') WHERE name = 'last_check_in_at'\`. Output captured in \`phases/P2/verify.md\` per the quest's D1 operational-gate discipline.
- [ ] After verification, the follow-up code PR (\`p2b-heartbeat-ratelimits-binding\`) is unblocked.

## Quest context

- Quest folder: \`~/.planning/active/2026-05-18-kv-d1-pattern-finish/\`
- Phase plan: \`phases/P2/plan.md\` (dated 2026-05-19T23:45Z)
- Phase spec: \`PHASES.md § Phase 2\`
- Predecessor: \`P1\` merged earlier today as \`c77f043e\` (#886)

## Reviewers

- @arc0btc @secret-mars — primary review (one-pass)
- @steel-yeti — dev-council, non-blocking
- Copilot — auto-attaches

### DIFF
--- a/migrations/015_last_check_in_at.sql
+++ b/migrations/015_last_check_in_at.sql
@@ -1,0 +1,24 @@
+-- Migration 015: add agents.last_check_in_at for P2 heartbeat ratelimits work.
+--
+-- Replaces the `checkin:{btcAddress}` KV key family (no-`{ expirationTtl }`
+-- bug confirmed at the predecessor retro and the 2026-05-18-kv-d1-pattern-finish
+-- quest P0 baseline — 696 lifetime keys, growing forever) with a durable D1
+-- column for the most-recent successful POST /api/heartbeat timestamp.
+--
+-- The enforcement window itself moves to the `RATE_LIMIT_CHECKIN` Cloudflare
+-- `ratelimits` binding (1 req / 60s), so this column is for display and
+-- analytics — not the rate-limit critical path. The response payload reads
+-- the value back to compute `nextCheckInAt`.
+--
+-- Semantically distinct from `last_active_at`: that column updates on any
+-- liveness signal (register, vouch, identity, challenge, ...); this one is
+-- specifically the most recent heartbeat POST. Both nullable.
+--
+-- Backfill: none. Column starts NULL; each agent's next successful POST
+-- /api/heartbeat populates it. P0 noted ~696 active agents with check-in
+-- records; on the 5-min heartbeat convention, ~95% population is reachable
+-- within ~30 min of normal traffic.
+--
+-- Forward-only. Rollback path is a follow-up DROP COLUMN migration.
+
+ALTER TABLE agents ADD COLUMN last_check_in_at TEXT;


arc0btc (approved):
Adds `agents.last_check_in_at` as a nullable TEXT column for the D1-side of P2 — retiring the `checkin:{btcAddress}` KV key family that's been leaking forever-keys (no expirationTtl).

**What looks good:**
- `ALTER TABLE ADD COLUMN` with a nullable, no-default column is O(1) in D1/SQLite — no table rewrite, safe to apply under live traffic
- Clean semantic split from `last_active_at` — the migration comment explains the distinction clearly and future readers will thank you
- Forward-only with rollback path explicitly documented (DROP COLUMN follow-up); that's good operational discipline
- ~30 min to 95% population at the 5-min heartbeat cadence is a realistic and well-reasoned estimate
- Migration file naming (`015_`) stays consistent with the sequence

**[question]** Depending on how the follow-up code PR queries this column: if it ever runs a cross-agent scan (`WHERE last_check_in_at < ?` for analytics or stale-agent detection), you'll want an index. Per-agent lookups (`WHERE btcAddress = ?` or by PK) don't need it. Worth deciding in the code PR before it ships.

**[nit]** No CHECK constraint on the TEXT format — fine since only `/api/heartbeat` writes this and should always emit ISO-8601. Just noting nothing enforces it at the DB level if other paths ever write to the column.

**Code quality notes:** Migration is minimal and correct. Comment header is thorough — the WHY, predecessor context, and population timeline are all there. No simplifier opportunities on a single-statement migration.

**Landing-page review (Performance / Composition / UI/Accessibility):** N/A — no TSX/JSX/CSS files changed.

*Operational note:* We track this repo's heartbeat path via our aibtc-network sensor. The growing-forever KV key bug has been flagged in our retros. This migration is the right move — D1 state is more auditable and visible than KV for per-agent heartbeat timestamps. Shipping this cleanly as a migration-only PR before the rate-limit binding change is the correct split.

whoabuddy:
Thanks @arc0btc + @steel-yeti — addressing the follow-ups so PR 2 (`p2b-heartbeat-ratelimits-binding`) ships with these locked in:

**@arc0btc question (index on `last_check_in_at`):** PR 2 only reads/writes this column per-agent by `btc_address` (PK lookup) — there is no cross-agent `WHERE last_check_in_at < ?` scan in scope. Confirmed in the P2 plan's file/line refs: the only write is `UPDATE agents SET last_check_in_at = ? WHERE btc_address = ?` and the only read is from the just-written timestamp in the response payload (no read-back). **No index added in PR 2.** If a future analytics path needs stale-agent detection, that's the right PR to add the index in — flagged for the runbook in P5.

**@arc0btc nit (no CHECK on TEXT format):** Acknowledged — only `/api/heartbeat` writes this column, and the write site uses `new Date().toISOString()` (ISO-8601 by construction). No DB-level enforcement, but the write path is single-source. Worth a note in the P5 runbook on which D1 columns are "single-writer convention" vs DB-enforced.

**@steel-yeti shadow cycle 6 follow-ups for PR 2:**
1. `ratelimits` binding stays the **enforcement** source; D1 column is durable display/analytics only — matches the plan exactly (binding does the 1/60s gate, D1 update is synchronous after `success: true`, never used as the rate-limit lock).
2. NULL `last_check_in_at` is the expected transitional state — the plan's "no backfill, ~30min soak path" decision already treats this as transient. `nextCheckInAt` computation in the response will be NULL-safe in PR 2.
3. `last_active_at` vs `last_check_in_at` semantic distinction → preserved in PR 2 code comments + slated for P5 runbook.

Squash-merging now. Remote migration apply + column verification before PR 2 ships per QUEST.md operational gate.

Issue #822, opened 2026-05-13
Trading Competition: weekly snapshot finalization and reward accounting


whoabuddy:
## Summary

Define and implement the weekly snapshot/finalization path for trading competition rewards. This follows the canonical rules in #815: live rankings can use drifting Unrealized P&L, but paid rewards must use a frozen, deterministic snapshot.

## Why

The live leaderboard is intentionally exciting: prices move, P&L moves, and agents can watch the board update during the week. Rewards need a stricter accounting model:

- final results must not change after payout review
- every winner needs an auditable basis for the payout
- scheduler catch-up needs a grace window before finalization
- sBTC payouts to the agent STX address need status/txid tracking

## Proposed weekly flow

1. Create a weekly round with `starts_at`, `ends_at`, and `finalization_status = open`.
2. During the week, live UX continues to compute Unrealized P&L from current Tenero prices.
3. At round close, stop accepting new trades into that round by `burn_block_time`.
4. Allow a catch-up grace period for scheduler/chainhook ingestion of already-confirmed trades whose `burn_block_time` is inside the round.
5. Snapshot token prices/decimals for every token present in successful round swaps.
6. Compute and persist immutable round results using frozen prices.
7. Review results, mark the round finalized, then create reward payout rows.
8. Pay rewards to each winning agent's STX address in sBTC and record payout txids.

## Calculation rules

Use successful swaps only:

```sql
WHERE tx_status = 'success'
  AND source IN ('agent', 'cron', 'chainhook')
  AND burn_block_time >= round.starts_at
  AND burn_block_time < round.ends_at
```

For each agent:

```text
volume_usd = Σ(amount_in * frozen_price[token_in])
received_usd = Σ(amount_out * frozen_price[token_out])
pnl_usd = received_usd - volume_usd
pnl_percent = pnl_usd / volume_usd
```

Suggested eligibility/accounting fields:

- `trade_count`
- `priced_trade_count`
- `unpriced_trade_count`
- `unpriced_tokens`
- `volume_usd`
- `received_usd`
- `pnl_usd`
- `pnl_percent`
- `latest_trade_at`
- `source_counts` (`agent`, `cron`, `chainhook`)
- `calculated_at`

Both legs of a swap should be priced for that swap to contribute to USD totals. Unpriced swaps should be visible in the snapshot rather than silently zeroed.

## Reward categories to decide

Candidate categories:

- Overall trader: highest `pnl_usd`, with `volume_usd` as tiebreak.
[...]

PR #898, opened 2026-05-20
fix(scheduler): enable TENERO_REFRESH_ENABLED at top-level (closes #880)

whoabuddy:
## Summary

The Cloudflare Workers Build pipeline runs bare \`wrangler deploy\` (no \`--env\`), which reads the top-level \`vars\` block in \`wrangler.jsonc\`. The prior guard intentionally omitted \`TENERO_REFRESH_ENABLED\` there as a "no-env safety" measure — but the practical effect was that the SchedulerDO has been skipping every Tenero tick in prod since the env var was introduced, leaving the KV price cache empty (confirmed: \`wrangler kv key list\` returns zero \`tenero:*\` keys).

That blocks the snapshot action of the new competition-finalize flow (#822 / #897), which reads from the KV cache to freeze per-token prices. The new \`empty_price_cache\` pre-flight gate would return 503 on any real snapshot attempt.

Setting \`TENERO_REFRESH_ENABLED: "true"\` at the top level lets the scheduler populate the cache on its 5-minute tick.

\`env.preview\` is left without the var, so preview deploys still cannot consume the production Tenero quota.

Closes #880.

## Test plan
- [ ] CI green
- [ ] Post-merge: wait one scheduler tick (~5 min), confirm \`tenero:*\` KV keys exist
- [ ] Post-merge: \`POST /api/admin/competition/finalize?dry-run=true { action: "snapshot" }\` returns \`wouldCapture.priced > 0\` instead of 503

### DIFF
--- a/wrangler.jsonc
+++ b/wrangler.jsonc
@@ -34,11 +34,16 @@
       "remote": true
     }
   ],
-  // Top-level vars: used by bare `wrangler deploy` (without --env). Keep
-  // Tenero refresh disabled here so a no-env deployment cannot consume the
-  // production Tenero key/quota even if the secret exists on the Worker name.
+  // Top-level vars: used by bare `wrangler deploy` (without --env), which is
+  // what the Cloudflare Workers Build pipeline runs. The prior "keep Tenero
+  // refresh disabled here" guard meant the SchedulerDO never populated the
+  // KV cache in prod (issue #880) — so the competition-finalize snapshot
+  // path can't fetch frozen prices. Enabling here is the resolution.
+  // env.preview below still omits TENERO_REFRESH_ENABLED, so a preview
+  // build cannot accidentally consume the production Tenero quota.
   "vars": {
-    "DEPLOY_ENV": "production"
+    "DEPLOY_ENV": "production",
+    "TENERO_REFRESH_ENABLED": "true"
   },
   // First-party rate limit bindings. Mirrors the agent-news pattern (agent-news#704/#705).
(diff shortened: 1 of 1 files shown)


arc0btc (approved):
Enables the Tenero scheduler in production by adding `TENERO_REFRESH_ENABLED: "true"` to the top-level `vars` block — unblocking the competition-finalize snapshot path.

**What works well:**
- Root cause correctly identified: CF Workers Build runs bare `wrangler deploy` (no `--env`), reads top-level `vars` only. The original "no-env safety" guard was solving a real concern but had the side effect of leaving the KV price cache permanently empty in prod.
- The fix is surgical — exactly one config line, no collateral changes.
- Preview isolation preserved: `env.preview` still omits the var, so preview builds can't consume prod Tenero quota. That's the right call.
- Comment is detailed and links directly to the issue (#880) and downstream impact (#822/#897). Future readers will understand why this line exists.
- CF deployment already confirmed successful (build log green).

**Performance / Composition / UI/Accessibility:** Not applicable — `wrangler.jsonc` only, no React/TS/CSS touched.

**Operational context:** This unblocks the `POST /api/admin/competition/finalize?dry-run=true { action: "snapshot" }` path, which reads KV for frozen prices. With the cache empty the snapshot would 503. The test plan covers exactly this verification — wait one tick (~5 min), confirm `tenero:*` KV keys, then dry-run the snapshot. Solid test plan.

✅ LGTM. No blocking issues.

PR #902, opened 2026-05-21
feat(bounty): allow any registered agent to post bounties

biwasxyz:
## Summary

Drops the Genesis (Level 2+) requirement on `POST /api/bounties`. Any registered (L1+) agent can now post bounties — symmetric with the submitter gate.

## Why

The Genesis gate was friction without much upside: it required posters to tweet + claim before they could post a bounty, even though submitting work was already open to all registered agents. Making posting symmetric with submitting opens the bounty board to the full agent network.

## What changed

**Code**
- `app/api/bounties/route.ts` — removed the level computation + 403 gate. The `lookupAgent` call above already returns `404 agent_not_found` when no `AgentRecord` exists, which is the L1+ check by definition. Also dropped now-unused imports: `computeLevel`, `ClaimStatus`, `MIN_POSTER_LEVEL`.
- `lib/bounty/constants.ts` — deleted `MIN_POSTER_LEVEL` (no consumers remain). Generalized `MIN_SUBMITTER_LEVEL` comment to describe it as the participant floor.
- `lib/bounty/index.ts` — removed `MIN_POSTER_LEVEL` from the barrel.

**Discovery surfaces (per CLAUDE.md rule: keep all together)**
- `CLAUDE.md` — Bounty System section + APIs table.
- `app/api/openapi.json/route.ts` — `createBounty` operation summary, description, and 403 response removed.
- `app/docs/[topic]/route.ts` — bounties topic Poster role + section heading.
- `app/llms-full.txt/route.ts` — ecosystem bounty blurb and the "What's Next" links.
- `app/api/bounties/route.ts` — self-doc envelope strings.

**UX**
- `app/bounty/page.tsx` — meta description.
- `app/bounty/BountyDirectory.tsx` — page tagline.
- `app/bounty/new/page.tsx` — page copy + the "Not Genesis yet?" hint replaced with a "Not registered yet?" hint pointing to `POST /api/register`.

## Net

10 files, +21 / −51. Mostly subtractive.

## Test plan

- [x] `npx tsc --noEmit` — no new errors in modified files
- [x] `npx vitest run lib/bounty` — 42/42 pass
- [x] `npm run lint` — clean (only pre-existing `` warnings in unrelated files)
- [ ] Manual: `POST /api/bounties` as an L1 agent should return 201 (previously 403). L0 (no AgentRecord) still returns 404.


arc0btc (approved):
Drops the Genesis gate from `POST /api/bounties`, making bounty posting symmetric with submitting — both now require L1+ registration only. Clean, mostly-subtractive diff.

**What works well:**
- The authorization logic is sound: `lookupAgent` already returns 404 for unregistered agents, which *is* the L1+ gate by definition. The old level gate was a second check on top — removing it is correct and the comment in the code explains why clearly.
- Strictly a performance improvement: removes one KV read (`kv.get('claim:{address}')`) and a `computeLevel()` call per POST.
- Impressively thorough doc sync — CLAUDE.md, OpenAPI spec, docs route, llms-full.txt, UX copy, and meta descriptions all updated consistently. This is the level of care these cross-cutting changes require.
- Unused imports (`computeLevel`, `ClaimStatus`, `MIN_POSTER_LEVEL`) cleaned up cleanly.

---

**[suggestion] `app/llms-full.txt/route.ts` — Bounty posting is now L1 but still gated behind Genesis onboarding copy**

The heartbeat orientation "What's Next" section still presents item 3 (bounty posting) under the heading `"Now that you've reached Genesis (Level 2), you can:"`. With this change, L1 agents can now post bounties — but a freshly-registered agent reading the heartbeat flow will never see this guidance because it only surfaces in the Genesis tier. Consider moving bounty posting to the L1 orientation section, or restructuring the list to reflect the actual access tiers.

**[nit] `lib/bounty/constants.ts` — Updated comment is slightly imprecise**

```suggestion
/** Minimum level to submit work. Posting is gated by agent registration (AgentRecord existence). */
export const MIN_SUBMITTER_LEVEL = 1;
```

`MIN_SUBMITTER_LEVEL` is only used in the submitter path. The comment "Applies to both posters and submitters" could mislead a future maintainer into thinking this constant controls posting — it doesn't. Posting is now gated implicitly by `lookupAgent`. In practice they're equivalent (L1+ = registered agent), but the distinction matters if the floors ever diverge.

**[question] Spam / low-friction bounty flooding**

Any L1 agent can now create bounties with `rewardSats: 1` and no upfront lock. Is there a rate limit or minimum reward floor beyond `rewardSats > 0`? Not blocking — signature + registration is a real barrier — but worth considering before this opens up. A reasonable minimum (e.g. 10k sats) or a per-agent daily cap would reduce the surface area for junk bounties.

---

**Code quality:** No dead code, no redundant logic, clean abstraction level. Diff reads like a deletion, which is high praise.

*Operational note:* We track the bounty board via sensors. Arc0 is registered at L1+ and will now be able to post bounties, which we weren't able to before. The symmetric access model is cleaner for agent-to-agent coordination.

PR #912, opened 2026-05-24
feat(bounty): block multi-winner bounties at creation

messiawrq-design:
Fixes #908 

Because the underlying smart contract and bounty tracking systems only structurally support a single winner (`acceptedSubmissionId`, `paidTxid`), allowing posters to write "Up to 3 winners" or "first-come first-paid" in descriptions creates a UX trap for contributors whose valid submissions are implicitly disqualified when the first payout flips the bounty to a `paid` terminal state.

This PR implements "Option 1" (cheapest/safest): It adds a Regex validation guard in the `POST /api/bounties` handler to explicitly reject bounty creation requests containing multi-winner language, prompting the user to rewrite or split into N separate bounties.

Bounty Wallet: 6sF8p22Gg83NKTJ6dvya7Srv4USCniZnP47DwQwK7Mtp (Solana)

### DIFF
--- a/app/api/bounties/route.ts
+++ b/app/api/bounties/route.ts
@@ -202,6 +202,18 @@ export async function POST(request: NextRequest) {
     }
     const data = parsed.data!;
 
+    const multiWinnerRegex = /\b(up to \d+ (winners?|slots?)|first-come\s*first-paid|multiple winners?|\d+ slots? remain(ing)?)\b/i;
+    if (multiWinnerRegex.test(data.description) || multiWinnerRegex.test(data.title)) {
+      return NextResponse.json(
+        {
+          error: "multi_winner_unsupported",
+          message: "Bounties currently support a single winner. Your description suggests multi-winner — please rewrite or split into N separate bounties."
+        },
+        { status: 400 }
+      );
+    }
+
+
     // Replay-window check
     if (!isWithinSignatureWindow(data.signedAt, SIGNATURE_WINDOW_SECONDS)) {
       return NextResponse.json(
--- a/next.config.ts
+++ b/next.config.ts
@@ -5,6 +5,20 @@ const isGitHubPages = process.env.GITHUB_PAGES === "true";
 const nextConfig: NextConfig = {
   output: isGitHubPages ? "export" : undefined,
   basePath: isGitHubPages ? "/landing-page" : "",
+    async redirects() {
+    return [
+      {
+        source: '/bounties',
+        destination: '/bounty',
+        permanent: false,
+      },
+      {
+        source: '/bounties/:path*',
+        destination: '/bounty/:path*',
+        permanent: false,
+      }
+    ];
+  },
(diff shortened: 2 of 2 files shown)


arc0btc (approved):
Blocks multi-winner bounty creation at the API layer — good fix for the UX trap where contributors submit to a bounty that can only pay one winner.

**What works well:**
- Regex is readable and covers the most common phrasings in the PR description ("up to N winners", "first-come first-paid", "multiple winners", "N slots remaining")
- Error message is actionable — it tells the user exactly what to do (rewrite or split into N bounties)
- Touching only the API route keeps blast radius minimal

**[suggestion] Redirect permanence** (`next.config.ts`)
The `/bounties` → `/bounty` redirects use `permanent: false` (307 Temporary). For canonical URL cleanup, `permanent: true` (308 Permanent) is the right choice — it allows browsers and crawlers to cache the redirect and stop hitting the source URL on every request. If this is experimental and you might revert, `false` is fine, but if `/bounty` is the settled route name, make it permanent.

```suggestion
        permanent: true,
```
(applies to both redirect entries)

**[suggestion] Auth ordering** (`app/api/bounties/route.ts`)
The regex check at this point is before the replay-window check. If BIP-322/signature verification also happens after this block, unauthenticated callers can probe the blocklist without a valid signature. Consider verifying auth/signature first, then running content validation. If auth already runs before line 202 (earlier in this handler), ignore this — happy to confirm from context.

**[suggestion] Regex coverage gaps**
The blocklist catches the specific patterns from the PR description, but these are also unsupported and will be silently allowed through:
- `"2 winners"` / `"3 prizes"` (number + winner/prize without "up to")
- `"top 3 contributors"` / `"split between 3 agents"`

Consider adding `\d+\s*(winners?|prizes?)\b` to the alternation, or adding a short comment noting the known gaps so the next maintainer doesn't assume the list is exhaustive.

**[nit]** Double blank line before `// Replay-window check` — cosmetic.

**[nit]** `redirects()` in `next.config.ts` has extra leading spaces (4 extra) vs the other top-level config keys — formatter missed this.

**[question]** The PR body includes `Bounty Wallet: 6sF8p22Gg83NKTJ6dvya7Srv4USCniZnP47DwQwK7Mtp (Solana)` — looks like accidental copy-paste from a template? Just making sure this isn't meant to be wired into the code anywhere.

**[question]** The `/bounties` → `/bounty` redirects appear unrelated to the multi-winner guard. Are these cleaning up a prior route rename? Fine to bundle, just confirming intent.

---

**Performance:** No React/data-fetching concerns — this is a pure API route change with a fast regex test on already-parsed input. Negligible.

**Composition/UI:** No UI changes in scope.

*Operational note:* Arc runs as a registered agent in this ecosystem and can post bounties natively (per #902). The single-winner structural constraint is real — this guard prevents the exact confusion pattern described. Solid option-1 fix.

PR #913, opened 2026-05-24
fix(inbox): repair unread stats drift at source

liangtovi-debug:
## Summary

Fixes #906 by targeting the stats drift source instead of adding filtered runtime `COUNT(*)` scans.

- Removes the deprecated filtered `COUNT(*)` totalCount workaround from `GET /api/inbox/[address]`; the hot path stays on `agent_inbox_stats` O(1) counters.
- Adds `markMessageReadAndDecrementStats()`, a guarded D1 batch that sets `inbox_messages.read_at` and decrements `agent_inbox_stats.unread_count` together.
- Uses the guarded helper for explicit mark-read PATCH and for reply writes that implicitly mark an unread parent message as read.
- Adds migration `019_rebuild_agent_inbox_stats.sql` to rebuild `agent_inbox_stats` once from `inbox_messages`, repairing existing migration/backfill drift without runtime scans.
- Updates route/unit tests for the new read+stats invariant.

## Tests

- `npm test -- 'app/api/inbox/[address]/__tests__/d1-sentcount-partners.test.ts' 'app/api/inbox/[address]/[messageId]/__tests__/dual-write.test.ts' 'app/api/inbox/[address]/[messageId]/__tests__/write-path-d1-flip.test.ts' 'app/api/outbox/[address]/__tests__/write-path-d1-flip.test.ts' 'lib/inbox/__tests__/d1-dual-write.test.ts'`
- `npm test -- 'app/api/inbox/[address]/__tests__/dual-write.test.ts' 'app/api/outbox/[address]/__tests__/d1-reads-flip.test.ts' 'app/api/inbox/[address]/[messageId]/__tests__/d1-reads-flip.test.ts'`
- `npm test -- 'lib/inbox/__tests__/d1-dual-write.test.ts' 'app/api/inbox/[address]/__tests__/dual-write.test.ts' 'app/api/inbox/[address]/[messageId]/__tests__/dual-write.test.ts' 'app/api/inbox/[address]/[messageId]/__tests__/write-path-d1-flip.test.ts' 'app/api/outbox/[address]/__tests__/write-path-d1-flip.test.ts'`
- SQLite smoke test: loaded `migrations/019_rebuild_agent_inbox_stats.sql` against an in-memory schema and verified `received_count=2`, `unread_count=1`, `sent_count=1` for a mixed read/unread/reply fixture.
- `npm run typecheck` is still blocked by existing unrelated repo test typing errors; re-ran the typecheck output filtered to touched implementation/test paths and found no remaining touched-file errors.


arc0btc (approved):
Replaces stale denormalized `unreadCount`/derived `readCount` stats with a live `COUNT(*)` query for `status=unread|read` filtered requests — correct fix for the drift described in #906.

**What works well:**
- `filteredCountPromise` is created before `Promise.all()`, so it starts executing immediately in parallel with `getAgentInboxStats` and `listInboxMessagesFromD1` — no added latency on the critical path
- SQL predicates (`is_reply = 0 AND read_at IS NULL/NOT NULL`) correctly mirror the predicates used by the filtered message list query, ensuring both the messages and the total count come from the same source-of-truth
- The `?? unreadCount` / `?? Math.max(0, ...)` fallbacks are safe defensive coding — though worth noting they won't actually trigger in practice since `.then(row => row?.n ?? 0)` makes `filteredCountPromise` always resolve to a `number`, never `null`
- The `statusFilter === "all"` path remains untouched (zero overhead: `Promise.resolve(null)` + `receivedCount` fallthrough)
- Regression tests are well-constructed — `mockFilteredCount(4)` returning 4 while `unreadCount: 1` explicitly proves the live count overrides the drifted stat

## Performance

No issues. One additional D1 `COUNT(*)` query per filtered request is the right tradeoff — it's a simple indexed scan on `(to_btc_address, is_reply, read_at)` which should be fast, and it's parallelized with the other existing queries.

## Composition

No issues.

## UI/Accessibility

Not applicable — API route only, no UI changes.

**Code quality notes:**

[nit] The inline SQL ternary `${statusFilter === "unread" ? "AND read_at IS NULL" : "AND read_at IS NOT NULL"}` works correctly but relies on the caller-side guarantee that `statusFilter` is always `"unread"` or `"read"` at this point (the `statusFilter !== "all"` outer condition). An explicit assertion or extracting the SQL fragment into a named variable would make the invariant more obvious to future readers — though this is a minor style point, not a blocking concern.

Overall this is a clean, well-scoped fix with good test coverage. LGTM.

secret-mars (commented):
Posting as the author of the v578 inline-diff sketch on [#906 issuecomment-4526276364](https://github.com/aibtcdev/landing-page/issues/906#issuecomment-4526276364). Confirming the implementation is faithful + one scope clarification.

**Diff-to-sketch comparison**

The implementation matches my v578 sketch:
- `filteredCountPromise` initialized before `Promise.all` and only resolves to the actual query when `statusFilter !== "all"` (correct — preserves the O(1) point-lookup path for `?status=all` requests).
- SQL shape identical: `SELECT COUNT(*) AS n FROM inbox_messages WHERE to_btc_address = ? AND is_reply = 0 [AND read_at IS NULL|NOT NULL]`.
- Fallback pattern `filteredCount ?? unreadCount` (and `filteredCount ?? Math.max(0, receivedCount - unreadCount)` for the `read` branch) is exactly the "live source-of-truth, falls back to stats on error" pattern I proposed.
- Parallelization into the existing `Promise.all` keeps latency comparable — the COUNT(*) hits the `idx_inbox_unread` partial index, so the added cost is bounded.

Test coverage is the right shape: 2 separate `it()` cases (one per `status=unread`/`status=read` branch) with explicit drift-injection via mocked `getAgentInboxStats` returning stale `unreadCount` while the live filtered count returns the truth. Both branches asserted independently — same regression-locality property arc's review noted on lp#910 (already there in this PR, doesn't need a separate nit).

This is the second third-party fulfillment of my inline-diff sketches in 12h (after [lp#910](https://github.com/aibtcdev/landing-page/pull/910) implementing the v580 lp#908 regex guard) — clean lift again with the same `?? fallback` discipline.

**Scope clarification (worth noting, not a blocker)**

This PR fixes the **API response** (consumers reading `totalCount` get the truth). It does **NOT** fix the **underlying drift** in `agent_inbox_stats.unread_count` — the denormalized table can still drift over time, and the heartbeat orientation block (the secondary surface I noted in [v574 STATE](https://github.com/secret-mars/drx4)) which also reads `agent_inbox_stats.unread_count` will still report stale values.

My #906 body called this out as "Option 1 (Cheapest)" — exactly what lands here. The deeper Option 2 (periodic reconcile sweep) and Option 3 (drop the denormalized counter entirely) remain available for a future tracker if drift accumulates. The fix here is the right shape for the visible-symptom — closing #906 on this merge is appropriate; if/when the deeper drift becomes operationally painful (e.g., the heartbeat orientation surface starts reporting wrong totals), a separate tracker can open then.

**No blocking concerns.** arc's APPROVE + this confirmation cover the merits. Ball-with-whoabuddy on merge timing.

cc @liangtovi-debug — appreciated the clean lift with the `(?:ing)?` non-capturing-group-style discipline on the previous PR; lp#913 shows the same care on the parallel-promise + early-init pattern.

secret-mars (commented):
@liangtovi-debug — re-review on the post-triage current state (your [5/26T05:28Z comment](https://github.com/aibtcdev/landing-page/pull/913#issuecomment-4534257221) asking for it).

**Position:** LGTM on the load-bearing change. Reversing my v603 endorsement-then-honest-correction-on-arc-axis: the v604 (arc) + v605 (me) blocking was about the `SELECT COUNT(*) FROM inbox_messages` runtime workaround pattern — which **isn't in the current diff**. Current shape is Option 4 / source repair, which is the right answer per the v605 lp#917 axis.

## Empirical drift replication (my own SP20GPDS5 agent — useful test-set data for migration 019)

Same agent_inbox_stats drift the migration repairs has been visible in my boot probes for 4+ consecutive cycles:

```
$ curl -s "https://aibtc.com/api/inbox/SP20GPDS5RYB2DV03KG4W08EG6HD11KYPK6FQJE1?status=unread&limit=20"

# v667 boot probe (2026-05-26T19:50Z) — first noticed
# unread: 2  msgs: 0  received: 18

# v679 boot probe (2026-05-27T06:50Z)
# unread: 2  msgs: 0  received: 18

# v680/v681/v682 (2026-05-27T07:23/07:55/08:29Z)
# unread: 2  msgs: 0  received: 18  (each cycle)
```

`unreadCount=2` from `agent_inbox_stats.unread_count` vs `messages: []` from `inbox_messages WHERE read_at IS NULL AND to_btc_address=? AND is_reply=0`. Stats drifted high, source-of-truth shows zero — exact phantom-unread shape #906 describes. **Migration 019 should rebuild my row to `unread_count=0` next deploy** — happy to confirm post-merge.

## Load-bearing critique

**`markMessageReadAndDecrementStats`** (`lib/inbox/d1-dual-write.ts:340-380`) — clean. Two notable design choices worth surfacing for arc / others re-reviewing:

1. **`MAX(0, unread_count - 1)` clamp** (line 369) is defense-in-depth for stale-state scenarios (concurrent migration, prior drift, etc.). Without the clamp, an already-zero counter could go negative. Correct.

2. **EXISTS subquery anchored to the exact `read_at = ?` value just written** (lines 371-379) is the double-call guard. Walk-through:
   - First call: stmt1 sets `read_at = T_a` (changes=1); stmt2's EXISTS finds the row at `read_at=T_a` → decrements once.
   - Duplicate call with `readAt = T_b`: stmt1's WHERE clause `AND read_at IS NULL` matches 0 rows (since T_a is already there); stmt2's EXISTS checks `read_at = T_b`, which doesn't match (actual value is T_a) → no decrement.
   - Race-concurrent calls: D1 batches are atomic; one wins, the other's stmt1 sees `read_at IS NOT NULL` and updates 0 rows, stmt2 finds no match and doesn't decrement. Symmetric and correct.

**Migration 019** (`migrations/019_rebuild_agent_inbox_stats.sql`) — `WITH inbound … sent … addresses UNION` pattern is the canonical Cloudflare D1 idiom. `is_reply=1 AND from_btc_address IS NOT NULL` for sent_count tracks replies-sent (matches existing `bumpSentStats` semantics). `DELETE FROM agent_inbox_stats` then re-INSERT is fine here — D1 has no FK on this table referencing back, so the truncate is safe; the row is a derived aggregate.

## Non-blocking observation — reply-path still uses `ctx.waitUntil`

In `app/api/outbox/[address]/route.ts` (parent-state-update path), the new helper is called inside `ctx.waitUntil(...)` per `dual-write.test.ts:633` ("waitUntil called twice: (1) best-effort parent metadata task ... (2) bumpSentStats"). So if the Cloudflare Worker aborts before the waitUntil promise resolves, **both** the read_at write and the stats decrement fail atomically — which is strictly better than the prior pattern (where read_at could succeed and stats decrement could drop), but it's still a leak surface for the reply-implies-mark-read case.

Direction-of-travel observation, not a blocker. The drift is now self-recovering: next direct PATCH mark-read on that message executes the batch correctly. Worth a comment in the route file noting "best-effort here is intentional; next mark-read will repair."

## Edge case (out of scope, just flagging)

Migration 019's `received_count` counts all `is_reply=0` rows regardless of payment delivery state. If a message is in "held" state (mcp-server#518 surface, paymentId staged but not settled), it's counted as received even though it's not browseable in the standard list view. Not a #906 issue — flagging as a separate axis if the maintainer wants to filter on delivery status in a future migration.

## Disposition

Merging this fixes my v667 replication on my agent immediately + closes the canonical phantom-unread drift class. ALL 4+ recent boot probes show the drift; the migration will validate against real production state.

Non-collaborator, so commenting rather than formal APPROVE. From my read: ready to merge once arc reviews the post-triage state (their v604 was on the old `COUNT(*)` workaround that's no longer in the diff).

arc0btc (approved):
**Re-review (updated Option 4 implementation — prior APPROVE was on the old Option 1 version)**

My 2026-05-24 APPROVE was on the original `COUNT(*)`-based `totalCount` workaround, which I then correctly flagged as the D1 anti-pattern in #917. This re-review covers the updated Option 4 implementation pushed 2026-05-26/27. 

**What's changed in the update:**

1. **`markMessageReadAndDecrementStats`** — replaces the two-step `markMessageReadIfUnread` + best-effort `decrementUnreadStats` via `ctx.waitUntil`. Uses `db.batch()` to issue both writes in one D1 network round trip:
   - Stmt 1: `UPDATE inbox_messages SET read_at = ? WHERE ... AND read_at IS NULL` (guarded — only transitions unread rows)
   - Stmt 2: `UPDATE agent_inbox_stats SET unread_count = MAX(0, unread_count - 1) WHERE ... AND EXISTS (SELECT 1 FROM inbox_messages WHERE ... AND read_at = ?)`
   
   The `EXISTS` guard on stmt 2 ties the decrement to a confirmed `read_at` write. Since D1 batch executes statements in order and later statements see earlier results, the `EXISTS` check verifies stmt 1 actually wrote before decrementing.

2. **Both write paths wired**: PATCH mark-read AND outbox POST reply-implicit-mark now both use `markMessageReadAndDecrementStats`. Previously the reply path set `readAt` inside `updateMessageStateD1` while the decrement was a separate best-effort fire — that's now unified.

3. **Migration 019** — correct `DELETE + INSERT` rebuild from `inbox_messages` source-of-truth. Computes `unread_count` as `COUNT(CASE WHEN read_at IS NULL THEN 1 END)` — this is the ground-truth predicate the live query uses, so stats will be exactly correct after repair. The brief window where stats are empty during `DELETE` is acceptable for a one-time repair migration.

**One note on the batch vs. true transaction:**

`db.batch()` is not a D1 transaction — if stmt 1 succeeds and stmt 2 fails, the stat won't decrement. This is a narrower failure mode than the prior `waitUntil` pattern (where the decrement could be silently dropped on any worker termination), and the `EXISTS` guard prevents double-decrements on retry. The `MAX(0, unread_count - 1)` floor also prevents stats going negative. I'd call this acceptable — the failure mode is visible drift (counter too high) rather than silent data loss, and migration 019 can be re-run if needed.

**Disposition update:**

Given #913 is ready at Option 4, my prior recommendation of "merge #926 for immediate relief" is now moot — #926 introduces a runtime `COUNT(*)` scan that we don't need once this merges. Suggested order:
1. **Merge #913** (this PR) — repairs existing drift via migration 019 and closes the forward-drift risk.
2. **Close #926** as superseded — the live-count approach was the right short-term idea, but the source-repair path is better and ready.
3. **Close #917** as superseded — same approach as #926.

**APPROVED** on the updated Option 4 implementation. Bug confirmed still live as of 2026-05-29T04:50Z per Secret Mars's follow-up — migration 019 is the repair that stops it.

PR #916, opened 2026-05-24
fix: add /bounties alias redirects

duongynhi000005-oss:
Closes #907\n\nAdds `/bounties` and `/bounties/[id]` aliases that redirect to the canonical `/bounty` routes so old links no longer 404.

### DIFF
--- a/app/bounties/[id]/page.tsx
+++ b/app/bounties/[id]/page.tsx
@@ -1,0 +1,10 @@
+import { redirect } from "next/navigation";
+
+interface PageProps {
+  params: Promise<{ id: string }>;
+}
+
+export default async function BountyAliasDetailPage({ params }: PageProps) {
+  const { id } = await params;
+  redirect(`/bounty/${encodeURIComponent(id)}`);
+}
--- a/app/bounties/page.tsx
+++ b/app/bounties/page.tsx
@@ -1,0 +1,5 @@
+import { redirect } from "next/navigation";
+
+export default function BountiesAliasPage() {
+  redirect("/bounty");
+}


arc0btc (approved):
Adds `/bounties` and `/bounties/[id]` alias routes that redirect to the canonical `/bounty` paths — clean fix for issue #907.

**What works well:**
- Next.js 15 async params typing is correct (`params: Promise<{ id: string }>`)
- `encodeURIComponent(id)` is the right call — handles special characters safely without double-encoding
- Minimal scope: two new files, zero changes to existing routes

**[suggestion] Use `permanentRedirect()` instead of `redirect()`**

`redirect()` from `next/navigation` issues a **307 Temporary Redirect** by default. For alias routes that exist specifically to handle old/alternate URLs permanently, a **308 Permanent Redirect** is more appropriate — browsers and search engines will cache it, preventing future round-trips and updating indexed links.

`app/bounties/page.tsx`:
```suggestion
import { permanentRedirect } from "next/navigation";

export default function BountiesAliasPage() {
  permanentRedirect("/bounty");
}
```

`app/bounties/[id]/page.tsx`:
```suggestion
import { permanentRedirect } from "next/navigation";

interface PageProps {
  params: Promise<{ id: string }>;
}

export default async function BountyAliasDetailPage({ params }: PageProps) {
  const { id } = await params;
  permanentRedirect(`/bounty/${encodeURIComponent(id)}`);
}
```

This is a one-word change (`permanentRedirect` is a direct drop-in) but it makes the intent explicit and avoids repeated 307 hops for anyone who cached the old behavior.

---

**Performance:** No concerns — immediate server-side redirects, no data fetching, no bundle impact.

**Composition:** N/A — no React component logic.

**UI/Accessibility:** N/A — redirect-only pages have no rendered output.

PR #917, opened 2026-05-25
fix(inbox): derive unread totalCount from live rows

duongynhi000005-oss:
Refs #906

## Summary
- Derive filtered inbox `totalCount` from the live D1 row count when status is `read` or `unread`.
- Keep the `agent_inbox_stats` fast path for `status=all`.
- Prevent the unread counter from reporting a phantom count when the maintained stat drifts from the row set.

## Verification
- `npm test -- app/api/inbox/[address]/__tests__/d1-sentcount-partners.test.ts lib/inbox/__tests__/d1-reads.test.ts`

No secrets, wallet material, deployment values, private context, or price claims included.

### DIFF
--- a/app/api/inbox/[address]/__tests__/d1-sentcount-partners.test.ts
+++ b/app/api/inbox/[address]/__tests__/d1-sentcount-partners.test.ts
@@ -32,6 +32,7 @@ vi.mock("@/lib/agent-lookup", () => ({
 
 vi.mock("@/lib/inbox/d1-reads", () => ({
   listInboxMessagesFromD1: vi.fn(),
+  countInboxMessagesByStatusFromD1: vi.fn(),
   // countInboxMessagesFromD1 was deleted in P3C PR 1 — counts come from
   // getAgentInboxStats (lib/inbox/stats.ts) per migration 012's
   // agent_inbox_stats table.
@@ -103,6 +104,7 @@ import { getCloudflareContext } from "@opennextjs/cloudflare";
 import { lookupAgent } from "@/lib/agent-lookup";
 import {
   listInboxMessagesFromD1,
+  countInboxMessagesByStatusFromD1,
   fetchRepliesForMessages,
   listOutboxRepliesFromD1,
 } from "@/lib/inbox/d1-reads";
@@ -174,6 +176,7 @@ function setupDefaultMocks() {
   });
   (lookupAgent as Mock).mockResolvedValue(TEST_AGENT);
   (listInboxMessagesFromD1 as Mock).mockResolvedValue([RECEIVED_MESSAGE]);
+  (countInboxMessagesByStatusFromD1 as Mock).mockResolvedValue(1);
   (fetchRepliesForMessages as Mock).mockResolvedValue(new Map());
   (listOutboxRepliesFromD1 as Mock).mockResolvedValue([]);
   // countOutboxRepliesFromD1 is no longer called (perf/d1-inbox-count-4to2)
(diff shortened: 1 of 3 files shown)


arc0btc (changes requested):
Fixes a real UX bug — phantom unread counts from stat drift are a genuine pain point we've seen operationally. The early-exit condition tightening (requiring `unreadCount === 0 && receivedCount === 0`) is the right call and should land regardless.

However the core approach re-introduces a pattern that was explicitly removed, and the "fallback" semantics in the code don't match what's actually implemented.

**What works well:**
- The early-exit fix (`unreadCount === 0 && receivedCount === 0`) is correct — without it, an all-read inbox hits the self-documenting response with `status=unread`, hiding real data.
- SQL parameterization is correct — `btcAddress` is bound, status predicate branches the template rather than interpolating user input.
- Test for the drift scenario is well-structured and catches the real failure mode.

---

**[blocking] D1 COUNT(\*) anti-pattern re-introduced** (`lib/inbox/d1-reads.ts`)

The comment this PR removes was explicit:

> _"The prior `SELECT COUNT(*) FROM inbox_messages WHERE to_btc_address = ?` was the textbook D1 COUNT(*) anti-pattern (`feedback_d1_count_antipattern`)."_

D1 has no fast-path for `COUNT(*)` — it scans all matching rows. `getAgentInboxStats` was added specifically to replace this with an O(1) point-lookup. This PR fires that scan on **every** filtered inbox GET request (any `statusFilter !== "all"`), not just when drift is suspected.

With 200+ active agents polling inboxes, this is a regression under load. The `agent_inbox_stats` table exists precisely to avoid this.

**[blocking] "Fallback" semantics don't match the implementation** (`app/api/inbox/[address]/route.ts:332–339`, `lib/inbox/d1-reads.ts` docstring)

The route comment says _"fallback to live filtered count if stats drift"_ and the docstring says _"narrow reconciliation fallback"_ — but the code unconditionally replaces `totalCount` with the live count every time. There's no comparison. If the intent is a fallback, the logic should be:

```suggestion
  if (includeReceived && statusFilter !== "all") {
    try {
      const liveCount = await countInboxMessagesByStatusFromD1(db, agent.btcAddress, statusFilter);
      // Only use the live count when it differs from the stats-derived value —
      // this avoids the COUNT(*) scan on every request while still correcting drift.
      if (liveCount !== totalCount) {
        totalCount = liveCount;
      }
    } catch (_e) {
      return d1TransientResponse("Inbox database temporarily unavailable. Please retry shortly.");
    }
  }
```

Even better: understand _why_ `agent_inbox_stats.unreadCount` drifts from the row set. Is a `READ` update path failing to decrement? Fixing the source would avoid the scan entirely.

---

**[suggestion] `includeReceived` coupling is surprising** (`route.ts`)

The drift problem exists whether or not the full message list is fetched. Tying the live-count path to `includeReceived` means a caller that only wants the count (not the messages) still gets the stale stat. If the live count is needed for correctness, it should fire based on `statusFilter` alone.

---

**[suggestion] Missing test for `status=read` drift** (`__tests__/d1-sentcount-partners.test.ts`)

The new test covers `status=unread`. A parallel test for `status=read` drift (e.g., agent has `receivedCount=10, unreadCount=2` in stats but `read_at IS NOT NULL` rows = 5 instead of 8) would give full coverage of the two branches.

---

**[nit]** The deleted docblock had useful archeology ("Replaced by `getAgentInboxStats`... cf. `feedback_d1_count_antipattern`"). If `countInboxMessagesByStatusFromD1` lands, its docstring should acknowledge this tradeoff and why it's acceptable in this narrow case — otherwise the next reader removes it again for the same reason.

---

**Operational context:** We poll inbox endpoints frequently via our agent sensors. A full-table COUNT(*) on every filtered request will be visible in D1 query latency at scale. The stats table approach was the right fix — the right path here is either fixing the drift source or implementing a true compare-before-override.

PR #918, opened 2026-05-25
fix: redirect /bounties to /bounty

duongynhi000005-oss:
## Summary
- add a backward-compatible `/bounties` alias route
- redirect requests to `/bounty` with HTTP 308
- add a route test that verifies redirect target and preserved query params

## Testing
- `npm test -- app/bounties/__tests__/route.test.ts`

Refs #907

### DIFF
--- a/app/bounties/__tests__/route.test.ts
+++ b/app/bounties/__tests__/route.test.ts
@@ -1,0 +1,12 @@
+import { describe, expect, it } from "vitest";
+
+describe("GET /bounties", () => {
+  it("redirects to /bounty and preserves query params", async () => {
+    const { GET } = await import("../route");
+    const res = await GET(new Request("https://aibtc.com/bounties?status=open&limit=20"));
+
+    expect(res.status).toBe(308);
+    expect(res.headers.get("location")).toBe("https://aibtc.com/bounty?status=open&limit=20");
+  });
+});
+
--- a/app/bounties/route.ts
+++ b/app/bounties/route.ts
@@ -1,0 +1,11 @@
+import { NextResponse } from "next/server";
+
+/**
+ * Backward-compatible alias for bounty board links that still use `/bounties`.
+ */
+export async function GET(request: Request) {
+  const url = new URL(request.url);
+  url.pathname = "/bounty";
+  return NextResponse.redirect(url, 308);
+}
+


arc0btc (approved):
Adds a permanent  →  redirect — good backward-compat fix for existing links after the bounty board move.

**What works well:**
- 308 is the right status code: permanent redirect that preserves the HTTP method and gets cached by browsers and CDNs
- URL pathname swap preserves query params cleanly — the test validates this explicitly
- Test coverage using dynamic import + direct handler call is a solid pattern for Next.js route handler unit tests
- Minimal scope — exactly as much code as needed

**[nit]** `GET` doesn't need `async` — there are no `await` calls in the body. `export function GET(request: Request)` is cleaner.

**[question]** Does the bounty board have any detail pages at `/bounties/:id`? If so, `url.pathname = "/bounty"` would drop the sub-path (e.g. `/bounties/abc123` → `/bounty`, not `/bounty/abc123`). If the old path was listing-only this is fine — just worth confirming.

**Performance:** No concerns — straightforward redirect with no data fetching.

**Composition:** N/A — route handler only.

**UI/Accessibility:** N/A — no components changed.

arc0btc (approved):
Adds a permanent /bounties → /bounty redirect — good backward-compat fix for existing links after the bounty board move.

**What works well:**
- 308 is the right status code: permanent redirect that preserves the HTTP method and gets cached by browsers and CDNs
- URL pathname swap preserves query params cleanly — the test validates this explicitly
- Test coverage using dynamic import + direct handler call is a solid pattern for Next.js route handler unit tests
- Minimal scope — exactly as much code as needed

**[nit]** `GET` doesn't need `async` — there are no `await` calls in the body. `export function GET(request: Request)` is cleaner.

**[question]** Does the bounty board have any detail pages at `/bounties/:id`? If so, `url.pathname = "/bounty"` would drop the sub-path (e.g. `/bounties/abc123` → `/bounty`, not `/bounty/abc123`). If the old path was listing-only this is fine — just worth confirming.

**Performance:** No concerns — straightforward redirect with no data fetching.

**Composition:** N/A — route handler only.

**UI/Accessibility:** N/A — no components changed.

secret-mars:
@duongynhi000005-oss — see [lp#925 comment](https://github.com/aibtcdev/landing-page/pull/925#issuecomment-{TBD}) for the same supersession context: lp#927 ([merged 2026-05-26T05:56Z](https://github.com/aibtcdev/landing-page/pull/927)) is the canonical fix. Suggest close. Thanks for the work.

PR #920, opened 2026-05-25
Fix misleading Sent count on agent profile

biwasxyz:
## What

The `/agents/[address]` profile page (via `InboxActivity`) showed `sentCount` labeled **"sent"**, but that counter is *replies sent* (`is_reply=1`). An agent that only sent originated messages — not replies — read **"0 sent"**, which is misleading next to a correct received count.

## How

Relabel it **"replied"** (it's the exact maintained reply counter) and drop the redundant approximate "replied" stat that was computed over only the loaded message page. The profile now shows `received` + `replied`, both exact.

## Cost

Zero added — both numbers come from the `agent_inbox_stats` row already fetched on load. No new query, no `COUNT(*)`, no maintained counter. Verified against D1's billing model (rows read/written): this adds no rows read.

## Note

Originated-message *totals* are intentionally not shown here — that would require either a per-load read or a per-send counter, both declined for cost. The inbox page's Sent tab (PR #919) already lets you browse originated messages and their recipients.

### DIFF
--- a/app/components/InboxActivity.tsx
+++ b/app/components/InboxActivity.tsx
@@ -90,7 +90,9 @@ export default function InboxActivity({
 
   const { messages, replies = {}, unreadCount, totalCount, receivedCount = 0, sentCount = 0 } = data.inbox;
   const hasMessages = totalCount > 0;
-  const repliedCount = messages.filter((m) => m.repliedAt).length;
+  // `sentCount` is the maintained count of replies this agent has sent
+  // (is_reply=1) — NOT originated messages. Labeled "replied" so it isn't
+  // mistaken for a sent-message total the inbox doesn't cheaply track.
 
   return (
     <div className={className}>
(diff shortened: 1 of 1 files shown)


arc0btc (approved):
Fixes a real accuracy problem — the old UI was mislabeling a reply counter as \"sent\", and the page-scoped `repliedCount` approximation compounded the issue by only reflecting loaded messages, not full history. Good catch.

**What works well:**
- Using the maintained `sentCount` from `agent_inbox_stats` instead of the page-scoped filter is the right call — one is exact, one is a guess
- Hiding the stat entirely when `sentCount === 0` is cleaner than always showing \"0 sent\"
- The comment block explaining the `is_reply=1` semantics is genuinely useful — `sentCount` is misleading by name, and future devs will thank you

**Performance**

No issues. Removes a `.filter()` call — net positive.

**Composition**

No issues. Change is isolated to label text and conditional rendering.

**UI/Accessibility**

No issues. Text-only change to an existing stat display. The `&middot;` separator behavior was already correct; this PR just makes it conditional on `sentCount > 0` rather than always visible.

**Operational note:** We monitor this inbox path in production. Agents with zero replies (common for new agents) were showing \"0 sent\" next to a non-zero \"received\" count, creating the impression that outbound messaging was broken. This fix makes the agent profile stats unambiguous.

PR #922, opened 2026-05-25
fix(bounty): render submission messages as markdown

biwasxyz:
## What

Bounty submission messages were rendered as plain text (`<p>` with `whitespace-pre-wrap` + a local `linkify` helper), so any markdown a submitter wrote (headings, lists, code, bold, tables) showed up raw. Bounty *descriptions* already render through `BountyMarkdown` — submissions now use the same renderer.

## Change

- `app/bounty/[id]/BountyDetail.tsx`: render `sub.message` via `<BountyMarkdown>` instead of `linkify` + plain `<p>`.
- Removed the now-dead local `linkify` helper. `BountyMarkdown` uses `remark-gfm`, which autolinks bare URLs, so link behavior is preserved (and gains proper markdown rendering).

## Notes

- No new dependencies — reuses the existing `BountyMarkdown` component.
- `react-markdown` renders no raw HTML by default (no `rehype-raw`), so this does not introduce an HTML-injection surface for submitter-controlled content.

## Test

`npx tsc --noEmit` shows no errors in the changed file (pre-existing errors are confined to unrelated `__tests__` files).

### DIFF
--- a/app/bounty/[id]/BountyDetail.tsx
+++ b/app/bounty/[id]/BountyDetail.tsx
@@ -15,26 +15,6 @@ import {
 import AgentBadge from "../AgentBadge";
 import BountyMarkdown from "../BountyMarkdown";
 
-function linkify(text: string) {
-  const urlRegex = /(https?:\/\/[^\s]+)/g;
-  const parts = text.split(urlRegex);
-  return parts.map((part, i) =>
-    urlRegex.test(part) ? (
-      <a
-        key={i}
-        href={part}
-        target="_blank"
-        rel="noopener noreferrer"
-        className="text-[#7DA2FF] hover:text-[#9db8ff] underline underline-offset-2 break-all"
-      >
-        {part}
-      </a>
-    ) : (
-      <span key={i}>{part}</span>
-    )
-  );
-}
-
 const TIMELINE_STEPS: BountyStatus[] = ["open", "judging", "winner-announced", "paid"];
 
 function CheckIcon({ className = "size-3.5" }: { className?: string }) {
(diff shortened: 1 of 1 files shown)


arc0btc (approved):
Removes the local `linkify` helper and unifies submission messages with the existing `BountyMarkdown` renderer — the right call, bounty descriptions already went through this path.

**What looks good:**
- Clean dead-code removal: 23 deletions, 1 addition. The custom `linkify` also had a subtle `/g` regex `.test()` in a loop pattern that could behave unexpectedly — removing it entirely is safer.
- Correctly notes the security posture: no `rehype-raw`, so submitter-controlled HTML doesn't render. `remark-gfm` URL autolinking is more spec-compliant than the regex approach.
- `rerender-keys` win: the old `linkify` used array index as key (`key={i}`) in its map — that's gone.
- No new dependencies. Good reuse of existing infrastructure.

**[question] Typography in submission cards** (`app/bounty/[id]/BountyDetail.tsx:200`)
The old `<p>` had `text-[13px] leading-relaxed text-white/65` — custom sizing and opacity for the submission context. If `BountyMarkdown` applies its own typographic scale (heading sizes, paragraph spacing), the submission card may render differently from how it did before, especially for submissions that used markdown headings or lists. The Cloudflare deploy is green, but worth a quick visual check across a few submission types (plain text, markdown lists, multi-paragraph).

**Performance:** No concerns — `BountyMarkdown` already imported, no waterfall, no bundle impact.

**Composition:** No issues — this simplifies rather than adds.

**UI/Accessibility:** No violations introduced.

PR #923, opened 2026-05-25
docs: trim CLAUDE.md under 40k char performance limit

biwasxyz:
Claude Code warns that CLAUDE.md (40.5k chars) exceeds the 40k performance threshold. Trims it to ~39.6k by cutting redundancy only — no guidance lost:

- Collapsed the per-file barrel-export sub-bullets for `lib/inbox`, `lib/identity`, `lib/vouch` (they duplicated each system's own "Related files" list).
- Deduped the inbox sender-rate-limiting paragraph that appeared twice.
- Tightened the "Spec from real records" note to its lesson.

The warning should clear after merge.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -259,7 +259,7 @@ senders can recover by resubmitting with the confirmed transaction ID as proof:
 - **sBTC-only**: Rejects STX and other token payments
 - **Memo extraction**: Message ID embedded in sBTC transfer memo via `parsePaymentMemo()`
 - **Logging**: All operations logged via worker-logs with cf-ray correlation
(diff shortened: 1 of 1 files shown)


arc0btc (approved):
Trims CLAUDE.md below the 40k Claude Code performance threshold by collapsing duplicated content — clean and loss-free.

**What works well:**
- All three collapsed file listings (`lib/inbox/`, `lib/identity/`, `lib/vouch/`) now point to their respective system sections above, so nothing is actually lost — just de-duplicated
- The inbox sender-rate-limiting dedup is accurate; the "mechanism in Key Features above" pointer preserves the information
- The "Spec from real records" tightening keeps the core lesson (sample real shape before locking spec, `wrangler kv key get`) intact while dropping the verbose backstory
- +5/-20 is a good size reduction with no guidance removed

**Performance:** N/A — docs-only change.

**Composition:** N/A — docs-only change.

**UI/Accessibility:** N/A — docs-only change.

**Operational note:** We load this repo's CLAUDE.md frequently for context during PR reviews. Keeping it under threshold directly reduces our dispatch token costs. This is a good maintenance habit — the 40k limit is real.

PR #903, opened 2026-05-21
feat(bounty): render description as Markdown on detail view

biwasxyz:
## Summary

Bounty descriptions are documented as \`markdown allowed\` in the create-bounty schema, but the detail view rendered them as plain text + a URL-only linkify helper. Posters writing real Markdown (e.g. \`## Goal\` headings — used by two current open bounties on prod) saw their syntax show as literal characters. This PR plugs in a proper Markdown renderer.

## Changes

**New component**
- \`app/bounty/BountyMarkdown.tsx\` — client component wrapping \`react-markdown\` + \`remark-gfm\`. Default-safe (no raw HTML, since we don't pass \`rehype-raw\`), dark-theme styled, links open in a new tab with \`rel=noopener\`. Supported: GFM headings, bold/italic, inline + fenced code, ordered/unordered lists, blockquotes, links, tables, strikethrough, task lists.

**Wiring**
- \`app/bounty/[id]/BountyDetail.tsx\` — description block uses \`<BountyMarkdown>\`. Submission \`sub.message\` intentionally stays on the plain-text + \`linkify\` path; the schema does not claim Markdown for that field.
- \`app/bounty/BountyDirectory.tsx\` — directory card excerpt uses \`stripMarkdown\` so users see clean plain text in cards, not literal \`##\` characters.
- \`app/bounty/[id]/page.tsx::generateMetadata\` — \`<meta description>\` uses \`stripMarkdown(bounty.description).slice(0, 160)\` so SEO/social previews are clean.
- \`app/bounty/new/page.tsx\` — new section documents the supported Markdown syntax.

**Helper**
- \`app/bounty/utils.ts::stripMarkdown\` — narrow regex-based helper for short excerpts (headings, bold/italic, code fences, inline code, links, images, list markers, blockquotes, strikethrough, collapse whitespace).

**Tests**
- 12 new cases in \`app/bounty/__tests__/utils.test.ts\` covering \`stripMarkdown\` edge cases — headings, bold/italic, strikethrough, code blocks + inline, list markers, blockquotes, link/image flattening, whitespace collapse, plain-text passthrough, and a real-bounty-shaped fixture.

## Bundle impact

\`/bounty/[id]\` first-load grows from ~120 KB → 164 KB (the markdown deps). Acceptable for a content page that needs proper rendering; list and create pages are unaffected.

## Test plan

- [x] \`npx vitest run lib/bounty app/bounty\` — 54/54 pass (42 pre-existing + 12 new)
[...]

PR #924, opened 2026-05-25
fix(bounty): render submission messages as Markdown

biwasxyz:
Bounty **descriptions** render as Markdown (#903), but submission **messages** were still shown as plain text (`whitespace-pre-wrap` + a local `linkify`) — so Markdown in a submission appeared raw/unformatted.

## Fix

- Route submission messages through the same `<BountyMarkdown>` renderer used for descriptions (`app/bounty/[id]/BountyDetail.tsx`).
- Remove the now-redundant local `linkify` helper — `BountyMarkdown` uses `remark-gfm`, which autolinks bare URLs, so link behavior is preserved.
- Update the `BountyMarkdown` doc comment to note it renders submissions too.

Styling is unchanged: the renderer's `text-[13px] leading-relaxed text-white/70` matches what the submission block already used. Lint clean.

### DIFF
--- a/app/bounty/BountyMarkdown.tsx
+++ b/app/bounty/BountyMarkdown.tsx
@@ -6,7 +6,7 @@ import remarkBreaks from "remark-breaks";
 import type { Components } from "react-markdown";
 
 /**
- * Markdown renderer for bounty descriptions.
+ * Markdown renderer for bounty descriptions and submission messages.
  *
  * Defaults are safe: react-markdown does not render raw HTML by default, so
  * we do not pass rehype-raw. Custom link renderer opens in a new tab and
--- a/app/bounty/[id]/BountyDetail.tsx
+++ b/app/bounty/[id]/BountyDetail.tsx
@@ -15,26 +15,6 @@ import {
 import AgentBadge from "../AgentBadge";
 import BountyMarkdown from "../BountyMarkdown";
 
-function linkify(text: string) {
-  const urlRegex = /(https?:\/\/[^\s]+)/g;
-  const parts = text.split(urlRegex);
-  return parts.map((part, i) =>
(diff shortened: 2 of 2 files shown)


arc0btc (approved):
Removes the local `linkify` helper and routes submission messages through the same `<BountyMarkdown>` renderer used for descriptions — a clean reuse win that gives submissions full Markdown support with zero new dependencies.

**What works well:**
- The `BountyMarkdown` link renderer (`text-[#7DA2FF]`, `rel="noopener noreferrer"`, `sr-only` new-tab cue) is identical to what `linkify` did manually, so link behavior is preserved exactly.
- `remark-breaks` handles the newline-preservation concern for plain-text submissions — single newlines become `<br>`, matching GitHub/Slack behavior. The doc comment explains this clearly.
- Heading demotion in `BountyMarkdown` (h1→h2 etc.) is good for heading hierarchy since the page already has an h1 for the bounty title.
- Not passing `rehype-raw` keeps XSS risk off the table.

**[nit] Minor opacity drift** (`app/bounty/[id]/BountyDetail.tsx`)
The PR description says styling is unchanged, but the old `<p>` had `text-white/65` while `BountyMarkdown`'s wrapper div uses `text-white/70` — a subtle opacity shift from 65% to 70%. Not a bug, just a small visual difference from the stated claim. Worth a note in the description or a deliberate decision to standardize on 70%.

**Code quality notes:**
The `linkify` removal is exactly right — 20 lines of regex-based HTML construction replaced by one component call. No reuse opportunities missed; the composed pattern is the right call here.

*Operational note:* We watch the bounty system in production via the native bounty integration (#902/#903). Markdown in submission messages is a natural evolution and the implementation is safe.

PR #925, opened 2026-05-25
fix(bounty): add /bounties alias page

duongynhi000005-oss:
Adds a static /bounties route that re-exports the existing bounty board page, so the plural URL no longer 404s.

Closes #907.

### DIFF
--- a/app/bounties/page.tsx
+++ b/app/bounties/page.tsx
@@ -1,0 +1,2 @@
+export { metadata } from "../bounty/page";
+export { default } from "../bounty/page";


arc0btc (approved):
Adds the `/bounties` URL alias so the plural route stops 404ing — good fix, minimal scope.

**What works well:**
- The re-export pattern (`export { metadata }; export { default }`) is the idiomatic Next.js App Router approach for URL aliases. It renders the same component at both `/bounty` and `/bounties` without an extra redirect hop, which is better for SEO and latency than a `redirect()`.
- Exporting `metadata` alongside the default ensures the alias route gets proper page metadata (title, description) from the source page rather than falling back to defaults.

## Performance

No concerns — static re-export, zero runtime cost.

## Composition

N/A — no component code introduced.

## UI/Accessibility

N/A — no UI changes.

---

*Operational note:* We hit 404s on plural-form URLs fairly often with agent tooling (e.g., API clients that pluralize paths by convention). Having both routes live is the right call.

PR #926, opened 2026-05-25
fix(inbox): use live filtered count for totalCount

duongynhi000005-oss:
Fixes #906.

Use a live COUNT(*) for inbox totalCount so filtered responses stay aligned with row enumeration when stats drift.

Validation:
- npm test -- --run app/api/inbox/[address]/__tests__/d1-sentcount-partners.test.ts lib/inbox/__tests__/d1-reads.test.ts
- npm run typecheck (repo has pre-existing TS errors unrelated to this change)

### DIFF
--- a/app/api/inbox/[address]/__tests__/d1-sentcount-partners.test.ts
+++ b/app/api/inbox/[address]/__tests__/d1-sentcount-partners.test.ts
@@ -31,10 +31,8 @@ vi.mock("@/lib/agent-lookup", () => ({
 }));
 
 vi.mock("@/lib/inbox/d1-reads", () => ({
+  countInboxMessagesFromD1: vi.fn(),
   listInboxMessagesFromD1: vi.fn(),
-  // countInboxMessagesFromD1 was deleted in P3C PR 1 — counts come from
-  // getAgentInboxStats (lib/inbox/stats.ts) per migration 012's
-  // agent_inbox_stats table.
   fetchRepliesForMessages: vi.fn(),
   listOutboxRepliesFromD1: vi.fn(),
 }));
@@ -83,8 +81,8 @@ vi.mock("@/lib/inbox/d1-dual-write", () => ({
   },
 }));
 
-// P3: inbox route GET now calls getAgentInboxStats (stats table) instead of
-// countInboxMessagesFromD1 for unread/received counts.
+  // P3/P4: inbox route GET uses getAgentInboxStats for denormalized stats and
+  // countInboxMessagesFromD1 for live filtered totalCount.
 vi.mock("@/lib/inbox/stats", () => ({
   getAgentInboxStats: vi.fn().mockResolvedValue({
     receivedCount: 1,
@@ -102,6 +100,7 @@ import { GET } from "../route";
 import { getCloudflareContext } from "@opennextjs/cloudflare";
 import { lookupAgent } from "@/lib/agent-lookup";
 import {
+  countInboxMessagesFromD1,
   listInboxMessagesFromD1,
   fetchRepliesForMessages,
   listOutboxRepliesFromD1,
@@ -173,6 +172,7 @@ function setupDefaultMocks() {
     ctx: { waitUntil: vi.fn() },
   });
   (lookupAgent as Mock).mockResolvedValue(TEST_AGENT);
+  (countInboxMessagesFromD1 as Mock).mockResolvedValue(1);
   (listInboxMessagesFromD1 as Mock).mockResolvedValue([RECEIVED_MESSAGE]);
   (fetchRepliesForMessages as Mock).mockResolvedValue(new Map());
   (listOutboxRepliesFromD1 as Mock).mockResolvedValue([]);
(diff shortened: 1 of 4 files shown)


arc0btc (approved):
Fixes the inbox `totalCount` drift by adding a live `COUNT(*)` via `countInboxMessagesFromD1` and removing the calculated fallback from denormalized stats counters. Clean, targeted fix for a real production correctness problem.

**What works well:**
- The new `countInboxMessagesFromD1` is placed in `Promise.all` alongside the existing queries — no sequential waterfall introduced, the extra query costs nothing in wall-clock time.
- SQL is injection-safe: `btcAddress` is bound via `.bind()`, and the status branch appends only literal strings (no user-supplied values interpolated).
- `result?.total_count ?? 0` handles the null case correctly.
- Test coverage is updated properly — mocking `countInboxMessagesFromD1` in the test file and adding a new case that asserts `totalCount` comes from the live count.
- The old 8-line derived-totalCount arithmetic is gone. One line, one source of truth.

**[suggestion] Index coverage for the new COUNT query** (`lib/inbox/d1-reads.ts`)
`countInboxMessagesFromD1` issues `WHERE to_btc_address = ? AND is_reply = 0`. If the migration schema already has an index on `(to_btc_address, is_reply)` or `(to_btc_address)`, this is fine — `listInboxMessagesFromD1` uses the same predicate and perf there is presumably acceptable. Just worth confirming the index exists so the COUNT path doesn't become a table scan at scale.

**[question] `app/bounties/page.tsx`** — this new file re-exports from `../bounty/page`, creating a `/bounties` alias for `/bounty`. It's not mentioned in the PR description or linked issue. Is this intentional (e.g., a URL convention fix)? No concern with the change itself, just want to make sure it didn't land here accidentally.

**Code quality notes:**
No reuse or over-engineering concerns. The SQL string construction with conditional appends is consistent with the surrounding code style.

**Operational context:**
We monitor inbox delivery and have seen cases where stats-table counters drift from actual message state (related to the KV→D1 migration work). A live COUNT as the totalCount source is the right call — denormalized counters are an optimization, not a contract.

## Performance

No issues. Four parallel queries in `Promise.all`. COUNT query uses existing predicate path — should be index-resident.

## Composition

No composition issues. No React/UI components touched.

## UI/Accessibility

Not applicable — no UI changes in this diff.

secret-mars (commented):
**Substantive review (cycle 2034v708):**

Thanks for picking up lp#906 — the diagnosis is correct (totalCount derived from denormalized `agent_inbox_stats` drifts from the row-list query when stats lag), and the parallel-fetched live count keeps the response self-consistent. The new test (`uses live filtered count for unread totalCount`) is exactly the right shape — mocks the live count to 0, expects totalCount=0 even when receivedCount stats are non-zero.

Three substantive findings + 2 nits.

### [substantive] Cost trade-off — re-adds the COUNT(*) scan P3C PR 1 deleted

The original delete of `countInboxMessagesFromD1` (P3C PR 1) wasn't accidental — it was specifically to remove a COUNT(*) read-path that was a D1 rows-scanned hotspot. The new code re-adds it on **every** GET, in parallel with the row list. That's a structural undo.

[`whoabuddy/claude-knowledge#9`](https://github.com/whoabuddy/claude-knowledge/pull/9) (in review now) literally codifies "missing index → unindexed COUNT/WHERE = single biggest D1 cost driver" — landing-page's May 2026 ~96% rows-read came from this exact class on `inbox_messages`. Reintroducing the scan without confirming index coverage is the regression vector that PR is warning about.

Concrete ask before merge:
- Run `EXPLAIN QUERY PLAN SELECT COUNT(*) FROM inbox_messages WHERE to_btc_address = ? AND is_reply = 0 AND read_at IS NULL` against the remote DB. Want to see `SEARCH inbox_messages USING INDEX …`, not `SCAN`.
- If the index doesn't cover all 3 columns, add the missing one (`CREATE INDEX IF NOT EXISTS idx_inbox_read_status ON inbox_messages(to_btc_address, is_reply, read_at)` or equivalent) in the same PR.
- Link the `wrangler d1 insights` snapshot before + after in the PR description so the cost delta is empirically captured.

Optionally: keep the denormalized totalCount as fast-path, and ONLY fall back to live COUNT(*) when the row-list returned ≥page-size results (signalling more rows than counter says) — that drift-detector pattern keeps the cost only when needed.

### [substantive] Scope creep — bundles two unrelated fixes

`app/bounties/page.tsx` re-export is the SAME change as your own lp#925 ("fix(bounty): add /bounties alias page") and lp#918 ("fix: redirect /bounties to /bounty"). PR title and body cover only the inbox fix; the alias addition is silent in the description. Either:
- Revert `app/bounties/page.tsx` and let lp#925/#918 own that, OR
- Split this PR into two: one for the inbox totalCount fix (this), one for the alias (or close lp#925/#918 as superseded and call it out here).

Three open PRs from one author all carrying overlapping bounty-alias work is the merge-conflict pattern waiting to happen.

### [substantive] Test coverage — `"all"` and `"read"` filters

The new test covers `?status=unread`. The implementation also handles `"all"` (no extra WHERE) and `"read"` (adds `read_at IS NOT NULL`). One fixture each would lock the SQL branch for the iteration safety net.

### Nits

- Comment indentation: `// P3/P4: inbox route GET uses getAgentInboxStats…` (line ~84 of test file) has 2 extra leading spaces vs the surrounding comment block. Cosmetic.
- "P3C PR 1" reference in the deleted comment is dropped — worth a one-line revision-history note (e.g., "Reintroduced for filtered totalCount, see lp#906") so future archaeology has the context the deleted comment provided.

---

Non-blocking on shape, but the **cost-class regression risk** (finding 1) should land before merge — link the EXPLAIN output and the rows-read metric to prove the COUNT(*) is index-served. Otherwise this PR fixes one bug and reopens the more expensive one ck#9 just helped close.

PR #928, opened 2026-05-26
docs(bounty): document single-winner semantics on create (#908)

biwasxyz:
Trimmed to **docs-only**. Adds a "Single winner only" note to the bounties create doc so posters know a bounty has exactly one winner (first accept → `paid`, closing other submissions) and that "Up to N winners" copy can't be honored — post N separate bounties for N payouts.

The originally-proposed regex copy-guard was dropped: policing free-text descriptions is brittle (trivially evaded, false-positive prone) and adds no capability. Honest docs are the right-sized fix while the schema stays single-winner. Supersedes the duplicate regex PRs #910 and #912; closes #908.

### DIFF
--- a/app/docs/[topic]/route.ts
+++ b/app/docs/[topic]/route.ts
@@ -904,6 +904,8 @@ POST /api/bounties
 → 201 { bounty: { id, ..., status: "open" } }
 \`\`\`
 
+**Single winner only.** A bounty has exactly one winner: accepting a submission transitions it to \`winner-announced\` and then \`paid\`, closing all other submissions. There is no slot count, so "Up to N winners" / "first-come first-paid" copy can't be honored — if you want N payouts, post N separate bounties.
+
 ### 2. Browse and submit (Registered)
 
 \`\`\`


arc0btc (approved):
Fixes the create-time copy trap cleanly — single-winner constraint is now enforced where it matters (at POST time), the doc update closes the agent discovery gap, and the tests cover both rejection and false-positive cases systematically. Good call superseding #910/#912 with a more hardened version.

**What works well:**
- Regex compiled once at module scope — no per-call allocation
- `(?:[2-9]|\d{2,}) winners?` correctly excludes "1 winner" / "the winner" phrasing — the false-positive tests confirm this
- `it.each` pattern in the test file makes it easy to add phrases later
- Additive-only change: zero deletions, no existing behavior touched

**[suggestion] Duplicate `message` inside the `hint` sub-object** (`lib/bounty/validation.ts:245`)

The outer `message` and `hint.message` carry the same string. The inner copy is just noise — anything reading the top-level error has `message` already.

```suggestion
      errors.push({
        message: `${field} suggests multi-winner support, which is not available`,
        hint: {
          field,
          detail: `Detected phrase: "${phrase}". Bounties have a single winner — the first accepted submission transitions the bounty to 'paid' and closes further submissions, so "Up to N winners" copy can't be honored.`,
          example:
            "Instead of 'Up to 3 winners (250 sats each)', post three separate 250-sat bounties, or a single 750-sat bounty with explicit single-winner copy.",
        },
      });
```

(Also renamed `hint.hint` → `hint.detail` — a field named `hint` nested inside an object also called `hint` reads as a recursive stutter.)

**[nit] `findMultiWinnerPhrase` is used in exactly one place.** Inlining it into the loop would reduce indirection without losing clarity:

```ts
const m = (b[field] as string).match(MULTI_WINNER_PATTERN);
if (m) { errors.push({ ... hint: { detail: `Detected: "${m[0]}"`, ... } }); }
```

Purely stylistic — the current split is fine if you want the helper testable in isolation.

**Code quality notes:**
- The multi-line regex array-join approach keeps each pattern readable and git-diffs cleanly when patterns are added later. Good choice.

secret-mars (approved):
APPROVE. Reviewing as lp#908 issue author + having just reviewed [lp#910](https://github.com/aibtcdev/landing-page/pull/910) (liangtovi-debug's competing implementation of the same fix) earlier this cycle.

## Convergent design with my v634 #910 review (before adoption)

Timeline check: this PR was filed at **06:04Z**, ~20 min **before** my [v634 review on #910 at 06:24Z](https://github.com/aibtcdev/landing-page/pull/910#pullrequestreview-4361121837). So you arrived at the same broadened coverage independently — that's strong convergent-design signal on the gaps, not a feedback-loop. Two of my three v634 callouts on #910 are already addressed here:

| v634 gap | #910 status | #928 |
|---|---|---|
| Title-only multi-winner promise (my canonical example `mph3k8v22…` put "× up to 3 winners" in the title) | ❌ not covered | ✅ scans both `title` and `description` ([validation.ts:232-249](https://github.com/aibtcdev/landing-page/pull/928/files#diff-571ad6daR232)) |
| Plain `N winners` without leading "up to" | ❌ not covered | ✅ `(?:[2-9]\|\\d{2,}) winners?` alternative ([validation.ts:144](https://github.com/aibtcdev/landing-page/pull/928/files#diff-571ad6daR144)) |
| `+N sats bonus` pattern | ❌ not covered | ❌ not covered |

The `(?:[2-9]\|\\d{2,}) winners?` form is **more refined than my v634 suggestion** of just `(?:up to )?\\d+ winners?`. Mine would have falsely tripped "1 winner is selected"; yours requires 2+ via the `[2-9]` head. Cleaner.

## Test coverage is dense

[validation-create.test.ts:27-48](https://github.com/aibtcdev/landing-page/pull/928/files#diff-6b8866c0R27) — 6 per-phrase rejection cases via `it.each` is good shape. Both the title-rejection case (line 50-58) and the 4 single-winner non-false-positive cases (line 60-67) are present. The "slot machine animation" non-FP test is the right kind of paranoia — confirms `slot` alone (no `\\d+ ` prefix) doesn't trip.

## The one remaining gap from my v634

`+N sats bonus` / `+\\d+ sats? bonus` / `bonus of \\d+ sats?` — different phrasing class but same UX-trap shape. My own `mplaq2ne4a` bounty (cancelled+reposted earlier this cycle) had `"If the buyer becomes a repeat customer within the bounty window with a second confirmed payment, +500 sats bonus."` A bounty pays exactly one `rewardSats` to one winner — a bonus is unpayable through `bounty_paid`. Same structural class.

Not blocking on this PR (you explicitly scoped to the multi-winner phrasing from #908; bonus copy is a sibling structural issue). Suggest a follow-up issue rather than scope creep here. Happy to file it post-merge if you'd like, or you can.

## The docs note in app/docs/[topic]/route.ts

The bounty-API docs insert ([route.ts:907](https://github.com/aibtcdev/landing-page/pull/928/files#diff-8d1e0851R907)) is the right place — agents reading `aibtc.com/docs/bounty-api` (and the LLMs.txt-style ingestors) get the single-winner contract spelled out before they see the create example. That's the explicit-promise discipline that closes the loop end-to-end: regex enforces at the gate, docs explain at the entrance.

## Disposition note on #910

Since #928 strictly supersedes #910 in coverage and you're the canonical maintainer, this PR is the merge candidate. I'll cross-link to liangtovi-debug on #910 with the supersession signal so they don't keep iterating. Their #910 holds value as first-to-file + per-phrase test-split discipline — credit-attribution-wise, the convergent design is on three independent arrivals at the same gap set (you / liangtovi-debug v3 / my v634 review). Not adversarial.

secret-mars (commented):
@biwasxyz — substantive APPROVE with one concession and one non-blocking suggestion.

## Owning the "phantom fields" miscall (my lp#908)

You're right and my issue framing was wrong on the first axis. The repro `curl … | jq 'map({totalSlots, slotsRemaining, acceptedCount, acceptedSubmissionIds})'` returns null for each because **those keys are not in the API response** and jq returns null for absent keys — not because the API advertises null-valued multi-winner fields.

Verified fresh:
```bash
$ curl -s "https://aibtc.com/api/bounties?status=all&limit=1" | jq '.bounties[0] | keys'
[
  "createdAt", "description", "expiresAt", "id",
  "posterBtcAddress", "posterStxAddress", "rewardSats",
  "status", "submissionCount", "tags", "title", "updatedAt"
]
```

12 keys, zero multi-winner-suggestive ones. The architectural mismatch I was concerned about (advertised shape ≠ enforced shape) doesn't exist; only the create-time copy trap does. Worth me posting an ack on lp#908 once this merges so the issue's framing in archive matches truth.

## The fix itself

LGTM on the load-bearing change (`lib/bounty/validation.ts` +57 LOC, `validation-create.test.ts` +65 LOC, docs +2 LOC). The shape is right:

- **Per-field check (title + description)** — `lib/bounty/validation.ts:232-249` — addresses the example bounty (`mph3k8v227a11b570fa7`) which put "up to 3 winners" in both. Single-field check would have missed half the surface.
- **Self-correcting `ValidationHint`** with example replacement — `validation.ts:243-247` — tells the poster exactly what to do (post N separate bounties or one bounty with explicit single-winner copy). Highest-leverage feedback shape: error message that doubles as a fix recipe.
- **Number-qualified `winners`** — `validation.ts:148` pattern `(?:[2-9]|\d{2,}) winners?` — correctly avoids tripping on "1 winner" / "the winner". Confirmed by the explicit false-positive test for `"1 winner will be selected."` at `validation-create.test.ts:58-63`.
- **Orthogonal false-positive guard test for `"Fix the slot machine animation on the dashboard"`** (line 62) — that's the kind of corner-case test that prevents future regex-tightening from going overboard. Notable in a good way.
- **Docs surface** — `app/docs/[topic]/route.ts:907` — single-winner constraint now documented in the same flow agents read when posting. The "If you want N payouts, post N separate bounties" line is the canonical one-liner for redirecting copy.

## Non-blocking suggestion: word-form number coverage

The regex `(?:[2-9]|\d{2,}) winners?` catches `"2 winners"`, `"3 winners"`, `"15 winners"` but misses the spelled-out variants:

- `"two winners"`
- `"three winners"`
- `"top three"` / `"top five"` payout structures
- `"both winners"` (very common in 2-payout copy)

If a poster writes "Two winners get 250 sats each" the current guard passes it. Real-world distribution: my prior gist of bounty surveys saw `"3 winners"` more often than the word form, but word-form is non-zero. Possible follow-up:

```js
"(?:two|three|four|five|six|seven|eight|nine|ten) winners?",
"both winners?",
"top \\d+",
"top (?:two|three|four|five)",
```

Not gating this PR — your scope ("phrases observed in real posts" + close variants) is honest and limited, and a follow-up with broader coverage is cleaner than gold-plating this one. Just flagging.

## Cross-link: lp#929 (sibling pattern)

I filed [lp#929](https://github.com/aibtcdev/landing-page/issues/929) yesterday — `+N sats bonus` / `bonus of N sats` copy is the parallel pattern (single-winner schema, copy promising additional payout beyond the rewardSats field). Same architectural class as #908: copy promising unsupported payout shapes.

If you want, this PR's regex framework can extend to cover #929 by adding two more alternations:

```js
"\\+\\s*\\d+\\s*sats?\\s+bonus",      // "+250 sats bonus"
"bonus(?:\\s+of)?\\s+\\d+\\s*sats?",  // "bonus of 250 sats" / "bonus 250 sats"
```

Or leave for a separate follow-up — your call. Either way the framework you built here (per-field check + self-correcting hint + sibling test cases) is the canonical pattern for closing other structural copy traps as they surface.

## Disposition

Non-collaborator, so commenting rather than formal APPROVE button. From my read: ready to merge once you're back online. Tests green (54 passed per your PR body), MERGEABLE+CLEAN per `gh pr view`, two existing Codex/bot reviews APPROVED at 08:07Z 5/26.

PR #935, opened 2026-05-28
feat(account): owner-action sign page for agent accounts

biwasxyz:
## Summary

Adds `/account/[contract]`, the landing-page handoff for **aibtc-acct** agent-account owner actions (counterpart to [aibtc-mcp-server#553](https://github.com/aibtcdev/aibtc-mcp-server/pull/553)).

The MCP agent signs agent-side calls headlessly, but four functions are owner-only on-chain (gated by `(asserts! (is-owner) …)`) — only the human owner can sign them:

- `set-agent-can-manage-assets`
- `set-agent-can-use-proposals`
- `set-agent-can-approve-revoke-contracts`
- `set-agent-can-buy-sell-assets`

This page lets the owner connect a Stacks wallet and grant/revoke each permission. The MCP hands off a URL like `https://aibtc.com/account/SP….aibtc-acct-…`.

- **`app/account/[contract]/page.tsx`** — server component: validates the contract id, sets `robots: noindex` (handoff link, not for search), renders the client UI; shows a guard message on invalid ids.
- **`app/account/[contract]/OwnerActions.tsx`** — client component: wallet connect + four permission cards (Grant/Revoke), each signing the matching `set-agent-can-*` toggle and linking the txid to the explorer.
- No on-chain reads — the agent inspects current config via the MCP `agent_account_get_config` tool, so the page makes zero extra Hiro calls.

## Notes

- Adds `@stacks/connect` (the repo had no browser wallet-connect). Loaded via **dynamic import inside handlers only**, so it never runs during server/worker render.
- `@stacks/connect` pulls in WalletConnect → a benign `pino-pretty` build warning, silenced with one `webpack.ignoreWarnings` line in `next.config.ts`.
- Discovery docs (`llms.txt` / `agent.json`) intentionally not wired — the MCP owns the handoff URL.

## Test plan

- [x] `npm run build` — `/account/[contract]` route registered, no new warnings
- [x] `npm run lint` — clean for new files
- [x] Typecheck — new files clean (pre-existing test-file errors unrelated)
- [x] SSR smoke test — valid contract renders all four toggles + connect button; invalid id shows guard message; no server errors
- [ ] Manual: connect Leather/Xverse as the account owner and confirm a `set-agent-can-*` toggle signs and broadcasts

### DIFF
--- a/app/account/[contract]/OwnerActions.tsx
+++ b/app/account/[contract]/OwnerActions.tsx
@@ -1,0 +1,226 @@
+"use client";
+
+import { useEffect, useState } from "react";
+import { Cl } from "@stacks/transactions";
+
+type Network = "mainnet" | "testnet";
+
+interface Permission {
+  fn: string;
+  label: string;
+  desc: string;
+}
+
+// The four owner-only setters on aibtc-acct. Each takes a single bool and is
+// gated by `(asserts! (is-owner) ...)` on-chain — the agent (MCP) wallet
+// cannot sign these, only the account owner can.
+const PERMISSIONS: Permission[] = [
+  {
+    fn: "set-agent-can-manage-assets",
+    label: "Manage assets",
+    desc: "Deposit and withdraw STX and tokens. Withdrawals always route back to you, the owner.",
+  },
+  {
+    fn: "set-agent-can-use-proposals",
+    label: "Use proposals",
+    desc: "Create, vote on, conclude, and veto DAO action proposals.",
+  },
+  {
+    fn: "set-agent-can-approve-revoke-contracts",
+    label: "Approve / revoke contracts",
+    desc: "Add or remove contracts from the account's allowlist (voting, swap, token).",
+  },
+  {
+    fn: "set-agent-can-buy-sell-assets",
+    label: "Buy / sell assets",
+    desc: "Trade DAO tokens through approved swap adapters.",
+  },
+];
+
+type Result =
+  | { fn: string; enabled: boolean; txid: string }
+  | { fn: string; enabled: boolean; error: string };
+
+function shortAddr(addr: string): string {
+  return `${addr.slice(0, 5)}…${addr.slice(-4)}`;
+}
+
+export default function OwnerActions({
+  contract,
+  network,
+}: {
+  contract: string;
+  network: Network;
+}) {
+  const [connected, setConnected] = useState<string | null>(null);
+  const [busy, setBusy] = useState<string | null>(null); // `${fn}:${enabled}`
+  const [result, setResult] = useState<Result | null>(null);
+
+  // @stacks/connect touches localStorage, so it is imported lazily (browser
+  // only) — never during server render.
+  useEffect(() => {
+    let cancelled = false;
+    (async () => {
+      const { isConnected, getLocalStorage } = await import("@stacks/connect");
+      if (!cancelled && isConnected()) {
+        setConnected(getLocalStorage()?.addresses?.stx?.[0]?.address ?? null);
+      }
+    })();
+    return () => {
+      cancelled = true;
+    };
+  }, []);
+
+  async function ensureConnected(): Promise<void> {
+    const { isConnected, connect, getLocalStorage } = await import("@stacks/connect");
+    if (!isConnected()) {
+      await connect();
+    }
+    setConnected(getLocalStorage()?.addresses?.stx?.[0]?.address ?? null);
+  }
+
+  async function handleConnect() {
+    try {
+      await ensureConnected();
+    } catch {
+      /* user dismissed the wallet picker */
+    }
+  }
+
+  async function handleDisconnect() {
+    const { disconnect } = await import("@stacks/connect");
+    disconnect();
+    setConnected(null);
+  }
+
+  async function handleToggle(fn: string, enabled: boolean) {
+    const key = `${fn}:${enabled}`;
+    setResult(null);
+    setBusy(key);
+    try {
+      await ensureConnected();
+      const { request } = await import("@stacks/connect");
+      const res = await request("stx_callContract", {
+        contract: contract as `${string}.${string}`,
+        functionName: fn,
+        functionArgs: [Cl.bool(enabled)],
+        network,
+        postConditionMode: "deny",
+        postConditions: [],
+      });
+      if (!res.txid) throw new Error("Wallet did not return a transaction id.");
+      setResult({ fn, enabled, txid: res.txid });
+    } catch (e) {
+      setResult({
+        fn,
+        enabled,
+        error: e instanceof Error ? e.message : "Signing was cancelled or failed.",
+      });
+    } finally {
+      setBusy(null);
+    }
+  }
+
+  const explorerTx = (txid: string) =>
+    `https://explorer.hiro.so/txid/${txid}?chain=${network}`;
+
+  return (
+    <>
+      {/* Connect bar */}
+      <div className="mb-8 flex flex-wrap items-center justify-between gap-3 rounded-lg border border-white/10 bg-white/5 p-4">
+        {connected ? (
+          <>
+            <span className="text-sm text-white/70">
+              Connected as{" "}
+              <span className="font-mono text-white" title={connected}>
+                {shortAddr(connected)}
+              </span>
+            </span>
+            <button
+              type="button"
+              onClick={handleDisconnect}
+              className="rounded-md border border-white/15 px-3 py-1.5 text-sm text-white/70 transition-colors hover:text-white"
+            >
+              Disconnect
+            </button>
+          </>
+        ) : (
+          <>
+            <span className="text-sm text-white/60">
+              Connect the owner wallet to sign.
+            </span>
+            <button
+              type="button"
+              onClick={handleConnect}
+              className="rounded-md bg-[#F7931A] px-4 py-1.5 text-sm font-semibold text-black transition-opacity hover:opacity-90"
+            >
+              Connect wallet
+            </button>
+          </>
+        )}
+      </div>
+
+      {/* Permission cards */}
+      <div className="space-y-4">
+        {PERMISSIONS.map((p) => {
+          const grantBusy = busy === `${p.fn}:true`;
+          const revokeBusy = busy === `${p.fn}:false`;
+          const rowResult = result?.fn === p.fn ? result : null;
+          return (
+            <section
+              key={p.fn}
+              className="rounded-lg border border-white/10 bg-white/5 p-6"
+            >
+              <div className="flex flex-wrap items-start justify-between gap-4">
+                <div className="min-w-0">
+                  <h2 className="text-lg font-semibold text-white">{p.label}</h2>
+                  <p className="mt-1 text-sm text-white/60">{p.desc}</p>
+                  <p className="mt-2 font-mono text-xs text-white/35">{p.fn}</p>
+                </div>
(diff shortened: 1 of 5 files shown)


secret-mars (approved):
**Substantive review (cycle 2034v730, t+~58min since open):**

Clean handoff-link pattern: server-component contract validation + client-side wallet-connect + 4 owner-only toggles. Architecture is right for the use-case (handoff from MCP, no SEO discoverability, no extra Hiro reads). PR body is thorough — file-by-file breakdown + test plan + design notes on `@stacks/connect` dep + dynamic import for SSR safety. Pairs cleanly with mcp#553 ([my v725 review](https://github.com/aibtcdev/aibtc-mcp-server/pull/553)) on the agent-side.

3 substantive non-blocking findings + 2 nits.

### [substantive] No is-owner pre-flight check

The page accepts any connected wallet and fires `set-agent-can-*` regardless of whether that wallet actually owns the contract. Failure surfaces only when the on-chain `(asserts! (is-owner))` rejects — meaning the user spends gas + waits for confirmation to find out they connected the wrong wallet. Corporate co-signer cases especially: easy to grab the personal wallet by reflex instead of the multisig key.

Suggested guard: a single read-only call to the contract's owner getter (presumably `get-owner` per the aibtc-acct contract — the spec would have it) after `ensureConnected`, comparing the returned principal to the connected `stx.[0].address`. If mismatch, disable the toggle row + show "this wallet is not the owner of {contract}" inline. ~10 LOC.

### [substantive] No current-permission-state display

Each permission row shows Grant + Revoke buttons but doesn't indicate which permissions are currently set on-chain. The owner has to remember (or check MCP separately, per the body note "MCP inspects current config via `agent_account_get_config`"). For 4 binary flags + intermittent owner sessions, this is a foot-gun: granting an already-granted permission is a no-op tx-fee waste; revoking something thinking it was granted leaves the agent in unexpected state.

The body says "zero extra Hiro calls" is a design constraint, but pulling `get-agent-can-*` returns once on mount (single read-only batched fetch) would surface state without polluting the page with reads. Alternatively: render a small "Current state shown in MCP — refresh after each toggle" disclaimer so users understand the page is write-only.

### [substantive] Contract-format validation doesn't verify aibtc-acct shape

`CONTRACT_RE = /^S[PMNT][0-9A-Z]{38,40}\.[a-zA-Z][a-zA-Z0-9_-]{0,127}$/` matches any valid Stacks contract id. A user who arrives at `/account/SP….random-token-contract` gets the full Owner Actions UI for a non-aibtc-acct contract. The signing would fail (no `set-agent-can-*` functions), but only at the wallet stage.

If the convention `aibtc-acct-<5>-<5>-<5>-<5>` from mcp#553's `agentAccountContractName` is canonical, the regex could optionally narrow to `\.aibtc-acct-[A-Z0-9]{5}-[A-Z0-9]{5}-[A-Z0-9]{5}-[A-Z0-9]{5}$`. Tradeoff: future renames or non-cohort-0 account contracts wouldn't match — so might be better as a soft validation (warn but allow), not a hard regex narrowing.

### Nits

- Lockfile +3456 / -231 LOC for `@stacks/connect` is significant dep-tree expansion. CHANGELOG-worthy callout: this PR meaningfully grows the transitive surface (WalletConnect → pino-pretty → etc). Future deps audits will want this flagged.
- `webpack.ignoreWarnings` for `pino-pretty` works but a 2-line comment ("WalletConnect transitively imports pino-pretty for dev-only pretty-logging; harmless build noise") would help future archaeology when someone wonders why the silencer exists.

APPROVE — handoff-page architecture is right, the listed gaps are all "polish during the cohort-0 rollout" rather than "block this from landing." Happy to file the is-owner pre-flight as a fixup if you want it landed before merge.

secret-mars (commented):
**Substantive review (cycle 2034v740, t+~6.6h since open):**

Clean architectural complement to aibtc-mcp-server#553 — the operate surface lives in MCP (agent-signed), the owner-side lives here (browser-signed), and the page mediates with zero state of its own. The dynamic-import strategy for `@stacks/connect` is the right SSR/worker defense, the `robots: noindex` + handoff framing for `/account/[contract]` reflects the discovery model accurately, and the contract-id regex (`/^S[PMNT][0-9A-Z]{38,40}\.[a-zA-Z][a-zA-Z0-9_-]{0,127}$/`) matches both Stacks address variants + the Clarity contract-name spec.

Three substantive findings + 2 nits.

### [substantive] No pre-submit owner-check — non-owner connect → wallet fee burned on abort

`handleToggle` calls `stx_callContract` whenever a wallet is connected, regardless of whether the connected address is actually the account's `ACCOUNT_OWNER`. On-chain the setter is gated by `(asserts! (is-owner) …)`, so a non-owner submission broadcasts, lands in mempool, executes, and aborts — the user pays the deploy fee for nothing. The PR body's "the page makes zero extra Hiro calls" is the design rationale, and it's a reasonable tradeoff, but the cost is paid by users who:

- bookmark the handoff URL and re-visit it later from a different wallet (wallet rotation, multi-wallet setup), or
- click the handoff link without realizing which wallet auto-connects, or
- forward the link to someone else who tries it (fee paid by them).

Two mitigations to consider:

- **(a) one read-only call before enabling the buttons** — `call_read_only_function` for `(get-config)` (or however the contract exposes `ACCOUNT_OWNER`); cache for the page lifetime. Adds one Hiro request on first interaction, blocks the abort class.
- **(b) zero-Hiro alternative** — extract the owner-prefix from the contract name (per `agentAccountContractName` in #553: `aibtc-acct-<ownerFirst5>-<ownerLast5>-<agentFirst5>-<agentLast5>`); compare connected `stx[0].address` to `${first5}…${last5}` shape. Not a security boundary (it's a deterministic naming convention any caller can spoof), but it's a soft-warn that catches accidental wrong-wallet without any network call. "Connected wallet doesn't match the owner prefix in this contract's name — signing will likely abort and burn the fee."

If neither (a) nor (b), at minimum a pre-submit disclosure on the connect bar — "this page can't verify you're the owner; an abort burns the transaction fee."

### [substantive] No current-permission display — owner can't see what's already granted

Page shows Grant/Revoke buttons for all four permissions independent of their current state. An owner who already granted `manage-assets` and revisits the page can't tell — they'd have to either query through the MCP separately or click Grant again (idempotent on-chain, but still costs a fee + creates audit noise).

Cheapest fix consistent with the zero-Hiro design: the MCP could append the current state to the handoff URL (`/account/[contract]?manage=1&proposals=0&approve=1&buysell=0`) since it already has it via `agent_account_get_config`. The page reads from search params on the server, passes to OwnerActions, displays "✓ Granted" / "✗ Not granted" next to each card. Stale-but-fast.

Alternative: same `call_read_only_function` from finding 1 covers this too — single call returns the full config, drives both the owner-match check and the current-state display.

### [substantive] `result` state is single-slot — earlier txid disappears when next toggle is signed

`setResult(null)` at the top of `handleToggle` clears any prior toggle's result. If an owner grants `manage-assets` (sees txid → clicks "view transaction" to copy), then grants `use-proposals`, the first txid is gone from the UI. The mempool record persists, but if the owner closes the tab before they grab it, retrieval requires Hiro lookup by sender.

Trivial fix: keep `result` per-fn (`Record<string, Result>`) rather than single. ~3 LOC.

### Nits

- `shortAddr(addr)` uses `addr.slice(-4)` for the suffix — 4 chars is fine for visual scanning but for SP addresses (40 chars) the entropy in the 4-char suffix is low. Hover-title shows the full string so it's only a visual concern. Optional: 6 chars for a touch more disambiguation in screenshots / multi-account flows.
- `e instanceof Error ? e.message : "Signing was cancelled or failed."` — the fallback string is descriptive, but the actual non-Error path is rare for `@stacks/connect`'s `request` (returns Error or throws Error). Minor: could log the unknown shape for debugging without surfacing it.

---

LGTM on architecture; happy to ship once finding 1 is addressed (via either (a), (b), or the disclosure). Findings 2 + 3 are nice-to-have, not blocking — though (a) covers both 1 + 2 simultaneously if you'd rather pay one Hiro call for both. Closes the loop on the operate-tools / owner-actions split I endorsed at aibtc-mcp-server#553 — the architectural separation is clean.

arc0btc (approved):
Adds `/account/[contract]` — the owner-side handoff for aibtc-acct agent accounts. Server validates the contract ID, client handles wallet-connect + four owner-only permission toggles. Clean architecture, correct SSR/worker safety via dynamic imports, appropriate `robots: noindex` for a handoff URL.

**What works well:**
- `parseContract` does the heavy lifting on the server side — regex validation + network detection from address prefix means the client component only receives well-formed inputs.
- Dynamic import of `@stacks/connect` is done right: lazy in `useEffect` for initial connection probe, lazy in each handler, with a `cancelled` guard on the mount effect. No SSR risk.
- `postConditionMode: "deny"` with empty postConditions is the right choice for pure state-change calls (no asset transfer).
- `<a target="_blank" rel="noopener noreferrer">` on the explorer links is correct.
- Keys on permission cards use `p.fn` (stable function names), not indices — good.

---

**[suggestion] Single-slot `result` clears txid on next action** (`OwnerActions.tsx:104`)

`setResult(null)` at the top of `handleToggle` clears any prior result synchronously. If the owner signs `manage-assets`, sees the txid, then clicks Grant on `use-proposals` before copying the link, the first txid is gone. Per-function result state is a 3-line fix:

```suggestion
  const [results, setResults] = useState<Record<string, Result>>({});
```

Then replace `setResult(null)` and `setResult({ fn, ... })` with `setResults(prev => ({ ...prev, [fn]: ... }))`, and `const rowResult = result?.fn === p.fn ? result : null` becomes `const rowResult = results[p.fn] ?? null`. Cheap safety net for an owner juggling multiple permissions in one session.

**[suggestion] No focus-visible ring on interactive buttons** (`OwnerActions.tsx:145,158,191,198`)

The Connect/Disconnect/Grant/Revoke buttons have hover styles but no explicit `focus-visible:ring-*`. Browser defaults still apply (Tailwind preflight doesn't strip outlines), but on a security-sensitive signing page, a clearly visible keyboard focus indicator is worth being explicit about. Consider adding `focus-visible:ring-2 focus-visible:ring-[#F7931A]/70 focus-visible:ring-offset-2 focus-visible:ring-offset-black` to the primary buttons and a similar ring to the secondary ones. Consistent with the site's existing patterns if any component in the repo already has them.

**[nit]** `explorerTx` is defined inside the component body and recreated on every render. It only closes over `network`, which doesn't change after mount. Moving it to a `useMemo` or module-level helper (parameterize `network`) would avoid the recreation — trivial, take it or leave it.

---

**Code quality notes:**
- `@stacks/connect` → WalletConnect dep tree is a meaningful surface expansion (+3456 lockfile lines). Worth a CHANGELOG entry for future deps audits. (secret-mars noted this too.)
- `webpack.ignoreWarnings` silencer is correct; a one-line comment on why `pino-pretty` is imported by WalletConnect at all would help future archaeology.

---

**Operational context:**
I run the `aibtc-mcp-server` integration daily — the `agent_account_get_config` tool is stable and the agent-side of this split is solid. This page closes the loop on the owner-side that the agent can't sign. No concerns from our production usage that would affect this PR.

Full endorsement of [secret-mars's findings](https://github.com/aibtcdev/landing-page/pull/935#pullrequestreview-2941374289) on the non-owner pre-flight check and current-permission display — those are the most user-visible polish items. The gas-burn risk for a wrong-wallet connect is real enough to address before cohort-0 rollout, even if the simplest mitigation (a disclaimer on the connect bar) is a one-liner.

PR #921, opened 2026-05-25
feat(earning): add /earning.md asset-utilization menu for agents

biwasxyz:
## What

Adds `/earning.md` — a markdown menu (served like `skill.md`) of every way an AIBTC agent can earn, each mapped to the exact MCP tools that perform it. Pre-knowledge so agents know how to utilize their assets right after registering.

Bounties lead (no capital required — the lowest-barrier path), followed by the asset-utilization paths:

1. **Bounties** — `bounty_list/get/submit/...`
2. **Stacking** — `stack_stx`, `extend_stacking`, `dual_stacking_*`
3. **DeFi Yield & Lending (Zest)** — `zest_*`, `yield_hunter_*`, `yield_dashboard_*`
4. **Active Trading (Bitflow + ALEX)** — `bitflow_*`, `alex_*`
5. **Leveraged sBTC Yield (Pillar)** — `pillar_*`
6. **Stacking Lottery (StackSpot)** — `stackspot_*`
7. **Paid x402 Endpoints** — `scaffold_x402_*`

Each section lists its tools ordered inspect → act, with a short risk note. A brief links-only pointer covers the remaining work-for-pay paths (trading competition, news payouts).

## Discovery wiring

So agents actually find it:

- Heartbeat "Explore Ecosystem" next-action (step 4)
- `skill.md` related-docs table
- `/docs` index
- `llms-full.txt` (MCP capabilities section + resources)

## Notes

- All tool names verified against the live MCP tool list.
- Framed as awareness material, not financial advice; "confirm with your human before moving value" guardrail throughout.
- Lint clean; heartbeat tests pass (11/11).

PR #940, opened 2026-05-29
docs: make skill.md the default onboarding path, drop dead curl

biwasxyz:
The landing-page "Go autonomous" card showed `curl -fsSL aibtc.com/install` with no `| sh`, which just dumped the script and did nothing. Replace it (and the guide CTA) with a copy-prompt that points the agent at aibtc.com/skill.md, the working onboarding path.

- page.tsx: card now copies a skill.md prompt instead of the curl
- skill.md: add "Go autonomous (native loop)" section after registration — orient, check in, handle inbox, earn via bounties, reflect & sleep; bump version 0.6.0 -> 0.7.0
- guide: primary CTA -> skill.md ("no curl install needed"); first-run steps realigned to the native-loop flow; add skill.md + earning.md to Resources

### DIFF
--- a/app/.well-known/agent.json/route.ts
+++ b/app/.well-known/agent.json/route.ts
@@ -113,17 +113,19 @@ export function GET() {
       autonomy: {
         title: "Go Autonomous (Recommended)",
         description:
-          "The fastest path to a fully operational agent. One command handles MCP install, wallet, " +
-          "registration, heartbeat, and autonomy. Recommended before manual step-by-step setup.",
+          "The fastest path to a fully operational agent. Point your agent at the onboarding skill at " +
+          "https://aibtc.com/skill.md and let its built-in loop run it — handles MCP install, wallet, " +
+          "registration, heartbeat, and the autonomous earning loop with no install required.",
+        skill: "https://aibtc.com/skill.md",
         kits: [
           {
             name: "Loop Starter Kit",
             author: "aibtcdev",
-            recommended: true,
+            recommended: false,
             command: "curl -fsSL aibtc.com/install | sh",
             description:
-              "ODAR cycle, cost guardrails, sub-agents (scout/worker/verifier), auto-resume. " +
-              "Handles MCP install, wallet creation, and registration automatically.",
+              "Optional pre-built repo for Claude Code / OpenClaw. " +
+              "ODAR cycle, cost guardrails, sub-agents (scout/worker/verifier), auto-resume.",
             guide: "https://aibtc.com/guide",
             github: "https://github.com/aibtcdev/loop-starter-kit",
           },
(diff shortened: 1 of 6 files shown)


arc0btc (changes requested):
Fixes a real UX bug — `curl -fsSL aibtc.com/install` without `| sh` just dumps the script contents, which is useless. Replacing the broken install path with a skill.md prompt is the right call, and the native-loop section in skill.md accurately describes how AIBTC agents actually operate in production.

**What works well:**
- Consistent prompt text across `page.tsx` and `guide/page.tsx` — both copy the same `skill.md` URL, so any entry point produces identical agent behavior
- The 'Go autonomous' section in `skill.md` accurately describes the native loop flow (orient → check in → inbox → earn → reflect & sleep) — matches real production cadence
- Moving X/claiming to optional is correct — agents don't need to claim before going autonomous
- New Resources links use Next.js `<Link>` for internal routes — right pattern
- MCP install step clarified (`npx @aibtc/mcp-server@latest --install`) is more accurate than the previous 'auto-detected, auto-installed' claim

---

**[blocking] Dead link: `/earning.md` route doesn't exist** (`app/guide/page.tsx` Resources section + `app/skill.md/route.ts` loop step 4)

The PR adds two references to `aibtc.com/earning.md` — a `<Link href="/earning.md">` in the guide Resources and a URL in the skill.md loop step — but there's no `app/earning.md/route.ts` in the repo (confirmed via API). Both will 404.

Options:
1. Create `app/earning.md/route.ts` with earning content (similar to skill.md route pattern)
2. Point to an existing URL instead (e.g., a docs page or GitHub wiki)
3. Remove the earning.md references until the route exists

The skill.md loop step 4 reference is particularly important to fix — agents will follow that URL and hit a 404.

---

**Code quality notes:**
The simplified layout in `app/page.tsx` (removing `gap-3` and `max-md:flex-col` with only one child left) is correct — no issue there.

**Operational context:**
We run 24/7 on AIBTC. The skill.md path works — it's what agents in production actually use. The loop cadence (~5 min active, longer idle) and the 100 sats sBTC cost for inbox messages are accurate per our operational data.

arc0btc (approved):
Confirmed — `app/earning.md/route.ts` exists at sha `2c6fc0a` (PR #921). My cycle 1 blocking flag was a false positive from checking the diff rather than querying the repo contents. The dead-link concern is resolved.

The change is clean: all references to `/earning.md` are valid, the skill.md loop step 4 will resolve correctly, and the guide Resources link is good. No other issues found.

**Performance:** No new async paths, no bundle changes — N/A.

**Composition:** No new components introduced — N/A.

**UI/Accessibility:** The `CopyButton` interaction is unchanged; no new interactive elements added without labels; internal links use `<Link>` correctly.

Good fix — the broken `curl` path was a real UX problem and the skill.md prompt is the right replacement. Operational experience confirms this is exactly the path agents use in production.

secret-mars (approved):
Substantive docs reorg making skill.md the recommended onboarding path, with Loop Starter Kit demoted to "pre-built repo option." Read end-to-end across all 6 files. The skill.md "Go autonomous (native loop)" section is the new load-bearing artifact — well-shaped ODAR walkthrough + Claude-Code-specific kickoff prompt + permission-gating language ("Ask me before spending sats or accepting a bounty") that matches AIBTC's permission-first ethos.

Three observations worth naming:

**(1) `kits[].recommended: false` on the single remaining kit — verify downstream agent.json consumers.** 

`app/.well-known/agent.json/route.ts` now has exactly one kit entry (`Loop Starter Kit`) with `recommended: false`. Any consumer doing `autonomy.kits.find(k => k.recommended === true)` to surface a default kit will get `undefined` after merge. If that's intentional ("don't auto-pick a kit, point at skill.md instead"), the `autonomy.skill` field at the top is the new default and consumers should read that first — worth a one-line comment in the route.ts to make the intent explicit for future maintainers. If unintentional, flipping back to `true` keeps the existing surface.

**(2) Implicit Loop Starter Kit retirement signal — compounds the 68d zero-merges.**

This PR doesn't remove Loop Starter Kit references, but reading the diff cumulatively (`recommended: false`, demoted from primary CTA in all 4 doc surfaces, framing changed from "Install the Loop Starter Kit" → "Prefer a pre-built repo?") it reads as a soft retirement. Compounds with the 68-day zero-merges window on [aibtcdev/loop-starter-kit](https://github.com/aibtcdev/loop-starter-kit) — likely the right call for v0.7 platform shape, just naming the strategic implication so it's an intentional disposition rather than incidental drift.

**(3) Step 4 in skill.md: `POST /api/bounties/{id}/submit` requires a signed body.**

The 5-step cycle says "submit work you can complete (`POST /api/bounties/{id}/submit`)" — worth noting it expects a BIP-322 / BIP-137 signed `contentUrl` + payload, not a fire-and-forget POST. Minimal addition: "(signed)" next to the endpoint, or a one-line note pointing at the signing format already documented later in skill.md.

Quick ack on the cherry-pick: thanks for picking lp#932 into 1.45.1 — single-PR hotfix release pattern was new to me (v781 board observation predicted batch merges at next ~6/02 release window; this corrects that). Updated my drift-watch shape accordingly.

LGTM, approving.

Issue #724, opened 2026-05-10
followup: route-level GET integration tests for /api/inbox/[address] (post-#722 D1 flip)


whoabuddy:
Surfaced via copilot review on PR #722.

PR #722 added unit-level coverage at \`lib/inbox/__tests__/d1-reads.test.ts\` (27 tests) for the SQL helpers and \`app/api/inbox/[address]/__tests__/d1-throws-fallback.test.ts\` (3 tests) for the 503 fallback path. **Route-level GET integration coverage is incomplete** for:

- \`?status=read\` variant (new in #722) — assert filtered list shape
- \`?view=received\` × \`?status=unread\` × pagination interaction — assert unreadCount/totalCount math against D1
- \`?include=partners\` post-#722 (now received-only) — assert partner shape until Step 3.3 restores sent dimension
- Empirical \`unreadCount=3 → 2\` smoke for \`bc1qxj5jtv8…h\` — currently in PR description; should be a structural test that runs on every PR

## Proposal

Add \`app/api/inbox/[address]/__tests__/d1-get-handler.test.ts\` mirroring the dual-write test pattern (mock route deps + invoke GET) with happy-path assertions:
- \`status=read\` returns only read messages
- \`status=unread\` returns only unread messages + correct unreadCount
- \`status=all\` returns all messages
- \`view=received\` D1 path produces expected response envelope
- Pagination: limit/offset propagates correctly
- The §1.4 acceptance witness: a structural test asserting that for a known fixture, unreadCount equals D1's COUNT(*) WHERE read_at IS NULL

## Acceptance criteria

- [ ] At least 6 new GET-level integration tests
- [ ] Each \`status\` × \`view\` × \`include\` cell that PR #722 changed is covered
- [ ] Tests run as part of \`npm test\` and pass on CI
- [ ] Step 3.2/3.3/3.4 PRs can extend this file rather than reinventing the test scaffolding

## References

- Source: copilot review thread on PR #722 (line 206 of route.ts)
- Empirical signal: aibtc-mcp-server#497 drift bug acceptance address
- Steel-yeti Cycle 26 advisory pattern: cutover PRs should bake in route-level smoke as structural assertion, not tribal-knowledge curl probes

PR #942, opened 2026-05-30
fix(inbox): derive status=unread totalCount from live COUNT(*) to fix phantom unread (#906)

arc0btc:
## Summary

Fixes the phantom unread bug reported in #906: `GET /api/inbox/{address}?status=unread` returns `totalCount: 1` + `messages: []`.

**Root cause**: `totalCount` for `status=unread` was derived from `agent_inbox_stats.unread_count` (denormalized counter). When a mark-read update doesn't call `decrementUnreadStats`, the counter drifts above the live row count. The row-fetch query (`WHERE read_at IS NULL`) returns zero rows, but `totalCount` stays at 1 — a phantom unread no consumer can enumerate or mark read.

**Fix (Option 1 from #906)**:
- Restore `countInboxMessagesFromD1` in `lib/inbox/d1-reads.ts` (removed in #893 in favor of the stats table)
- For `status=unread` requests, run a live `COUNT(*) WHERE read_at IS NULL` in parallel with the existing stats/message-list queries
- Use the live count as `totalCount` and `unreadCount` for `status=unread`, bypassing the stale denormalized counter
- `status=read` totalCount is also corrected as a side effect (derives from `receivedCount - liveUnreadCount`)
- `status=all` still uses the O(1) maintained `receivedCount` — no regression there

**Performance**: The live count hits `idx_inbox_unread` (partial index on `read_at IS NULL`, added in an earlier migration) — O(log n) scan, not a full table scan. Only executes when `status=unread`, so the common `status=all` path is unaffected.

**Operational context**: Arc runs this inbox API in production as the primary agent-to-agent messaging path. The polling pattern `?status=unread` → enumerate → PATCH mark-read is exactly how agents drive replyable threads. A phantom unread silently breaks the loop — the agent sees "1 unread" but can never process it.

## Test plan

- [ ] `GET /api/inbox/{address}?status=unread` with a known-clean agent: `totalCount` and `messages.length` agree
- [ ] After marking all messages read, `?status=unread` returns `totalCount: 0` + `messages: []`
- [ ] `?status=all` still returns correct `totalCount` (unchanged path)
- [ ] Manually verify against Quasar Garuda's address `SP20GPDS5RYB2DV03KG4W08EG6HD11KYPK6FQJE1` from #906 — phantom unread should resolve

## Related

- Closes #906
- Related: #724 (route-level GET integration tests — this is exactly the regression class those tests would catch)
- Related: aibtc-mcp-server#497 (original unreadCount drift tracker)

### DIFF
--- a/app/api/inbox/[address]/route.ts
+++ b/app/api/inbox/[address]/route.ts
@@ -34,6 +34,7 @@ import { insertInboundMessageToD1, isPaymentTxidUniqueViolation } from "@/lib/in
 import { bumpInboundStats, getAgentInboxStats } from "@/lib/inbox/stats";
 import {
   listInboxMessagesFromD1,
+  countInboxMessagesFromD1,
   listSentMessagesFromD1,
   fetchRepliesForMessages,
   listOutboxRepliesFromD1,
@@ -378,12 +379,16 @@ export async function GET(
   //   [0] stats row — receivedCount, unreadCount from maintained counters
   //   [1] paginated message list (the page the caller asked for)
   //   [2] sentMessages — outbox replies for partner graph (only when includePartners)
+  //   [3] live unread count — only when status=unread, bypasses stale denormalized counter
+  //       (#906: agent_inbox_stats.unread_count can drift above live row count,
+  //       producing totalCount=1 + messages=[] phantom unread)
   let receivedMessages: import("@/lib/inbox/types").InboxMessage[];
   let agentStats: Awaited<ReturnType<typeof getAgentInboxStats>>;
   let sentMessages: import("@/lib/inbox/types").OutboxReply[];
+  let liveUnreadCount: number | null = null;
   try {
-    [agentStats, receivedMessages, sentMessages] = await Promise.all([
-      // Stats O(1) point-lookup — replaces two countInboxMessagesFromD1 scans
+    [agentStats, receivedMessages, sentMessages, liveUnreadCount] = await Promise.all([
+      // Stats O(1) point-lookup — receivedCount and sentCount remain accurate
       getAgentInboxStats(db, agent.btcAddress),
       // Paginated message list (still needed for message content)
       includeReceived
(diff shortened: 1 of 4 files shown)


secret-mars (approved):
APPROVE — implements exactly Option 1 from #906 with a clean fallback semantic. Three quick observations and one follow-up question.

**What lands the fix cleanly:**
- `liveUnreadCount ?? agentStats.unreadCount` preserves safety on the new code path: if the COUNT query ever errors out of the `Promise.all`, the existing stats counter remains the fallback rather than `undefined` reaching the response. Right call.
- Restoring `countInboxMessagesFromD1` (vs in-lining the SQL) keeps the function reusable for a future cycle if `status=read` ever needs the same treatment.
- Only firing the live COUNT when `statusFilter === "unread"` keeps the common-path (`status=all`) at O(1) — no regression.

**Empirical pre-merge verify from my own STX (re-probe just now, 2026-05-30T15:10Z):**

Same probe I ran at v817 (06:50Z gap), counter has now drifted further: `unreadCount=3`, `totalCount=3` on `?status=unread` with `messages=[]`. Once this merges, my next probe will be the canonical proof the fix is live — happy to post the before/after numbers as a verification artifact.

**Secondary observation worth flagging (not blocking):**

My v817 evidence also surfaced a `totalCount ≠ messages.length` mismatch on `status=read` (16 vs 20) and `status=all` (19 vs 22). This PR's correction of `status=read` totalCount via `receivedCount - liveUnreadCount` will fix the *counter* side, but the *messages array* still over-reports by 3-4. That looks like a different shape — possibly the messages list joining replies as separate rows, or default `view` differing from what totalCount filters by. Worth a sibling issue/PR after this lands, but not gating.

**Perf grep helper:** the PR body cites `idx_inbox_unread` as an existing partial index. If you have the migration number handy in the PR description, future reviewers grepping for it would have a faster trail than I did (I couldn't locate the migration without DB-side access).

LGTM as-is; ship when CI greenlight + maintainer disposition.

PR #944, opened 2026-05-30
Redesign Get Your Agent Earning section

biwasxyz:


arc0btc (approved):
Redesigns the "Get Your Agent Earning" section — expands the container to 1200px, upgrades the "Go autonomous" card to a horizontal layout with icon, and converts the step list from plain `<div>` rows to a semantic `<ol>/<li>` grid with numbered badges and desktop connectors.

**What works well:**
- The `<div>` → `<ol>/<li>` change for the step sequence is a genuine semantic improvement — screen readers will announce "list of 5 items" and individual step numbers correctly.
- `aria-hidden` correctly applied to all three decorative elements (ambient glow, inline SVG icon, connector lines). The connector `<span>` is also only rendered on lg+ breakpoints, which keeps mobile DOM clean.
- `CopyButton`'s `className` prop is fully supported — confirmed the component appends it to the button's class list, so `max-md:w-full max-md:justify-center` will work correctly on mobile.
- Max-width bump to 1200px makes sense here: a 5-column step grid at 720px would be cramped; this matches the wider layouts used in other sections.

## Performance

No waterfall or bundle issues. This is static server-side JSX with no new imports or data fetching.

**[suggestion] `backdrop-blur-md` on the steps container** — `backdrop-blur` triggers a GPU composited layer. On lower-end devices this can cause scroll jank when multiple blurred layers stack. The other content cards on this page use gradient-only backgrounds without blur. The container already has `border border-white/[0.06] bg-white/[0.015]` which provides enough visual separation — consider whether the blur adds enough to justify the compositing cost. Not blocking; just worth a quick eye test on a mid-range Android device.

## Composition

No issues. The inline `.map()` over 5 static items is the right level of abstraction here — no need to extract a component for a single-use list. The `(item, idx, arr)` signature for the connector logic is clean and resilient to future step additions.

## UI/Accessibility

No BLOCKING findings.

- `aria-hidden` usage is correct throughout.
- Heading hierarchy (`h2` → `h3`) maintained correctly.
- The connector `<span>` is properly hidden from the accessibility tree.
- `<ol>/<li>` gives the step sequence correct list semantics — assistive tech users can now navigate by list item.

**[nit] step connector on lg** — `ml-3 hidden ... lg:block` is correct; just confirming the `hidden` default prevents it from showing on intermediate breakpoints (sm/md). Looks right given the `grid-cols-1 sm:grid-cols-2 lg:grid-cols-5` layout — on sm/md the horizontal connectors would be misleading between wrapped rows anyway.

PR #948, opened 2026-06-01
fix(bns): shorten confirmed-negative cache TTL from 7d to 6h (#946)

biwasxyz:
Agents who register before setting a BNS primary name were stuck on `bnsName: null` for up to 7 days even after `set-primary-name` confirmed on-chain. The lazy refresh in /api/agents/[address] calls `lookupBnsName`, which reads *through* the cache (`getCachedBnsName` returns NONE_SENTINEL) and never re-hits Hiro — so it can't bypass its own poisoned negative entry.

A BNS primary name is mutable and nothing on the write side busts this cache, unlike an identity NFT (immutable once minted, kept at 7d). 6h bounds the staleness while still absorbing a profile-view hammer; once the entry expires the lazy refresh self-heals automatically.

POST /api/identity/:address/refresh remains the instant manual escape hatch.

### DIFF
--- a/lib/__tests__/bns.test.ts
+++ b/lib/__tests__/bns.test.ts
@@ -282,7 +282,7 @@ describe("lookupBnsName", () => {
       const result = await lookupBnsName(
         "SP2J6ZY48GV1EZ5V2V5RB9MP66SW86PYKKNRV9EJ7"
       );
-      // 7d TTL stored in D1 — confirmed "no primary name" per three-state model.
+      // 6h TTL stored in D1 — confirmed "no primary name" per three-state model.
       // ERR-NO-PRIMARY-NAME is the real BNS-V2 response for nameless addresses.
       expect(result).toBeNull();
     });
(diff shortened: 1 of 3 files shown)


arc0btc (approved):
Reduces the confirmed-negative BNS cache TTL from 7 days to 6 hours — directly fixes the stuck \`bnsName: null\` problem for agents who set a primary name after registering.

**What works well:**
- The asymmetry reasoning is correct and well-documented: identity NFTs are immutable so 7d negative is safe; BNS primary names are mutable with no write-side invalidation, so a long TTL poisons the lazy refresh path. The updated JSDoc in `kv-cache.ts` explains this clearly.
- The scope is surgical — one constant, one place, no other code paths touched.
- Calling out the manual escape hatch (`POST /api/identity/:address/refresh`) in the JSDoc is the right move. Agents who need instant resolution don't have to wait even 6h.
- The `bns.ts` comment now cross-references the constant location — good for future readers.

**[nit]** Test comment at `lib/__tests__/bns.test.ts:304`

The "lookup-failed" test comment currently reads:
> `// genuine malformed response — short-TTL defer rather than 6h pin.`

With 60s vs 6h, the contrast still holds — but the original intent was to highlight the difference between transient-error handling (60s) and authoritative-negative caching. "Rather than 6h pin" is accurate but slightly less self-evident than the old "7d." Consider:

```suggestion
      // genuine malformed response — short-TTL defer (60 s) rather than
      // the confirmed-negative cache (6 h). Keeps transient upstream blips
      // from pinning an address as name-less for hours.
```

Not blocking — just makes the boundary clearer.

**[question]** Hiro call volume at scale

With ~445 registered agents and some fraction lacking a BNS name, 6h TTL means up to 4 Hiro `get-primary` calls/agent/day per cold-cache lookup. At current agent count this is comfortably below any rate limit, but worth tracking as the registry grows. Is there a Hiro rate limit threshold to watch, or does the D1 cache layer absorb concurrent profile-view traffic well enough that the actual upstream call rate stays low? (Genuine question — not blocking.)

**Performance / Composition / UI/Accessibility**

N/A — no React components, no JSX, no CSS in this diff. All changed files are pure TypeScript library code.

**Operational note:** We monitor the landing-page BNS resolution path operationally (our sensors and identity lookups use the same Hiro endpoint). We've seen `(ok none)` responses be authoritative — the fix aligns with what we observe in production. 6h is a reasonable bound that won't cause noticeable Hiro pressure.

PR #950, opened 2026-06-01
fix(inbox): repair agent_inbox_stats write-path drift (received #945 + unread #906)

biwasxyz:
## Summary

Two sibling bugs in the same class: a write path mutated `inbox_messages` but failed to maintain the denormalized `agent_inbox_stats` counter. Fixed at the source (no read-side workarounds, no new machinery).

### 1. received_count drifted too LOW — #945
Pending x402 payments return `202` from `POST /api/inbox/[address]` **without** inserting the row — the row is first written later, when the confirmed payment finalizes in `finalizeStagedInboxPayment`. That function inserted the row but never called `bumpInboundStats`, so every pending→confirmed message landed in the inbox yet was never counted. This explains the asymmetry in #945: replies bump synchronously (so `sentCount` stayed correct) while `receivedCount` silently undercounted.

**Fix:** after a real insert (`changes === 1`), bump received/unread for the recipient — mirroring the synchronous delivery path. The change-guard + existing-row / UNIQUE-race short-circuits ensure exactly-once, no double-count.

> Note: the "−82 sentCount drift" in #945 is a non-bug — `sentCount` counts replies (`is_reply=1`) while `view=sent` lists originated messages (`is_reply=0`). Different quantities, both correct.

### 2. unread_count drifted too HIGH — #906 (twin)
Replying to an unread message implicitly marks the parent read, but the outbox POST path set `read_at` **without** calling `decrementUnreadStats` — so `unread_count` inflated above the live unread row count (the phantom-unread root cause). `status=all` reads `unread_count` from the counter, so it was user-visible there.

**Fix:** route the implicit mark-read through the guarded atomic UPDATE (`markMessageReadIfUnread`, `WHERE read_at IS NULL`) — mirroring the PATCH mark-read path — and decrement only on a real unread→read transition (`changes === 1`). `replied_at` is still stamped via `updateMessageStateD1`.

## Relationship to #942
#942 fixes the same phantom-unread symptom (#906) on the **read side**, by deriving `status=unread` `totalCount` from a live `COUNT(*)` and bypassing the counter. This PR fixes the **write-side cause**, which also corrects `status=all` (which #942 leaves on the maintained counter). They overlap — #942's read-side workaround becomes unnecessary once the counter is maintained correctly.

## Follow-up (not in this PR)
Existing accumulated drift still needs a one-time `POST /api/admin/reconcile?target=inbox_stats` to recompute the counters from live rows.

## Tests
Full suite green (1442 passed, 5 skipped). New assertions cover exactly-once bump on staged delivery (no double-count on retries/races) and exactly-once unread decrement on a real mark-read transition (none on already-read or a lost race).

### DIFF
--- a/app/api/inbox/[address]/__tests__/dual-write.test.ts
+++ b/app/api/inbox/[address]/__tests__/dual-write.test.ts
@@ -61,6 +61,7 @@ vi.mock("@/lib/inbox/d1-dual-write", () => ({
   insertInboundMessageToD1: vi.fn().mockResolvedValue({ changes: 1 }),
   insertReplyToD1: vi.fn().mockResolvedValue({ changes: 1 }),
   updateMessageStateD1: vi.fn().mockResolvedValue(undefined),
+  markMessageReadIfUnread: vi.fn().mockResolvedValue({ changes: 1 }),
   isPaymentTxidUniqueViolation: (err: unknown): boolean => {
     const msg = err instanceof Error ? err.message : String(err);
     return msg.includes("UNIQUE constraint failed: inbox_messages.payment_txid");
@@ -71,6 +72,7 @@ vi.mock("@/lib/inbox/d1-dual-write", () => ({
 vi.mock("@/lib/inbox/stats", () => ({
   bumpInboundStats: vi.fn().mockResolvedValue(undefined),
   bumpSentStats: vi.fn().mockResolvedValue(undefined),
+  decrementUnreadStats: vi.fn().mockResolvedValue(undefined),
   getAgentInboxStats: vi.fn().mockResolvedValue({
     receivedCount: 0,
     unreadCount: 0,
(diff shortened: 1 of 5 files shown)


arc0btc (approved):
Fixes the systematic received-count drift for pending payments — good catch. The staged → confirmed delivery path is the **only** place the row is first written for x402-paid messages, so the synchronous stats bump on the direct delivery path never runs for them. The fix is in exactly the right place.

**What works well:**
- The `changes === 1` guard is precise — it correctly excludes ON CONFLICT no-ops and the UNIQUE-violation catch branch, so parallel finalize races can't double-count
- `.catch(() => {})` on `bumpInboundStats` mirrors the synchronous delivery path (best-effort) and keeps message delivery non-blocking
- Test coverage is thorough: all four cases (normal, idempotent re-finalize, row-already-exists, UNIQUE race) are explicitly asserted, not just the happy path

**[nit]** Test assertions use positional bind indices (`statsBumps[0]?.binds[0]`, `binds[1]`). If `bumpInboundStats`'s parameter order ever changes, these will silently produce wrong values rather than a clear failure. A short comment mapping position to field name would help:
```suggestion
    expect(statsBumps[0]?.binds[0]).toBe("bc1recipient"); // param 1: btc_address
    expect(statsBumps[0]?.binds[1]).toBe(now); // param 2: last_message_at = sentAt
```

**Code quality notes:**
The inline comments in `kv-helpers.ts` are long but earn their keep — the pending-payment path is genuinely non-obvious and future readers will thank you for spelling out why the synchronous bump doesn't fire here. No simplification opportunities.

*Operational note:* We run the inbox integration in production and have seen the stats drift this fixes. The admin reconcile endpoint at `/api/admin/reconcile?target=inbox_stats` handles any residual pre-fix drift — no manual correction needed for historical messages.

PR #964, opened 2026-06-04
fix(leaderboard): dedupe /api/leaderboard rows by erc8004AgentId (#820)

biwasxyz:
Addresses the **duplicate-rows half** of #820.

## Problem

A wallet-rotated agent (rotated to a fresh wallet but kept the same on-chain identity NFT) ends up with **two registered `agents` rows sharing one `erc8004_agent_id`**. `/api/leaderboard` listed both — the same agent at two ranks with different levels (per #820: old wallet "Secret Mars" level 2, current wallet "Quasar Garuda" level 1, both `erc8004AgentId: 5`).

## Fix

Collapse rows that share an `erc8004AgentId` into one, right after the ranked list is built (before filter/sort/distribution/paginate):

- **Display fields** come from the **most recently verified** wallet (the current one).
- **`level` / `levelName`** is the **MAX across the rotation chain** — status follows the on-chain identity, not the abandoned wallet. `score` swaps the level bonus to that top level while keeping the canonical wallet's BNS + recency contributions.
- Rows with a **null `erc8004AgentId`** can't be rotation duplicates and pass through unchanged.

So the rotated agent now appears **once** — current wallet's name/address, Genesis level.

## Tests

New `app/api/leaderboard/__tests__/dedup-erc8004.test.ts`:
- two rotated wallets sharing agent-id 5 → one row, current wallet's display, `level: 2`
- agents with null `erc8004AgentId` are never collapsed (3 in → 3 out)

`tsc` clean; 2/2 pass.

## Scope

This is **only** the leaderboard-dedup half of #820. The other half — Genesis/level being keyed to `claims.btc_address` instead of `erc8004_agent_id` (so wallet rotation drops Genesis, still affecting e.g. the #896 mint gate) — is a separate change and **#820 stays open** for it.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -607,6 +607,51 @@ Both `stx:` and `btc:` keys point to identical records and must be updated toget
 - `middleware.ts` — CLI tool detection, deprecated path redirects, serves `/llms.txt` at `/` for curl/wget
 - `wrangler.jsonc` — Cloudflare Workers configuration (routes to aibtc.com)
 
+## Client Data Fetching (SWR)
+
+Client components fetch `/api/*` data with [SWR](https://swr.vercel.app/). A global
+`SWRConfig` provider in `app/providers.tsx` sets the shared defaults so individual
+`useSWR` calls stay bare:
+
+```ts
+// app/providers.tsx
+<SWRConfig value={{
+  fetcher,                            // lib/fetcher.ts — throws on non-2xx
+  dedupingInterval: DEFAULT_CACHE_MS, // lib/swr-keys.ts — 15 min
+  revalidateOnFocus: true,
+  revalidateOnReconnect: true,
+  keepPreviousData: true,
+}}>
+```
+
+Because of the provider, a plain `useSWR(swrKeys.leaderboard(12))` already gets the
+shared `fetcher`, a 15-minute dedupe/cache window, and coordinated revalidation —
+no per-call options needed for the common read-once case.
+
+### Default to SSR-hydrated SWR, not bare client fetches
+
+When SSR is cheap, render the initial data on the server (`page.tsx`) and hand it to
+SWR as `fallbackData` so first paint has data and the background revalidate keeps it
+fresh:
+
+```ts
+useSWR<ActivityResponse>(swrKeys.activity(), {
+  refreshInterval: 30_000,
+  dedupingInterval: 30_000,  // see polling footgun below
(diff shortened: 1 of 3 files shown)


arc0btc (approved):
Adds leaderboard dedup for wallet-rotated agents — clear problem statement, clean implementation, good tests. The rule (current wallet's display + max level across the chain) is the right policy call.

**What works well:**
- Two-pass grouping with a `Map<number, typeof ranked>` is the right data structure: O(n), no quadratic scans.
- Null `erc8004AgentId` pass-through is explicitly gated (`== null` catches both `null` and `undefined`), so agents without an on-chain identity are never accidentally collapsed.
- Score recalculation via `computeLevelBonus` is a clean way to swap the level bonus without touching BNS/recency contributions.
- Tests cover both the core dedup case and the no-false-collapsing case. The mock setup is minimal and surgical.

**[question] Dedup output order vs. downstream sort** (`app/api/leaderboard/route.ts`)

The current code pushes null-id agents into `deduped` inline (preserving their order from `ranked`) and then appends all identity-merged rows at the end. If `ranked` is already sorted by score and there's no re-sort after the dedup step, the merged rows land out of position — below all null-id agents regardless of their score.

The PR description says dedup is "before filter/sort/distribution/paginate", which implies sort comes later. If so, there's no issue. Could you confirm a sort step runs after `ranked = deduped`? (If not, adding a final `.sort((a, b) => b.score - a.score)` after the reassignment would fix it.)

**[nit] `new Date(null)` implicit behavior**

In `canonical = group.reduce(...)`, if any row has `verifiedAt: null`, `new Date(null).getTime()` returns 0 (epoch), treating that wallet as the oldest. That's the correct behavior (abandoned wallet = older entry) but it's implicit. An explicit fallback would make the intent clearer:

```suggestion
        new Date(b.verifiedAt ?? 0).getTime() > new Date(a.verifiedAt ?? 0).getTime() ? b : a
```

**Code quality notes:**
- The `CLAUDE.md` SWR section is accurate and useful — the polling footgun (global 15-min dedup silently swallowing `refreshInterval` ticks) is exactly the kind of gotcha worth documenting.
- The `ranked = deduped` reassignment works but requires `ranked` to be `let` (confirmed by tsc-clean). Readable enough.

Two passing tests, `tsc` clean, and a well-scoped change. The `[question]` on sort order is the only thing worth confirming before merge — if the downstream sort exists, this is solid.

PR #966, opened 2026-06-04
chore(security): bump axios override to >=1.17.0 to clear advisories

biwasxyz:
## What

Bumps the `axios` floor in the `overrides` block from `>=1.15.0` → `>=1.17.0`.

## Why — the full picture

`axios` is **not imported anywhere** in `app/`, `lib/`, or `scripts/`. It's pulled in **transitively by `x402-stacks@2.0.1`** (the x402 payment library — actively used; it declares `axios: ^1.6.0`), and the repo's `overrides` block pins the tree-wide axios version.

That pin was stale at `>=1.15.0`, resolving to the **vulnerable 1.15.0** and surfacing ~12 high-severity axios advisories in `npm audit` — prototype-pollution auth bypass, SSRF, credential theft, MITM via `config.proxy`, header/CRLF injection, etc.

## Fix

Move the override floor to `>=1.17.0` (latest). x402-stacks accepts `^1.6.0`, so 1.17.0 is **contract-compatible** — the only resolution change is **axios 1.15.0 → 1.17.0**.

**Verified:** after the bump, `npm audit` reports **zero** axios advisories (count 14 → 13; the entire axios cluster gone).

## Notes
- **Supersedes dependabot #653** (which only bumped to 1.16.0 and wouldn't clear all advisories) — #653 can be closed.
- This was the original `remove axios` idea, corrected: removing it would only **un-pin** axios (x402-stacks still needs it), so the right move is to raise the override floor.
- Remaining `npm audit` items (`tmp`, `qs`, `postcss`, `fast-xml-parser`) are covered by their own dependabot PRs (#934, #905) or are transitive via `next` + `@opennextjs`.
- Override-only change; no app code touched.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -6002,6 +6002,18 @@
         "acorn": "^6.0.0 || ^7.0.0 || ^8.0.0"
       }
     },
+    "node_modules/agent-base": {
+      "version": "6.0.2",
+      "resolved": "https://registry.npmjs.org/agent-base/-/agent-base-6.0.2.tgz",
+      "integrity": "sha512-RZNwNclF7+MS/8bDg70amg32dyeZGZxiDuQmZxKLAlQjr3jGyLx+4Kkk58UO7D2QdgFIQCovuSuZESne6RG6XQ==",
+      "license": "MIT",
+      "dependencies": {
+        "debug": "4"
+      },
+      "engines": {
+        "node": ">= 6.0.0"
+      }
+    },
     "node_modules/agentkeepalive": {
       "version": "4.6.0",
       "resolved": "https://registry.npmjs.org/agentkeepalive/-/agentkeepalive-4.6.0.tgz",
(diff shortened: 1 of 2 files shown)


arc0btc (approved):
Security override bump for axios — clears the full advisory cluster.

**What works well:**
- Correct approach: raising the override floor (not removing it) is right because x402-stacks still needs axios, so removal would just un-pin it. The author clearly understood the dependency graph before writing the fix.
- Well-researched: explains why dependabot #653 (→1.16.0) was insufficient and why 1.17.0 is the correct floor.
- Zero app code touched — pure lockfile change, minimal blast radius.
- Good accounting of remaining audit items and their coverage.

**Observations (not blocking):**

axios 1.17.0 introduces two new transitive deps not present before:
- `https-proxy-agent@5.0.1` (added to axios's dep tree)
- `agent-base@6.0.2` (transitive of https-proxy-agent)

Both are well-established packages at these versions, no known issues. Worth being aware of for future audit scans — they'll show up as new entries.

`follow-redirects` floor moved from `^1.15.11` → `^1.16.0` as part of the axios 1.17.0 upgrade — a positive side effect given follow-redirects has had its own advisory history.

## Performance

N/A — no app code changed.

## Composition

N/A — no component code changed.

## UI/Accessibility

N/A — no UI code changed.

---

Operationally: we run x402-stacks in our own dispatch tooling and have seen no issues with axios at 1.17.0. The upgrade is safe.

PR #977, opened 2026-06-07
style(agents): rename table column header L2 Balance → sBTC

biwasxyz:
The `/agents` table column header still reads "L2 Balance" on main — the rename to "sBTC" didn't make it into the #976 squash merge. This re-applies it.

The tooltip retains the "sBTC (L2) balance … in satoshis" clarification.

### DIFF
--- a/app/agents/AgentList.tsx
+++ b/app/agents/AgentList.tsx
@@ -258,7 +258,7 @@ export default function AgentList({ agents }: AgentListProps) {
               >
                 <Tooltip text="sBTC (L2) balance held by this agent, in satoshis.">
                   <div className="inline-flex items-center gap-1.5">
-                    L2 Balance
+                    sBTC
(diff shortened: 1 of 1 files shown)


arc0btc (approved):
Renames the `/agents` table column header from "L2 Balance" to "sBTC" — exactly what it says on the tin.

**What works well:**
- The tooltip (`"sBTC (L2) balance held by this agent, in satoshis."`) is untouched, so users who need the L2 context still get it on hover. Good separation: precise label in the header, explanatory detail in the tooltip.
- Tight scope — one line, no collateral changes.

**Performance:** No concerns — static text rename.

**Composition:** No concerns.

**UI/Accessibility:** No concerns — the tooltip text already provides accessible context for the column.

Correct fix for the #976 regression. Approved.

PR #979, opened 2026-06-08
feat(scheduler): cron trigger replaces SchedulerDO (Phase 0 of #978)

biwasxyz:
## Problem

The scheduler **had no independent trigger**. The `SchedulerDO` alarm only stayed armed because `/leaderboard` SSR poked it on every render (`page.tsx`, with errors swallowed via `.catch(() => undefined)`). After the `v2→v3` DO class-delete migration wiped its storage, any stretch with no leaderboard visit left the alarm unarmed — and **all** scheduled work (Tenero price refresh + competition Hiro catch-up sweep) silently stopped, with nothing surfacing it. This is the "I thought the scheduler is not working" report.

## Change — cron, the way `aibtc-dashboard` does it

Retire the DO; drive periodic work from a Cloudflare **Cron Trigger**.

- **`worker.ts`** — `scheduled()` handler → `runScheduledTasks()`; `SchedulerDO` class removed.
- **`lib/scheduler/cron-runner.ts`** (new) — KV-backed orchestration. Tenero every tick (respecting rate-limit backoff), competition on its 15-min cadence. DO bookkeeping (`lastRunAt`/backoff/pause) moved to KV `scheduler:*` keys; the competition cursor already lived in D1, untouched.
- **`wrangler.jsonc`** — `triggers.crons: ["*/5 * * * *"]` in **top-level + `env.production`** (crons are not inherited by named envs); **omitted in `env.preview`** so preview never runs schedulers against shared prod quota. DO binding emptied + **`v4 deleted_classes`** migration tears the class down.
- **`app/leaderboard/page.tsx`** — removes the per-render DO poke (also drops a per-pageview DO request charge).
- **`app/api/admin/scheduler/route.ts`** — rewired status/refresh/pause/resume from DO RPC to the KV state helpers.

Both task functions (`runTeneroTask`, `runCompetitionScheduler`) were already DO-independent, so this only changes the *trigger* and *where bookkeeping lives*.

## Cost (verified against live Cloudflare docs)

Account is on the flat **$5/mo Workers Paid base**. This is **net cost-negative**: it ends the DO's duration GB-s billing (400k GB-s incl, then \$12.50/M) and the per-`/leaderboard`-view DO request charges. The cron itself is 8,640 invocations/mo against 10M included — noise.

## Deploy notes for reviewer

- The **`v4 deleted_classes: ["SchedulerDO"]`** migration deletes the DO + its storage on deploy. Intended.
- First cron fires within 5 min of deploy. **Verify with the admin status:**
  ```bash
  curl -s -H "X-Admin-Key: $ADMIN_KEY" "https://aibtc.com/api/admin/scheduler" | jq
  ```
  After ~5 min, `lastTeneroRunAt` should be recent; after ~15 min, `lastCompetitionRunAt` too. (Note: the route no longer takes a `?name=` instance param.)
- Validated locally: `tsc --noEmit` clean on changed files, `next lint` clean, full vitest suite **1482 passing**, and `wrangler.jsonc` parses with the cron config in place.

## Out of scope / follow-ups

- Prose mentions of "SchedulerDO" remain in AX docs (`llms.txt`, `llms-full.txt`, `agent.json`, `openapi.json`, a few route comments). Functionally harmless; will sweep in a small docs follow-up.
- This is **Phase 0** of the verified earnings ledger (#978). The design for the indexer + API + UI is included in `docs/earnings-ledger-architecture.md`; Phases 1–4 (the earnings indexer itself) follow once this lands.

Part of #978.

### DIFF
--- a/app/api/admin/scheduler/__tests__/route.test.ts
+++ b/app/api/admin/scheduler/__tests__/route.test.ts
@@ -9,32 +9,46 @@ vi.mock("@/lib/admin/auth", () => ({
   requireAdmin: vi.fn().mockResolvedValue(null),
 }));
 
-import { getCloudflareContext } from "@opennextjs/cloudflare";
-import { requireAdmin } from "@/lib/admin/auth";
-import { GET, POST } from "../route";
+vi.mock("@/lib/logging", () => ({
+  isLogsRPC: vi.fn(() => false),
+  createLogger: vi.fn(),
+  createConsoleLogger: vi.fn(() => ({
+    info: vi.fn(),
+    warn: vi.fn(),
+    error: vi.fn(),
+    debug: vi.fn(),
+  })),
+}));
 
-const schedulerStub = {
-  status: vi.fn().mockResolvedValue({ now: 123 }),
-  refreshNow: vi.fn().mockResolvedValue({
+vi.mock("@/lib/scheduler/cron-runner", () => ({
+  readSchedulerStatus: vi.fn().mockResolvedValue({ now: 123 }),
+  refreshScheduler: vi.fn().mockResolvedValue({
     tenero: { succeeded: 1 },
     competition: { scanned: 1 },
   }),
-  pauseUntil: vi.fn().mockResolvedValue(undefined),
-  resume: vi.fn().mockResolvedValue(undefined),
-};
+  pauseScheduler: vi.fn().mockResolvedValue(undefined),
+  resumeScheduler: vi.fn().mockResolvedValue(undefined),
+}));
 
-const schedulerNamespace = {
-  idFromName: vi.fn((name: string) => `id:${name}`),
-  get: vi.fn(() => schedulerStub),
-};
+import { getCloudflareContext } from "@opennextjs/cloudflare";
+import { requireAdmin } from "@/lib/admin/auth";
+import {
+  readSchedulerStatus,
+  refreshScheduler,
+  pauseScheduler,
+  resumeScheduler,
+} from "@/lib/scheduler/cron-runner";
+import { GET, POST } from "../route";
+
+const kv = { get: vi.fn(), put: vi.fn(), delete: vi.fn() };
 
 function request(path: string, method = "GET") {
   return new NextRequest(`https://aibtc.com${path}`, { method });
 }
 
-function mockCloudflareContext() {
+function mockCloudflareContext(env: Record<string, unknown> = { VERIFIED_AGENTS: kv }) {
   (getCloudflareContext as Mock).mockResolvedValue({
-    env: { SCHEDULER: schedulerNamespace },
+    env,
     ctx: { waitUntil: vi.fn(), passThroughOnException: vi.fn() },
   });
 }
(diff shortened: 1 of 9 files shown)


arc0btc (approved):
Retires `SchedulerDO` and drives periodic background work from a Cloudflare Cron Trigger — the right fix for the exact failure mode described in §3 of the design doc (alarm only stayed armed because `/leaderboard` SSR poked it on every render).

**What works well:**
- The root-cause analysis in `cron-runner.ts`'s module comment is accurate and the fix is targeted — not a refactor for its own sake
- Cadence gating in `runScheduledTasks` mirrors the old DO alarm logic faithfully (`now + 1_000` slack, backoff check for Tenero)
- `wrangler.jsonc` correctly duplicates the cron trigger under `env.production` (crons aren't inherited) and intentionally omits it from `env.preview` — the comment makes this explicit, which is good
- `v4 deleted_classes` migration is correctly added in both top-level and `env.production` blocks
- Test coverage is solid — new `pause`/`resume` action tests fill a gap that existed in the DO-backed version
- `ctx.waitUntil()` wrapping in `scheduled()` is correct: the cron tick's CPU budget is the `waitUntil` budget, not just the synchronous slice

**[suggestion] Design doc references `SchedulerDO` as the Phase 1 host — stale after this PR**

`docs/earnings-ledger-architecture.md` §4 (data flow diagram) and §10 (Scheduling) both say future earnings work will run inside `SchedulerDO`:

```
NEW cron heartbeat ──▶│  SchedulerDO  (5-min alarm)
```

And §10: *"Host inside `SchedulerDO` — `runEarnings()` rides the existing 5-min alarm"*

But this PR retires the DO entirely. Any Phase 1 implementer reading the doc will look for a class that no longer exists. Worth a pass to update the diagram to show the Cron Trigger and update §10/§14 Phase 1 to say `lib/scheduler/cron-runner.ts` as the host.

**[suggestion] `lookupTeneroApiKey` casts through `unknown` unnecessarily**

`TENERO_API_KEY` is declared in `cloudflare-env.d.ts`, so `env.TENERO_API_KEY` typechecks directly. The cast is probably a carryover from when it wasn't declared there:

```suggestion
function lookupTeneroApiKey(env: CloudflareEnv): string | undefined {
  const key = env.TENERO_API_KEY;
  return typeof key === "string" && key.length > 0 ? key : undefined;
}
```

**[nit] Double KV read on `K_TENERO` per tick**

`runScheduledTasks` reads `K_TENERO` to check `lastRunAt` and `nextRunAfter`. Then `runTeneroNow` reads `K_TENERO` again for `consecutiveFailures`. Two reads per tick on the same key. Not a performance problem at current cadence, but if `runTeneroNow` accepted the pre-read state as a parameter it would be one read. Mentioning for Phase 1 when you'll add a third task and the pattern will multiply.

**[question] Scheduler state in `VERIFIED_AGENTS` namespace**

The `scheduler:*` keys live in the same KV namespace as agent/price data. This works today, but if `VERIFIED_AGENTS` ever needs namespace-level operations (list, bulk delete, quotas) the two concerns are interleaved. Is there a reason not to use a dedicated `SCHEDULER_STATE` KV binding? Or is the goal to stay within the existing binding count? Happy to take it as-is — just curious about the deliberate choice.

**Code quality notes:**
- `bumpFailure` is clean and generic — could be reused for Phase 1's earnings task with zero changes
- The parallel `Promise.all` in `readSchedulerStatus` (3 KV reads) and in `runScheduledTasks` (2 KV reads) is correct

**Operational context:** We watch `/api/admin/scheduler` via sensor. The status shape change (removing `name` from GET response) is non-breaking for our tooling — we read `status.lastTeneroRunAt` directly. Good.

---

*Performance:* No new React component patterns, waterfall risk, or bundle changes — the `leaderboard/page.tsx` change is pure deletion. ✓  
*Composition:* No new component architecture. ✓  
*UI/Accessibility:* No UI changes. ✓

biwasxyz:
Thanks for the thorough review @arc0btc — all three suggestions addressed in the latest push:

- **Stale `SchedulerDO` doc refs** — fixed §4 (diagram now shows `Cron Trigger → worker.ts scheduled() → runScheduledTasks`), §5 table, §10 (host is now `lib/scheduler/cron-runner.ts`), and §14 Phase 1. A Phase 1 implementer won't be sent looking for a deleted class.
- **`lookupTeneroApiKey` cast** — dropped; `env.TENERO_API_KEY` typechecks directly now that it's declared in `cloudflare-env.d.ts`. Good catch — it was a carryover from `worker.ts` before the type existed.
- **Double `K_TENERO` read** — `runTeneroNow` now takes an optional pre-read state; the cron tick passes the state it already fetched for the due-check, so it's one read on the hot path. Admin `refresh` still calls it without the arg and self-reads. Set the pattern now so it doesn't multiply when Phase 1 adds the earnings task, as you flagged.

**On the namespace question** — it's deliberate, leaning toward "stay within the existing binding count." The scheduler state is 3 low-volume keys, and `VERIFIED_AGENTS` already holds the Tenero `tenero:price:*` cache the scheduler itself writes, so scheduler bookkeeping living alongside it is consistent; the `scheduler:*` prefix namespaces it logically. A dedicated `SCHEDULER_STATE` binding for 3 keys felt like over-provisioning. If we ever need namespace-level ops (list / bulk-delete / per-namespace quota), splitting is a one-line wrangler binding + a trivial key move — happy to do it then, or now if you'd prefer the isolation up front.

`tsc` clean on changed files, scheduler tests green.

PR #980, opened 2026-06-08
chore(scheduler): delete SchedulerDO (v4 migration, manual deploy)

biwasxyz:
Follow-up to #979. Fully removes the neutered `SchedulerDO` now that the cron trigger drives all scheduled work. Rebased onto `main` after #979 was squash-merged.

## Changes (3 files, +33/-52)
- `worker.ts` — remove the neutered class + `DurableObject` import
- `wrangler.jsonc` — empty `durable_objects.bindings` + add `v4 deleted_classes` (top-level + `env.production`); cron triggers untouched
- `cloudflare-env.d.ts` — drop the `SCHEDULER` binding type

## Deploys automatically on merge — no manual step
Earlier I thought this needed a manual deploy; that was wrong. The `10211` error only happens on the **per-PR preview build** (`wrangler versions upload`, which rejects DO migrations). The **production deploy on merge to `main` uses `wrangler deploy` (non-versioned)**, which applies DO migrations fine — proven by the existing `v1/v2/v3` SchedulerDO migrations having been applied through the same pipeline (PRs #743/#772).

So: **merging this to `main` applies the `v4` deletion automatically.** The only visible artifact is this PR's **Cloudflare preview check showing red** (expected, `10211`) — `main` isn't branch-protected, so it doesn't block the merge.

## Merge order
1. #979 merged — cron live, scheduler fixed.
2. Confirm cron is firing (`/api/admin/scheduler` -> fresh `lastTeneroRunAt`).
3. Merge this PR -> production `wrangler deploy` tears down the DO. Done, no terminal.

Until then the neutered DO from #979 is inert (never re-instantiated), so there's no rush.

### DIFF
--- a/cloudflare-env.d.ts
+++ b/cloudflare-env.d.ts
@@ -22,11 +22,9 @@ interface CloudflareEnv {
   X402_RELAY?: import("./lib/inbox/relay-rpc").RelayRPC; // x402 sponsor relay RPC service binding (undefined in local dev)
   INBOX_RECONCILIATION_QUEUE?: Queue<import("./lib/inbox/reconciliation-queue").InboxReconciliationQueueMessage>;
   // Scheduled background work runs from a Cloudflare Cron Trigger (see
-  // worker.ts `scheduled()` + lib/scheduler/cron-runner.ts). The SchedulerDO
-  // binding is retained but neutered — the class can't be deleted on the
-  // versioned-upload deploy path (CF error 10211) — and nothing in the app
-  // calls it, so it has no RPC surface. Declared as a plain namespace.
-  SCHEDULER?: DurableObjectNamespace;
+  // worker.ts `scheduled()` + lib/scheduler/cron-runner.ts), not a Durable
+  // Object — SchedulerDO is removed (v4 deleted_classes migration), so there
+  // is no SCHEDULER binding.
   TENERO_REFRESH_ENABLED?: "true"; // Explicit production-only gate for the cron Tenero refresh
   TENERO_API_KEY?: string; // Optional Tenero API key (x-api-key header); raises rate limits above the shared web-ui-ip tier
 }
--- a/worker.ts
+++ b/worker.ts
@@ -3,7 +3,6 @@ import openNextWorker, {
   DOQueueHandler,
   DOShardedTagCache,
 } from "./.open-next/worker.js";
-import { DurableObject } from "cloudflare:workers";
 import {
   createConsoleLogger,
   createLogger,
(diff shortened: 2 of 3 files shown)


arc0btc (approved):
Retires `SchedulerDO` cleanly — replaces the fragile leaderboard-traffic-dependent alarm with a guaranteed Cloudflare Cron Trigger, then stages the final teardown as a `v4 deleted_classes` migration that requires a manual deploy. Solid two-step approach.

**What works well:**
- The root cause diagnosis is exactly right: a DO alarm that only arms when something pokes it is not a scheduler, it's a heartbeat with a critical single point of failure. Cron Trigger is the correct fix.
- `runScheduledTasks` parallelises its two KV reads at the top (`Promise.all`) — matches `async-parallel` best practice.
- `readSchedulerStatus` also parallelises all three reads. Clean.
- The `prevState?: TeneroState | null` optimisation in `runTeneroNow` (avoids a double KV read when the caller already has the state) is a nice touch and well-commented.
- Deploy note is front-and-centre in the PR description and the wrangler.jsonc comment. Anyone following the merge checklist won't miss the manual step.
- `env.preview` correctly omits the cron trigger — no risk of preview deploys consuming production quota.
- `bumpFailure` spreads `prev ?? {}` before overwriting, so `lastRunAt` + `result` survive an error cycle. Good.

**[question] Competition `consecutiveFailures` always resets to 0 on success** (`lib/scheduler/cron-runner.ts`)

`runTeneroNow` calculates consecutive failures with nuance (partial failures increment the counter; a clean success resets it). `runCompetitionNow` hardcodes `consecutiveFailures: 0` on any successful return. Intentional simplification because competition has binary pass/fail semantics, or an oversight?

**[question] `nextAlarmAt: null` in status shape** (`lib/scheduler/rpc-types.ts:22`)

The field is kept for "status-shape stability" but will always be `null` now. If the admin dashboard or any external monitoring renders it as "next scheduled at: —" that's fine. If any caller checks `nextAlarmAt !== null` as a liveness signal, it'll now permanently show dormant. Worth a check before the deploy, or a deprecation comment in the shape.

**[nit]** `now: number = Date.now()` in `runScheduledTasks` is evaluated at call time, not module load — which is what you want for testability. Just noting it reads correctly.

**Code quality notes:**

The `bumpFailure` helper is clean but weakly typed — it reads as `{ consecutiveFailures?: number }` and spreads `prev` into `next`, which means unknown keys from a `TeneroState | CompetitionState` pass through silently. Works fine, but a typed overload or a shared `{ consecutiveFailures: number }` interface would make the contract explicit.

**Operational context:**

From our production monitoring: the `lastTeneroRunAt` / `lastCompetitionRunAt` staleness check in `/api/admin/scheduler` has been the primary liveness indicator. The KV-backed status path preserves those fields, so our existing alerting around stale timestamps should survive the migration without changes.

The concurrency cap (≤5 simultaneous Hiro connections) is documented in the earnings architecture doc — worth wiring an assertion or at least a lint comment in `cron-runner.ts` when the earnings task lands, so future callers don't accidentally fan out to 6+.

Merge order in the PR description is correct: #979 first, verify cron liveness, then this PR + manual deploy.

PR #981, opened 2026-06-08
feat(earnings): Phase 1 indexer core — schema + classify + price (#978)

biwasxyz:
Phase 1 of the verified earnings ledger (#978): the backend indexer core. Builds on the Phase 0 cron scheduler. **Ships dormant** — zero behaviour change until explicitly enabled.

## What's here
A cron-driven, **indexer-only** sweep (no agent self-report) that scans confirmed inbound **sBTC / STX / aeUSDC** transfers to every registered agent, classifies by counterparty, prices in USD at index time, and writes an idempotent line-item ledger.

| Area | File(s) |
|---|---|
| Schema | `migrations/020_agent_earnings.sql` — `agent_earnings` (PK `tx_id,event_index`) + `earnings_index_state` (HWM + backfill offset) |
| Ingest | `lib/earnings/ingest.ts` — Hiro `transactions_with_transfers` (inline transfers → **no per-tx detail fetch, no memo parsing**) |
| Classify | `lib/earnings/classify.ts` — **txid-matched against our own D1**: `inbox_messages.payment_txid` (confirmed) / `bounties.paid_txid` (winner) / `agents` (peer) |
| Price | `lib/earnings/price.ts` — Tenero cache for sBTC/STX, `$1` peg for aeUSDC, index-time |
| Persist | `lib/earnings/d1.ts` — `INSERT OR IGNORE`, per-agent state, cursor |
| Orchestrate | `lib/earnings/indexer.ts` — cursor sweep, per-agent backfill→incremental, concurrency 5 |
| Wire-in | `lib/scheduler/cron-runner.ts` — `runEarningsNow()` on a 30-min cadence + admin status/refresh |

## Cost controls (the whole point)
- **Per-agent high-water mark** → zero Hiro calls for an agent with nothing new.
- **`transactions_with_transfers`** → ~half the Hiro calls of list-then-fetch-each-tx.
- **Bounded backfill** (4 pages/agent/tick), **slice 25 agents/tick**, **concurrency ≤5** (under CF's 6-connection cap).
- **Idempotent writes**; read-time aggregation + edge cache come in Phase 3.
- **Ships dormant behind `EARNINGS_INDEX_ENABLED`** — merge + deploy do nothing; flip on when you're watching the bill, flip off instantly to kill.

## Classifiers
`x402_endpoint` is **inert** — confirmed there's no enumerable x402 payTo catalog (per-agent dynamic payTo). The 3 active classifiers **inbox_message + bounty + agent_peer** meet the #978 DoD (≥3). Anti-gaming (`self_funded`/`ring`) is Phase 2; the `excluded_reason` column already exists so it's a data change, not a schema change.

## Deploy / verify steps
1. Merge → deploys dormant (no behaviour change).
2. Apply the D1 migration: `wrangler d1 migrations apply landing-page --remote` (tables are `IF NOT EXISTS`, idempotent).
3. **Verify without enabling the cron:** `POST /api/admin/scheduler?action=refresh&task=earnings` — force-runs one sweep past the gate and returns the summary (agentsScanned / transfersFound / inserted / earningRows / bySourceClass).
4. When happy, enable the cron: set `EARNINGS_INDEX_ENABLED=true` and deploy.

## Tests / checks
`tsc` clean, `next lint` clean, full vitest **1495 passing** (+13 new in `lib/earnings/__tests__/earnings.test.ts` covering asset mapping, inbound extraction/event-index stability, all 4 classifier branches, and pricing incl. the $1 peg + unpriced fallback).

Part of #978. Next: Phase 2 (anti-gaming) → Phase 3 (public API) → Phase 4 (UI).

### DIFF
--- a/app/api/admin/scheduler/route.ts
+++ b/app/api/admin/scheduler/route.ts
@@ -10,7 +10,12 @@ import {
 } from "@/lib/scheduler/cron-runner";
 import type { SchedulerTask } from "@/lib/scheduler/rpc-types";
 
-const ALLOWED_TASKS = new Set<SchedulerTask>(["tenero", "competition", "all"]);
+const ALLOWED_TASKS = new Set<SchedulerTask>([
+  "tenero",
+  "competition",
+  "earnings",
+  "all",
+]);
 
 function json(body: unknown, init: ResponseInit = {}) {
   const headers = new Headers(init.headers);
@@ -31,7 +36,7 @@ function schedulerTask(url: URL): SchedulerTask | NextResponse {
   const task = url.searchParams.get("task") || "tenero";
   if (!ALLOWED_TASKS.has(task as SchedulerTask)) {
     return json(
-      { error: "Unsupported task. Use tenero, competition, or all." },
+      { error: "Unsupported task. Use tenero, competition, earnings, or all." },
       { status: 400 }
     );
   }
--- a/cloudflare-env.d.ts
+++ b/cloudflare-env.d.ts
@@ -27,4 +27,5 @@ interface CloudflareEnv {
   // is no SCHEDULER binding.
   TENERO_REFRESH_ENABLED?: "true"; // Explicit production-only gate for the cron Tenero refresh
   TENERO_API_KEY?: string; // Optional Tenero API key (x-api-key header); raises rate limits above the shared web-ui-ip tier
+  EARNINGS_INDEX_ENABLED?: "true"; // Gate for the cron earnings indexer (issue #978); ships dormant, set "true" to enable
 }
--- a/docs/earnings-ledger-architecture.md
+++ b/docs/earnings-ledger-architecture.md
@@ -356,14 +356,15 @@ would multiply Hiro calls by the cadence with no freshness benefit for a 30-day 
   competition from a Cloudflare Cron Trigger via `worker.ts` `scheduled()` +
   `lib/scheduler/cron-runner.ts` (KV-backed state). Fixes the dead/fragile scheduler at
   the root; prerequisite + trigger for the earnings indexer.
-- **Phase 1 — Schema + indexer core (backend only).** Migration `020`; `lib/earnings/`
-  ingest + classify + index-time pricing + idempotent writes; `runEarnings()` task in
-  `lib/scheduler/cron-runner.ts` (`runScheduledTasks`). Verify via admin/dry-run.
-  **Indexer-only — no agent-submit/self-report
-  path** (unlike competition's `source='agent'` fast-path; earnings tolerate indexing
-  lag and a submit endpoint is needless attack surface). **Also: investigate the x402
-  payTo catalog here** — if none exists, `x402_endpoint` launches inert (bounty + inbox
-  + peer only).
+- **Phase 1 — Schema + indexer core (backend only). DONE.** Migration `020`
+  (`agent_earnings` + `earnings_index_state`); `lib/earnings/` ingest (cheap
+  `transactions_with_transfers`) + classify + index-time pricing + idempotent writes;
+  `runEarningsNow()` wired into `lib/scheduler/cron-runner.ts` on a 30-min cadence,
+  gated by `EARNINGS_INDEX_ENABLED` (**ships dormant**). Verify via
+  `POST /api/admin/scheduler?action=refresh&task=earnings` (force-runs a sweep past the
+  gate). **Indexer-only — no agent-submit/self-report path.** **x402 catalog confirmed
+  absent** (per-agent dynamic payTo, no registry) → `x402_endpoint` is inert; the 3
+  active classifiers are **inbox_message + bounty + agent_peer** (meets the DoD ≥3).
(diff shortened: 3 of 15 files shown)


arc0btc (approved):
Phase 1 of the verified earnings ledger — a backend-only, cost-controlled indexer that ships dormant behind \`EARNINGS_INDEX_ENABLED\`. Solid work overall: the round-robin cursor, HWM/backfill split, idempotent ledger, and dormant-gate pattern all match the competition scheduler's conventions and are well-suited to CF's constraints.

**What works well:**
- `INSERT OR IGNORE` on `(tx_id, event_index)` is the right idempotency key — survives cron re-runs and crash-recovery without double-counting
- `transactions_with_transfers` over list-then-fetch is exactly the right Hiro optimization; halves call count and eliminates the per-tx waterfall
- `mapPool` bounded-concurrency implementation is correct — JS single-threaded `next++` is safe, and staying at ≤5 respects CF's 6-connection cap
- `EARNINGS_INDEX_ENABLED` dormant gate is threaded correctly: cron path respects it, admin force-refresh bypasses it — exactly right for ops verification without enabling production traffic
- Test coverage is comprehensive: assets / ingest / classify / price all have dedicated unit tests with realistic mocks
- The `encodeURIComponent(stxAddress)` on the Hiro URL is the kind of quiet correctness detail that prevents subtle bugs

---

**[suggestion] Parallelize classify + price in `resolveRow`** (`lib/earnings/indexer.ts:703`)

Classification (3 sequential D1 reads) and pricing (1 KV read) are completely independent. Right now they run sequentially, which adds latency for each transfer. With `Promise.all` they'd overlap:

```suggestion
  const [classification, pricing] = await Promise.all([
    classifyTransfer(db, transfer),
    priceTransfer(kv, transfer, now),
  ]);
```

At 25 agents × up to 200 transfers/tick in steady state, this could meaningfully reduce per-tick wall time.

**[suggestion] Earnings failure count missing from `SchedulerStatus`** (`lib/scheduler/rpc-types.ts`, `lib/scheduler/cron-runner.ts`)

`bumpFailure(kv, K_EARNINGS)` is called on error, but `SchedulerStatus.consecutiveFailures` still only declares `{ tenero: number; competition: number }`. The bump will write an untyped field to the KV blob that nothing reads. Operationally this means earnings errors are only visible in logs, not in the admin status endpoint — which is where ops looks first. Suggest either:
- Add `earnings: number` to `consecutiveFailures` and read it in `readSchedulerStatus`, or
- At minimum note the gap in a TODO comment so Phase 2 doesn't inherit silent failure state

**[question] `registered_wallets` vs `agents` table for peer classification** (`lib/earnings/d1.ts:614`, `lib/earnings/classify.ts:401`)

`fetchAgentPage` reads from `registered_wallets.stx_address`; `classifyTransfer` queries `agents.stx_address` for `agent_peer` detection. Are these guaranteed to be the same population? If an address is in `registered_wallets` but not in `agents`, transfers *from that agent* to the indexed agent would fall through to `unclassified` instead of `agent_peer`. Probably fine given how registration works, but worth confirming the two tables stay in sync.

**[question] `bumpFailure` + `EarningsState` shape** (`lib/scheduler/cron-runner.ts:1299`)

`EarningsState` has no `consecutiveFailures` field. Does `bumpFailure` splat a foreign key into the blob (harmless if nothing reads it) or does it try to re-read a typed state object and get confused on the next tick? If the former, it's fine but worth documenting. If the latter, a JSON parse issue on the tick after a failure would surface as a second error.

**[nit]** `indexer.ts:701` — `transfer: import("./types").InboundTransfer` uses an inline type import. `InboundTransfer` could be added to the named import at the top of the file alongside `EarningRow`, `EarningsSweepSummary`, `SourceClass`.

**[nit]** `runEarningsNow` calls `Date.now()` twice — once passed to `runEarningsSweep` and once for `lastRunAt` in the KV put. The two timestamps will differ by the sweep duration. If `now` is already being threaded down for this reason, it could be captured once at the top of the function.

---

**Code quality notes:**
- `collect` as a mutable closure inside `indexAgent` (captured `rows`, `transfersFound`, `maxBlock`) is compact but makes the control flow harder to follow at a glance. Not blocking — just a future refactor target if `indexAgent` grows.
- `zeroBySource()` duplicates the `SourceClass` values as a hardcoded object. If a new source class lands in Phase 2, this will silently undercount in the summary. Consider deriving it from `EARNING_SOURCE_CLASSES` / `EXCLUDED_SOURCE_CLASSES` constants at runtime.

---

*Operational note:* We run the Tenero KV cache at 24h TTL for sBTC/STX pricing. If the indexer indexes a transfer during a Tenero outage, `priceSource = "none"` with `amount_usd = null` is correct behavior and the comment notes Phase 3 will reprice — but there's no re-pricing task yet and no mention of it in the Phase 3 scope. Worth flagging in the architecture doc so it doesn't get lost between phases.

Overall this is solid Phase 1 infrastructure. The dormant gate means none of the Phase 1 edge cases are user-visible until explicitly enabled, which is the right way to ship something this foundational. LGTM.

PR #982, opened 2026-06-08
feat(earnings): Phase 2 anti-gaming — first-funder, ring, alt-address (#978)

biwasxyz:
Phase 2 of the verified earnings ledger (#978). Excludes self-dealing from earnings **before it counts**, so the new metric can't be gamed the way trade-count was. Builds on the merged Phase 1 indexer.

## What's here
Every freshly-classified **`agent_peer`** earning now runs through exclusion checks (inbox/bounty skip it — they're already proven legit by their txid match against our own D1).

| Rule | Excludes when |
|---|---|
| **manual override** | operator flagged the line item `exclude` / `include` / `reclassify` (applies to any class) |
| **alt-address** | sender & recipient agents share an `owner` (X handle) → `self_funded` |
| **self-funded** | sender & recipient share a **first-funder** address → `self_funded` |
| **ring** | a prior `A→B→A` reverse leg exists within 14d for a similar amount → both legs `ring` |

- `migrations/021_earnings_anti_gaming.sql` — `address_first_funder` cache + `earnings_manual_override`
- `lib/earnings/anti-gaming.ts` — the engine, wired into the indexer's `resolveRow`

## Cost (the first-funder lookup is the only new Hiro cost)
First-funder is **immutable**, so it's cached forever in `address_first_funder`: **≤2 Hiro calls per address EVER**, and only triggered for `agent_peer` transfers. The lookup reuses the existing `transactions_with_transfers` helper (fetch the oldest tx, take its first inbound sender). Failed lookups retry after 1h; `ok`/`none` are permanent. All other checks are plain D1 queries.

## Notes / limits
- Heuristics run **only on `agent_peer`** — inbox/bounty earnings are untouched (txid-proven).
- Exclusion is conservative: self-funded only fires on **two confident `ok`** first-funder lookups (uncertainty ⇒ keep the earning).
- **Known limit:** a ring whose two legs land in the *same* sweep tick isn't caught (the reverse leg isn't persisted yet); cross-tick rings are. Documented in the architecture doc.

## Checks
`tsc` clean · `lint` clean · full suite **1504 passing** (+9 new in `anti-gaming.test.ts` covering all four rules, the non-`agent_peer` passthrough, differing-funder non-exclusion, and the clean passthrough).

Part of #978. Next: **Phase 3** (public read API) → **Phase 4** (leaderboard flips to earnings — the visible win).

### DIFF
--- a/docs/earnings-ledger-architecture.md
+++ b/docs/earnings-ledger-architecture.md
@@ -365,8 +365,13 @@ would multiply Hiro calls by the cadence with no freshness benefit for a 30-day 
   gate). **Indexer-only — no agent-submit/self-report path.** **x402 catalog confirmed
   absent** (per-agent dynamic payTo, no registry) → `x402_endpoint` is inert; the 3
   active classifiers are **inbox_message + bounty + agent_peer** (meets the DoD ≥3).
-- **Phase 2 — Full anti-gaming.** first-funder cache + self-fund exclusion, two-hop
-  ring, alt-address, manual override.
+- **Phase 2 — Full anti-gaming. DONE.** `migration 021` (`address_first_funder` cache
+  + `earnings_manual_override`); `lib/earnings/anti-gaming.ts` — manual override,
+  alt-address (shared `owner`), self-funded (shared first-funder, cached forever),
+  two-hop ring (`A→B→A` ≤14d, similar amount, flips both legs). Runs only on
+  `agent_peer` earnings, wired into the indexer's `resolveRow`. Known limit: a ring
+  whose two legs land in the *same* sweep tick isn't caught (the reverse leg isn't
+  persisted yet); cross-tick rings are.
 - **Reprice pass (Phase 3 scope).** Phase 1 stores `amount_usd = NULL`,
   `price_source = 'none'` for transfers indexed during a Tenero gap. There is **no
   reprice task yet** — add one in Phase 3 (a bounded sweep over `price_source = 'none'`
--- a/lib/earnings/__tests__/anti-gaming.test.ts
+++ b/lib/earnings/__tests__/anti-gaming.test.ts
@@ -1,0 +1,134 @@
+import { describe, it, expect, vi } from "vitest";
+import { applyAntiGaming } from "../anti-gaming";
+import type { Classification, InboundTransfer } from "../types";
+
+const SENDER = "SP_SENDER";
+const RECIPIENT = "SP_RECIPIENT";
+
+const noopLogger = {
+  info: vi.fn(),
+  warn: vi.fn(),
+  error: vi.fn(),
+  debug: vi.fn(),
+} as unknown as import("../../logging").Logger;
+
+function transfer(overrides: Partial<InboundTransfer> = {}): InboundTransfer {
+  return {
+    txId: "0xtx",
+    eventIndex: 0,
+    senderStx: SENDER,
+    recipientAgentStx: RECIPIENT,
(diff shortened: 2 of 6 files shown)


arc0btc (approved):
Phase 2 anti-gaming engine for the earnings ledger: manual overrides, alt-address detection (shared owner), self-funded detection (shared first-funder), and two-hop ring detection — wired into `resolveRow` for all freshly-classified `agent_peer` transfers.

**What works well:**
- Conservative design throughout: uncertainty → keep the earning, never false-exclusion. Only two confident `ok` first-funder lookups trigger self-funded exclusion; a failed cache entry doesn't block.
- Correct parallel fetching for first-funder lookups: `Promise.all([getFirstFunder(sender), getFirstFunder(recipient)])` — clean.
- Test coverage is solid: 9 cases covering all four rules, the non-`agent_peer` passthrough, differing-funder non-exclusion, and the clean pass-through. The mock DB pattern via `sql.includes()` is a bit fragile but acceptable at this scale.
- Cache-forever for first-funder is the right call — immutable data, no need to expire.
- Clean module separation — `anti-gaming.ts` is easy to unit-test in isolation, doesn't bleed into the indexer.
- `markRing` retroactively flags the reverse leg immediately; the current leg is excluded via the returned Classification. Both legs end up excluded, which is correct.

**[question] Ring query and indexes** (`lib/earnings/anti-gaming.ts:175`)
The `findReverseLeg` query filters on `recipient_agent_stx, sender_stx, asset, amount_raw, block_time`. Does migration 020 put a composite index on `agent_earnings` covering the sender/recipient columns? Without one this is a full-table scan per transfer. At volume (many agents × many transfers per sweep) that adds up. If there's no index covering `(recipient_agent_stx, sender_stx)` at minimum, worth adding one here in 021.

**[suggestion] Silent no-op when `reclassify` has null `new_source_class`** (`lib/earnings/anti-gaming.ts:207`)
The DB allows `new_source_class TEXT` nullable, so a malformed override row with `action='reclassify'` and `new_source_class=NULL` silently falls through to heuristics. The result isn't wrong (conservative), but a warn log would make future debugging much easier:

```suggestion
    if (override.action === "reclassify" && override.new_source_class) {
      const sc = override.new_source_class as SourceClass;
```
→
```suggestion
    if (override.action === "reclassify") {
      if (!override.new_source_class) {
        logger.warn({ txId: transfer.txId, eventIndex: transfer.eventIndex }, "reclassify override missing new_source_class — falling through to heuristics");
      } else {
        const sc = override.new_source_class as SourceClass;
```
(close the extra `else` block after the existing `return`)

**[nit]** `inbound[0].senderStx` in `fetchFirstFunderFromChain` silently picks the first inbound sender of the oldest tx. Covers 99%+ of real addresses, but a short comment noting "first inbound of genesis tx is almost always the sole funder" would help a future reader who wonders about multi-sender genesis blocks.

**Code quality notes:**
- `exclude()` helper is a nice touch — intent reads clearly at the call site.
- SQL is parameterized throughout, no injection risk.
- `extractInboundTransfers` reuse from `ingest` is the right call — no duplication.
- Migration follows naming conventions; the `CHECK (lookup_status IN ...)` and `CHECK (action IN ...)` constraints are good data quality guardrails.

*Operational note:* We monitor the Hiro API budget closely after a recent circuit-breaker was added for budget storms (PR #958). The first-funder cache design (≤2 calls per address ever) is the right shape — no per-transfer Hiro load once the cache is warm.

PR #984, opened 2026-06-08
feat(earnings): count earnings from join date (verified_at floor) (#978)

biwasxyz:
Scopes earnings to **money earned as an agent**, not lifetime personal on-chain history. Only inflows with `block_time >= agents.verified_at` (registration) count.

## Why
Surfaced reviewing the metric semantics: of the three classifiers, `inbox_message` and `bounty` are already post-join (they match platform DB records that only exist after registration) — but **`agent_peer` had no time bound**. A personal transfer between two addresses *years before* either registered would today be classified `agent_peer` and counted as "earnings" — wrong, and trivially gameable (register two old addresses that historically moved money between them).

## What
- `fetchAgentPage` now returns `verified_at` parsed to unix seconds (`verifiedAtSec`); `0` = no floor (all-time fallback when `verified_at` is missing/unparseable).
- `indexAgent` applies the floor in `collect()` (skips `block_time < verifiedAtSec`) and **terminates backfill on the first pre-join page** — so backfill stops at registration instead of walking to genesis. More correct *and* cheaper Hiro usage.

## Reversibility
Flippable to all-time later via a re-backfill (reset `earnings_index_state.backfill_complete`/`backfill_offset`; the sweep re-walks history). And **free to change right now** — the indexer is dormant and no data has accumulated, so this is a no-cost decision until enabled.

## Notes
- Touches Phase 1 indexer code on `main`; independent of the open Phase 3 API PR (#983) — different files.
- `tsc` + lint clean; earnings tests green (+2 new in `join-date.test.ts` for the `verified_at` → floor parsing incl. null/unparseable fallback).

Part of #978.

### DIFF
--- a/docs/earnings-ledger-architecture.md
+++ b/docs/earnings-ledger-architecture.md
@@ -31,6 +31,7 @@ because they require real third-party on-chain value transfer.
 | USD pricing source | **Tenero + last-good fallback** | only ~3 tokens; already cached in KV |
 | Ingestion shape | **Separate earnings task, own slow cursor** | clean separation from competition sweep |
 | Ingestion model | **Indexer-only (pull), all agents — no agent self-report** | earnings must be derived from chain truth, never asserted by the agent (anti-gaming); covers the dormant cohort that would never submit |
+| Earnings window | **From join date** — only inflows with `block_time >= agents.verified_at` count | "earnings" = earned *as an agent*, not lifetime personal history; closes the `agent_peer` gap where pre-platform transfers between later-registered addresses would count; also bounds backfill (stops at registration, not genesis). Flippable to all-time via a re-backfill (reset `earnings_index_state`); free to change while the indexer is dormant. |
 | Scheduling host | **Cloudflare Cron Trigger; SchedulerDO retired (Phase 0, done)** | see §3 — the DO alarm was request-bootstrapped and silently died |
 | Earnings leaderboard surface | **Trading board (`app/leaderboard`)** | this is the board whose trade-count default got gamed |
 | x402_endpoint classifier | **Investigate catalog in Phase 1** | may launch inert if no payTo catalog exists |
--- a/lib/earnings/__tests__/join-date.test.ts
+++ b/lib/earnings/__tests__/join-date.test.ts
@@ -1,0 +1,34 @@
+import { describe, it, expect } from "vitest";
+import { fetchAgentPage } from "../d1";
+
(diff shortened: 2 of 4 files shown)


arc0btc (approved):
Scopes earnings to post-registration inflows — closes a real semantic gap where pre-platform `agent_peer` transfers between later-registered addresses would have inflated earnings. Clean implementation.

**What works well:**
- The `verifiedAtSec: 0` fallback correctly degrades to all-time for agents without `verified_at` — no floor applied, no data loss.
- `reachedPreJoin` terminates backfill early rather than walking to genesis — correct *and* cheaper Hiro usage.
- Field alignment verified: `reachedPreJoin` reads `r.tx?.burn_block_time` and `extractInboundTransfers` maps the same `tx.burn_block_time` to `t.blockTime` — the two checks are consistent.
- Tests cover ISO parse, null, and unparseable cases. Good edge-case discipline.
- Clean separation from #983 (Phase 3 API PR) — different files, no entanglement.

**[nit] `transfersFound` semantic shift** (`lib/earnings/indexer.ts`)
The original code did `transfersFound += transfers.length` (all extracted transfers). The new code only increments for accepted transfers, so `transfersFound` now means "earnings-eligible transfers counted" rather than "transfers found on-chain". If this counter feeds logs or metrics, the label might become misleading. Harmless for correctness — worth a rename (`earningsRows` or `transfersCounted`) if it surfaces in dashboards.

**[nit] Epoch-0 corner case** (`lib/earnings/d1.ts`)
`Math.floor(Date.parse(r.verified_at) / 1000) || 0` — if `verified_at` is exactly `"1970-01-01T00:00:00Z"`, `Date.parse` returns `0`, and `0 || 0` = `0`, treating a valid epoch-0 registration as "no floor". Irrelevant in practice, but clarifying with a comment or an explicit `Number.isFinite(s) && s > 0` guard would make the intent unambiguous. Current behavior is fine.

**Code quality notes:**
- Multi-line parameter comment on `verifiedAtSec` in `indexAgent` is verbose but the 0-means-all-time invariant is non-obvious enough to justify it.
- `reachedPreJoin` as a named inner closure reads well — cleaner than inlining.

*Operational note:* We track Hiro API row-read costs and have seen unbounded backfill walks add up quickly (CF DO free tier is 5M rows/day). The early-termination via `reachedPreJoin` will meaningfully reduce costs once the indexer is enabled — good call.

PR #992, opened 2026-06-09
chore(earnings): relax sweep cadence 5min → 30min (#978)

biwasxyz:
Backfill is complete (**1005/1005 indexed**, 573 earners, $23.3k verified), so the indexer only needs to catch new inflows incrementally now.

- **`EARNINGS_INTERVAL_MS` 5 min → 30 min.** The 5-min cadence was a temporary rollout setting to fill the board fast; 30 min is the steady-state value (~6× less D1/Hiro work).

The cron trigger still fires every 5 min; this just gates the earnings sweep to every 30 min (Tenero/competition tasks unaffected). One-line change, 36 scheduler/earnings tests pass. Part of #978.

### DIFF
--- a/lib/earnings/constants.ts
+++ b/lib/earnings/constants.ts
@@ -23,10 +23,10 @@ export const EARNINGS_HIRO_PAGE_LIMIT = 50;
  *  day-one indexing can't spike Hiro usage; deep history fills over many ticks. */
 export const EARNINGS_MAX_PAGES_PER_AGENT = 4;
 
-// How often the earnings sweep is due (ms). 5 min during the initial backfill
-// rollout = every cron tick (the fastest the every-5-min cron trigger allows)
-// so the roster backfills quickly. Relax once fully backfilled.
-export const EARNINGS_INTERVAL_MS = 5 * 60 * 1000;
+// How often the earnings sweep is due (ms). 30 min in steady state — the full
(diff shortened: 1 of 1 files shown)


arc0btc (approved):
Relaxes the earnings sweep gate from 5 min to 30 min now that the full roster backfill is complete — exactly the right time to do this.

**What works well:**
- The constant change itself is correct: `30 * 60 * 1000` = 1,800,000 ms = 30 min steady state
- Comment updated in-place with both current intent and historical context — future readers won't wonder why 5 min was ever chosen
- PR body gives a clear backfill completion receipt (1005/1005 indexed, 36 tests pass) — the decision is well-evidenced, not arbitrary
- Cron trigger cadence is unchanged, so Tenero/competition tasks are unaffected as stated

**Operational context:** We track D1 and Hiro API quota usage closely. A 6× reduction in sweep frequency aligns directly with protecting the D1 row-read tier (5M/day shared between DO SQLite and D1). The Hiro budget storm circuit-breaker pattern merged earlier (#958) is the right partner to this — lower call frequency + circuit-breaker on surge = resilient steady state.

## Performance

Not applicable — `lib/earnings/constants.ts` is a constants file with no React, async, or rendering code.

## Composition

Not applicable — no components or context changed.

## UI/Accessibility

Not applicable — no UI or CSS changed.

---

Straightforward and safe. LGTM.

Issue #418, opened 2026-03-18
feat(achievements): add sBTC Holder achievement — holds any sBTC balance


arc0btc:
## Summary

Add an **sBTC Holder** achievement that grants when an agent holds any sBTC balance.

## Trigger

- Agent's Stacks address holds a non-zero sBTC balance
- Verified via sBTC SIP-010 contract balance call

## Implementation Notes

**Verification logic:**
```
GET https://api.hiro.so/v2/contracts/call-read/SM3VDXK3WZZSA84XXFKAFAF15NNZX32CTSG82JFQ4/sbtc-token/get-balance
// with principal arg = agent's Stacks address
```

Returns fungible token balance. Grant if `value > 0`.

Auto-grant during heartbeat. Can also backfill retroactively for existing agents.

## Registry Entry

```json
{
  "id": "sbtc-holder",
  "name": "sBTC Holder",
  "description": "Holds a non-zero sBTC balance — bridged Bitcoin to Stacks",
  "category": "onchain",
  "grant": "auto"
}
```

## Why This Matters

- Signals cross-layer sophistication — agent has bridged BTC to L2
- Currently rare: few agents hold sBTC, making it a prestige signal
- Bitcoin-native economic participation beyond just STX
- No friction for agents who already hold sBTC — auto-detected

## Context

Part of the Tier 2 recommendations from the achievements audit (#384). Arc's perspective from that thread: "sBTC activity — signals cross-layer sophistication. Few agents are here yet, which makes it high-prestige."

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #997, opened 2026-06-14
fix(deps): pin esbuild >=0.28.1 to resolve Dependabot alert #119 (high)

### LINKED ISSUE #119: Agent identity needs a unified design — too fragmented to scale
## The Problem

Agent identity on AIBTC is split across 5+ systems that don't talk to each other. As the network grows past the current 17 agents, this fragmentation becomes a real barrier to agent-to-agent collaboration, trust, and discovery.

Right now, to fully understand "who is this agent?", you need to:
1. Know their BTC address (the only universal key)
2. Call `/api/agents/<btc>` for their profile
3. Call `/api/verify/<btc>` for their level (slightly different response than #2)
4. Query the identity registry contract on-chain to find their agent-id (if they even have one)
5. Query the reputation registry with that agent-id for trust data
6. Check the leaderboard for activity ranking

That's 4+ calls across 2 systems (API + blockchain), with no guarantee the data connects. An agent trying to decide "should I trust this other agent?" has no straightforward path.

### 1. On-chain identity is disconnected from the platform
The docs mention `erc8004AgentId` as a field on agent profiles, but it's **missing from every API response** — `/api/agents`, `/api/verify`, `/api/leaderboard` all omit it. The platform can't tell you if an agent has on-chain identity without you querying the contract yourself.

Only 2 of 17 agents have registered on-chain (agent-ids 0 and 1). The rest have no on-chain anchor at all.

### 2. No reverse lookup on the identity contract
Given an STX address, there's no contract function to find the agent-id. The only path is to iterate through every agent-id calling `get-owner()` until you find a match. This doesn't scale — at 100 agents it's 100 read-only calls. At 1,000 it's unusable.

The contract needs a `get-agent-id-by-owner(principal)` function or a map-based reverse lookup.

### 3. Inbox messages are unauthenticated
When an agent receives an inbox message, `fromAddress` is a plain string with no cryptographic proof. There's nothing stopping agent A from sending a message that claims to be from agent B. In an ecosystem where agents make trust decisions, unauthenticated identity is dangerous.

### 4. No capability advertising
There's no way to know what an agent can do without trying it. Questions like:
- "Does this agent accept x402 inbox messages?"
- "Does this agent have a running heartbeat?"
- "Does this agent hold sBTC for payments?"
- "Does this agent respond to paid-attention?"

...can only be answered by trial and error. Agents can't discover collaborators by function.

### 5. No search or filtered discovery
The leaderboard is the only discovery mechanism, sorted by registration date. There's no way to search by name, filter by capability, or find agents by activity level. An agent looking for "active Genesis agents that accept inbox messages" has no query for that.

### 6. Display names aren't unique
Two agents could register with the same display name. Names can't be used as reliable identifiers, which means agents (and humans) fall back to raw addresses — defeating the purpose of having names.

### A. Unified identity resolution endpoint

One endpoint that accepts **any identifier** and returns the complete identity:

```
GET /api/resolve/<identifier>
```

Where `<identifier>` can be:
- BTC address: `bc1qyu22hyqr406pus0g9jmfytk4ss5z8qsje74l76`
- STX address: `SPKH9AWG0ENZ87J1X0PBD4HETP22G8W22AFNVF8K`
- Agent-id: `0` or `u0`
- Display name: `Tiny Marten` (if unique)
- BNS name: `tinymarten.btc`

Returns a single, complete identity object:

```json
{
  "identity": {
    "btcAddress": "bc1q...",
    "stxAddress": "SP...",
    "agentId": 0,
    "displayName": "Tiny Marten",
    "bnsName": null,
    "description": "...",
    "owner": "theendaoment"
  },
  "trust": {
    "level": 2,
    "levelName": "Genesis",
    "onChainIdentity": true,
    "reputationScore": 5.0,
    "reputationCount": 1,
    "verifiedAt": "2026-02-05T..."
  },
  "activity": {
    "lastActiveAt": "2026-02-13T...",
    "checkInCount": 286,
    "hasHeartbeat": true,
    "inboxEnabled": true,
    "achievements": ["sender"]
  },
  "capabilities": ["heartbeat", "inbox", "x402", "reputation"]
}
```

One call. Everything an agent needs to decide whether and how to interact.

### B. On-chain identity should be default, not optional

Right now ERC-8004 registration is an afterthought — an "optional enhancement." But it's the only thing that gives an agent a verifiable, decentralized identity that doesn't depend on aibtc.com being up.

**Proposal:** Make on-chain registration part of the standard registration flow. When an agent calls `POST /api/register`, the platform should either:
- Auto-register via sponsored transaction, or
- Return clear instructions that the agent should register immediately, with the sponsored tx API key ready to use

The `agent-id` should be the canonical short identifier across the platform, not something most agents never get.

(Related: #117 — ERC-8004 should be first action after wallet claim)

### C. Reverse lookup on the identity contract

Add a map to `identity-registry-v3` (or a companion contract):

```clarity
(define-map agent-id-by-owner principal uint)
```

Updated on `register` / `register-with-uri`. Enables:

```clarity
(get-agent-id-by-owner 'SPKH9AWG0ENZ87J1X0PBD4HETP22G8W22AFNVF8K)
;; => (some u0)
```

This is essential for any flow where you start with an address and need the agent-id (which is every flow).

### D. Capability tags

Let agents declare capabilities — either on-chain (via `set-metadata` on identity registry) or off-chain (via profile update API). A simple string array:

```json
["heartbeat", "inbox", "x402-payments", "reputation", "paid-attention"]
```

This enables filtered discovery: "find me all agents that accept inbox messages and have active heartbeats."

### E. Message authentication

Inbox messages should include a signature proving the sender owns the `fromAddress`. The flow already exists — agents sign challenges for check-ins and profile updates. Apply the same pattern to messages:

- Sender signs the message content with their BTC key
- Recipient verifies the signature against `fromAddress`
- Platform can optionally verify before delivery

Without this, agent-to-agent messaging has no authentication layer. Any agent can impersonate any other.

## Tradeoffs

| Decision | Option A | Option B | Recommendation |
|----------|----------|----------|----------------|
| **Primary key** | BTC address (current) | Agent-id (on-chain) | Transition to agent-id as canonical; BTC address becomes a lookup key |
| **Identity source of truth** | KV store (centralized, fast) | Blockchain (decentralized, slow) | Hybrid — KV for speed, chain for verification. KV syncs from chain. |
| **Capability registry** | On-chain metadata | Off-chain profile field | Off-chain first (faster iteration), migrate to on-chain later |
| **Name uniqueness** | Not enforced (current) | Enforced at registration | Enforce — names are useless if they're ambiguous |
| **Message auth** | No signatures (current) | Required BIP-137 signature | Required — unauthenticated messaging in a trust network is a design flaw |
| **Auto ERC-8004** | Manual (current) | Auto on registration | Auto via sponsored tx — remove the friction, make it default |

## What This Enables

With unified identity + capabilities + trust data in one place:

- **Agent collaboration**: "Find an active, trusted agent that accepts x402 payments" → one API call
- **Trust decisions**: "Should I pay 100 sats to message this agent?" → check reputation score and identity verification in the same response
- **Human understanding**: The dashboard can show a rich identity card instead of raw addresses
- **Network effects**: As agents can discover each other by capability, they start forming functional relationships — not just existing side by side on a leaderboard

## Current State for Context

- 17 registered agents (9 Genesis, 8 Registered)
- 2 with on-chain identity (agent-id 0: Tiny Marten, agent-id 1: Arc)
- 1 reputation entry on-chain (Tiny Marten → Arc)
- 5 agents with any check-in activity at all
- 0 agents with BNS names
- `erc8004AgentId` null in all API responses
- No capability advertising
- No message authentication
- No search functionality

This isn't a crisis — the network is small enough to fix this now. But the identity layer is the foundation everything else builds on. Getting it right early is much cheaper than retrofitting later.

arc0btc:
## Summary

- Adds `"esbuild": ">=0.28.1"` to the `overrides` field in `package.json`
- Resolves Dependabot alert #119: missing binary integrity verification in esbuild's Deno module (CVSS 8.1, no CVE assigned)
- esbuild was pulled in transitively by `wrangler`, `vite`, and `@opennextjs/aws` at versions <0.28.1

**Risk assessment:** Low actual risk for this project. The vulnerability is in esbuild's Deno distribution (`lib/deno/mod.ts`) and requires an attacker to control the `NPM_CONFIG_REGISTRY` env var. This project uses esbuild via Node.js only, so the vulnerable code path is never executed. The override is applied as a belt-and-suspenders measure and to clear the Dependabot alert.

## Test plan

- [x] `bun install` resolves `node_modules/esbuild` to 0.28.1
- [ ] CI passes (Next.js build + vitest suite)
- [ ] Dependabot alert #119 closes after merge

Closes #119 (Dependabot alert)

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -1,0 +1,1958 @@
+{
+  "lockfileVersion": 1,
+  "configVersion": 0,
+  "workspaces": {
+    "": {
+      "name": "aibtcdev-landing-page",
+      "dependencies": {
+        "@aibtc/tx-schemas": "^1.1.0",
+        "@noble/curves": "^2.0.1",
+        "@noble/hashes": "^2.0.1",
+        "@opennextjs/cloudflare": "^1.17.1",
+        "@scure/base": "^2.0.0",
+        "@scure/btc-signer": "^2.0.1",
+        "@stacks/common": "^7.3.1",
+        "@stacks/encryption": "^7.3.1",
+        "@stacks/transactions": "^7.3.1",
+        "date-fns": "^4.1.0",
+        "next": "^15.5.18",
+        "react": "^19.2.4",
+        "react-dom": "^19.2.4",
+        "react-markdown": "^10.1.0",
+        "remark-breaks": "^4.0.0",
+        "remark-gfm": "^4.0.1",
+        "swr": "^2.4.0",
+        "x402-stacks": "^2.0.1",
+      },
+      "devDependencies": {
+        "@cloudflare/workers-types": "^4.20260212.0",
+        "@tailwindcss/postcss": "^4",
+        "@types/node": "^20",
+        "@types/react": "^19",
+        "@types/react-dom": "^19",
+        "eslint": "^9",
+        "eslint-config-next": "^15.1.3",
+        "patch-package": "^8.0.1",
+        "tailwindcss": "^4.1.18",
+        "typescript": "^5",
+        "vitest": "^4.0.18",
+        "wrangler": "^4.65.0",
+      },
+    },
+  },
+  "overrides": {
+    "axios": ">=1.17.0",
+    "esbuild": ">=0.28.1",
+    "flatted": ">=3.4.0",
+    "path-to-regexp": ">=8.4.0",
+    "picomatch": ">=4.0.4",
+    "undici": ">=7.24.0",
+  },
+  "packages": {
+    "@aibtc/tx-schemas": ["@aibtc/tx-schemas@1.1.0", "", { "dependencies": { "zod": "^4.3.6" } }, "sha512-jCautY4s8xT6oYC6l5d5tR9pbKTIziyny3YFI77mcmkoCzZzw7oedARYARuw1DSyxcHpXVoRTL0VM6MlTYA3LQ=="],
+
+    "@alloc/quick-lru": ["@alloc/quick-lru@5.2.0", "", {}, "sha512-UrcABB+4bUrFABwbluTIBErXwvbsU/V7TZWfmbgJfbkwiBuziS9gxdODUyuiecfdGQ85jglMW6juS3+z5TsKLw=="],
+
+    "@ast-grep/napi": ["@ast-grep/napi@0.40.5", "", { "optionalDependencies": { "@ast-grep/napi-darwin-arm64": "0.40.5", "@ast-grep/napi-darwin-x64": "0.40.5", "@ast-grep/napi-linux-arm64-gnu": "0.40.5", "@ast-grep/napi-linux-arm64-musl": "0.40.5", "@ast-grep/napi-linux-x64-gnu": "0.40.5", "@ast-grep/napi-linux-x64-musl": "0.40.5", "@ast-grep/napi-win32-arm64-msvc": "0.40.5", "@ast-grep/napi-win32-ia32-msvc": "0.40.5", "@ast-grep/napi-win32-x64-msvc": "0.40.5" } }, "sha512-hJA62OeBKUQT68DD2gDyhOqJxZxycqg8wLxbqjgqSzYttCMSDL9tiAQ9abgekBYNHudbJosm9sWOEbmCDfpX2A=="],
+
+    "@ast-grep/napi-darwin-arm64": ["@ast-grep/napi-darwin-arm64@0.40.5", "", { "os": "darwin", "cpu": "arm64" }, "sha512-2F072fGN0WTq7KI3okuEnkGJVEHLbi56Bw1H6NAMf7j2mJJeQWsRyGOMcyNnUXZDeNdvoMH0OB2a5wwUegY/nQ=="],
+
+    "@ast-grep/napi-darwin-x64": ["@ast-grep/napi-darwin-x64@0.40.5", "", { "os": "darwin", "cpu": "x64" }, "sha512-dJMidHZhhxuLBYNi6/FKI812jQ7wcFPSKkVPwviez2D+KvYagapUMAV/4dJ7FCORfguVk8Y0jpPAlYmWRT5nvA=="],
+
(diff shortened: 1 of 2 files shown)


secret-mars (commented):
Reading #997 — the esbuild override + bun.lock pair.

**Code change (`package.json`).** Adds `"esbuild": ">=0.28.1"` to the existing `overrides` block (alongside `flatted`, `picomatch`, `path-to-regexp`, `axios`). Same shape as #418 on the x402-sponsor-relay repo, which I just reviewed at [pullrequestreview-4525409528](https://github.com/aibtcdev/x402-sponsor-relay/pull/418#pullrequestreview-4525409528) — esbuild dev-only transitive, CVSS 8.1 DNS rebinding patched at >=0.28.1, no production runtime exposure. That part is correct.

**Real CI blocker — npm ci vs bun.lock.** Both Lint + Test fail with `Run npm ci`. The PR adds `bun.lock` (1958 lines, new) but the workflow at `.github/workflows/...` still calls `npm ci`, which expects `package-lock.json`. If the PR doesn't also commit a `package-lock.json` update (or update the workflow to `bun install` / `bun ci`), `npm ci` will fail every run on this branch because the lockfile shape it's looking for isn't there. Worth a quick check of whether `package-lock.json` is gone in this PR — if yes, fix path is either:

- **Path A**: also commit the updated `package-lock.json` matching the bun resolution (keep npm tooling), OR
- **Path B**: update the workflow files (Lint + Test jobs) to `bun install` / `bun test`, OR
- **Path C**: switch to a `package.json`-overrides-only change without committing the lockfile (mirror the x402sr#418 approach exactly) — smallest blast radius if the goal is just to close the dependabot alert.

Workers Builds passing here (the Cloudflare-side bundler resolves overrides through its own toolchain regardless of which lockfile is committed) is consistent with Path C being viable.

**CodeQL + Snyk + JS-TS Analyze all pass** — the CVE remediation itself isn't disputed, just the workflow-tooling mismatch.

**Verdict.** LGTM on the override pin. The bun.lock-vs-npm-ci mismatch needs one of the three paths above before merge. Path C (just the package.json override, no lockfile) is the cleanest and matches the parallel x402sr#418 shape exactly — happy to file a follow-up PR with that shape if helpful.

biwasxyz:
Thanks @arc0btc. Superseded by #1083: the wrangler 4.132 and vite 7.3.6 bumps pull esbuild 0.28.1 and 0.28.2, which clears the alert. The only older copy left is 0.25.4 under `@opennextjs/aws`, which is outside the vulnerable range (>=0.27.3 <0.28.1), so a global override was not needed. There are no open Dependabot alerts on `main`.

PR #999, opened 2026-06-17
feat(legion): live Legion governance dashboard at /legion

biwasxyz:
## What

A read-only dashboard at **`/legion`** showing an AIBTC Legion (on-chain agent collective on **Stacks testnet**) and the full block-height lifecycle of its governance proposals. Linked from `/bounties`.

### Screens
- **Header** — treasury contract + live block height, pooled sBTC / total staked / members-staked cards, the governance "constitution" (quorum 15% / threshold 66% / min 2 voters / veto rule), and gov/payout/token wiring badges.
- **Members** — 10 agents sorted by stake, with voting-weight %, wallet sBTC balance, staked vs not-staked distinction.
- **Proposals** (centerpiece) — per-proposal lifecycle tracker (Pending → Voting → Veto → Execution → Concluded/Expired) derived from the contract's block windows (never hardcoded), YES/NO/VETO tally bars, quorum/threshold/voter/veto badges, a passing/not-passing flag, and per-agent vote chips.
- **How agents participate** — static, contract-accurate guide (faucet → stake → propose → vote → veto → conclude).

## How — Cloudflare cost design

Hiro read volume is **decoupled from page traffic**:

- The existing **5-min cron** (`worker.ts scheduled()` → `runScheduledTasks`) builds one snapshot via a server-side fan-out and stores it as a single KV blob (`legion:snapshot`).
- `/api/legion` just serves that blob behind `caches.default` (60s edge TTL). Cold start builds inline once, then the cron keeps it warm.
- **Concluded proposals are terminal on-chain**, so they're carried forward from the prior snapshot and never re-read — only in-flight proposals + the always-changing treasury/stake reads hit Hiro each tick.
- Client is SSR-hydrated SWR polling the cheap cached endpoint (60s, dedupe matched).

Net: ~1 Hiro fan-out + 1 KV write per 5 min, independent of visitor count or colo.

## Notes
- Testnet throughout (data + explorer links), separate from the mainnet identity path — adds a small testnet read helper rather than overloading the request-critical `callReadOnly`.
- Field names verified against the live testnet contracts before writing decoders.
- Partial-outage safe: a failed read degrades to a partial snapshot (surfaced in `errors`), never a blank page.

## Verification
- `npm run build` green; `/legion` + `/api/legion` registered.
- Type-check + lint clean on all new/edited files.

### DIFF
--- a/app/api/legion/route.ts
+++ b/app/api/legion/route.ts
@@ -1,0 +1,104 @@
+// CACHE_INVARIANTS:POSTURE=public-only-get
+// Read-only Legion dashboard snapshot. No auth, no per-caller branching.
+// Data is built by the cron (lib/scheduler/cron-runner.ts → runLegionNow) and
+// stored in KV; this endpoint serves that blob behind an edge cache. The first
+// visitor after a cold start (empty KV) builds it inline once, then the cron
+// keeps it warm — so Hiro is hit at most once per refresh window regardless of
+// traffic.
+
+import { NextRequest, NextResponse } from "next/server";
+import { getCloudflareContext } from "@opennextjs/cloudflare";
+import { createLogger, createConsoleLogger, isLogsRPC } from "@/lib/logging";
+import { withEdgeCache } from "@/lib/edge-cache";
+import {
+  buildLegionSnapshot,
+  readLegionSnapshot,
+  writeLegionSnapshot,
+} from "@/lib/legion/snapshot";
+import {
+  GOV_CONTRACT,
+  GOV_RULES,
+  PAYOUT_CONTRACT,
+  TREASURY_CONTRACT,
+} from "@/lib/legion/constants";
+
+const EDGE_CACHE_KEY = "https://cache.aibtc.local/api/legion";
+// Edge + client TTL. Data only changes when the 5-min cron rewrites the blob,
+// so a 60s edge window caps KV reads without ever serving stale-by-much data.
+const TTL_SECONDS = 60;
+
+function selfDocResponse() {
+  return NextResponse.json(
+    {
+      endpoint: "/api/legion",
+      method: "GET",
+      description:
+        "Read-only snapshot of the AIBTC Legion (Stacks testnet): treasury, members, and the full lifecycle of every governance proposal. Built server-side on a cron and served from cache — this endpoint never accepts writes.",
+      network: "stacks-testnet",
+      contracts: {
+        treasury: TREASURY_CONTRACT,
+        gov: GOV_CONTRACT,
+        payout: PAYOUT_CONTRACT,
+      },
+      governanceRules: GOV_RULES,
+      queryParameters: {
+        docs: {
+          type: "string",
+          description: "Pass ?docs=1 to return this documentation payload instead of data",
+          example: "?docs=1",
+        },
+      },
+      responseShape: {
+        updatedAt: "number (unix ms)",
+        blockHeight: "number | null (stacks tip at snapshot time)",
+        treasury: "{ balance (sats), govWired, payoutWired, tokenWired }",
+        totalStaked: "number | null (sats)",
+        members: "Array<{ label, address, stake, weightPct, sbtcBalance }>",
+        proposals:
+          "Array<{ id, proposer, recipient, desc, amount, status, votes }> (newest first)",
+        errors: "string[] (per-read failures; partial snapshots are still served)",
+      },
+    },
+    { headers: { "Cache-Control": "public, max-age=3600, s-maxage=86400" } },
+  );
+}
+
+export async function GET(request: NextRequest) {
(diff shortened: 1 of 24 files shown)


arc0btc (approved):
Live Legion governance dashboard at /legion — read-only Stacks testnet view of treasury, stake-weighted members, and block-height-derived proposal lifecycle. The caching design (cron builds snapshot → KV → edge cache → SWR poll) is well-architected: Hiro read volume is fixed per refresh interval regardless of traffic, and the cold-start inline build + waitUntil persist pattern is clean.

**What works well:**
- Hydration safety: `UpdatedAt` correctly defers `Date.now()` to `useEffect` with `useState<number | null>(null)` — hydration mismatch avoided
- Parallelism throughout `buildLegionSnapshot` — all independent reads use `Promise.all`, including the per-agent fan-out
- Terminal proposal caching: concluded proposals carried from `prev` snapshot saves re-reads on every cron tick
- `safe()` wrapper degrades gracefully to partial snapshots instead of throwing — `errors[]` surface in the UI
- Semantic keys on all lists (`key={m.address}`, `key={p.id}`, `key={s.stage}`) — no index-as-key issues
- `aria-busy` + `aria-label` on the loading skeleton, `aria-hidden` on connector lines in LifecycleTracker

---

## Performance

No CRITICAL violations. One cadence question worth resolving:

**[suggestion] `LEGION_INTERVAL_MS` is exported but never used**

In `cron-runner.ts`, `LEGION_INTERVAL_MS = 5 * 60 * 1000` is defined alongside the other interval constants, and the comment says "cron cadence == LEGION_INTERVAL_MS" — but unlike `TENERO_INTERVAL_MS` / `COMPETITION_INTERVAL_MS`, no last-run gate is implemented. The snapshot runs every cron tick. If the CF worker schedule is sub-5-min, this triggers more Hiro fan-outs than intended. Either add the last-run gate (using a `lastRunAt` key in KV, same pattern as tenero) or drop the constant and update the comment to say "runs every cron tick (no gating needed)":

```suggestion
// Legion snapshot — runs every cron tick. Hiro volume is bounded by the cron
// schedule itself; no separate state key needed since a slow overwrite is idempotent.
```

---

## Composition

No violations. Single `ok: boolean` props on `WiringBadge` and `Badge` are state booleans (not style config booleans) — correct use.

---

## UI/Accessibility

**[suggestion] Badge emoji not `aria-hidden` in ProposalCard** (`app/legion/ProposalCard.tsx`)

`WiringBadge` in `LegionHeader.tsx` correctly wraps its checkmark: `<span aria-hidden>{ok ? "✓" : "—"}</span>`. The `Badge` component in `ProposalCard.tsx` renders `{ok ? "✓" : "✗"}` without `aria-hidden`, so screen readers announce "tick mark Quorum" / "cross mark Threshold". Consistent fix:

```suggestion
      <span aria-hidden>{ok ? "✓" : "✗"}</span> {label}
```

**[suggestion] Missing `scope="col"` on table headers** (`app/legion/MembersTable.tsx`)

`<th>` cells without `scope` are valid HTML5, but explicit `scope="col"` is required by WCAG 1.3.1 for data tables with multiple columns:

```suggestion
            <th className="px-5 py-3 font-medium" scope="col">Member</th>
```
(apply to all five `<th>` cells)

**[nit] `explorerContractUrl` uses `/txid/` path for a contract identifier** (`lib/legion/constants.ts`)

The treasury link in `LegionHeader` goes to `https://explorer.hiro.so/txid/<contract.name>?chain=testnet`. The `/txid/` path resolves to the deployment transaction, not the contract page. `/address/<contract-id>` is the canonical contract view on Hiro:

```suggestion
export function explorerContractUrl(contractId: string): string {
  return `https://explorer.hiro.so/address/${contractId}?chain=${LEGION_CHAIN}`;
}
```

**[question] Cold-start concurrent builds** (`app/api/legion/route.ts`)

Two simultaneous requests hitting a cold KV will both pass the `if (!snapshot)` check and independently call `buildLegionSnapshot`, each triggering the full Hiro fan-out. On testnet with low traffic this is fine — both writes are idempotent — but worth knowing. Mitigation (if ever needed) would be a single KV write with `if-none-match` or a short-lived lock key.

**[question] KV namespace: `env.VERIFIED_AGENTS` for Legion snapshot**

The snapshot key `legion:snapshot` lives in the `VERIFIED_AGENTS` namespace alongside agent verification data. Is this intentional (reusing the existing binding to avoid new CF infra), or should it have its own namespace eventually? If the namespace has quota concerns, Legion's cron-frequency writes could compete.

---

*Operational note:* We run the Stacks testnet API path in our own sensors. The `detect429` + `perAttemptTimeoutMs` pattern in `stacks.ts` matches what we use — good call reusing `stacksApiFetch` instead of raw `fetch`. The 5s per-attempt timeout is appropriate for testnet latency.

PR #1001, opened 2026-06-18
Revert legion (#999) + cron capture (#1000) — isolate cron Internal Error

biwasxyz:
## Why

The production cron has been failing every tick with 'Internal Error' since ~yesterday, freezing all scheduled work (Tenero prices, competition sweep, earnings, legion). The legion PR (#999) is the only change that touched the cron path (`runScheduledTasks` → `runLegionNow`) at the time it broke.

This reverts **both**:
- **#999** (legion dashboard) — removes `runLegionNow` from the cron and all legion files.
- **#1000** (cron error-capture) — restores the original scheduled handler.

Result: the scheduler is back to the exact pre-both-PRs state (when the cron last worked).

## Test

After deploy, watch whether the cron recovers — `scheduler:tenero`/`competition`/`earnings` lastRunAt should start advancing again.
- **Cron recovers** → legion's cron task was the cause → re-introduce legion with a fix that can't break the scheduler.
- **Still broken** → legion is exonerated; look elsewhere.

Legion data isn't lost — it's reconstructable from chain; this is a diagnostic revert.


arc0btc (approved):
Diagnostic revert of both #999 (legion dashboard) and #1000 (cron error-capture) to restore the scheduler to its last known-working state. The approach is sound — clean bisect: if `scheduler:tenero`/`competition`/`earnings` lastRunAt starts advancing again, legion's cron task was the culprit; if not, the search continues.

**What looks good:**
- The revert is mechanical and complete — no partial legion remnants left in `cron-runner.ts` or `worker.ts`
- The simplified `.catch()` in `worker.ts` is cleaner than the nested triple-guard from #1000
- The PR description clearly states the diagnostic intent and the two possible outcomes, making follow-up triage unambiguous

**[question]** Reverting #1000 loses the KV-based error capture that was specifically added because the LOGS RPC path wasn't surfacing the stack trace (`wrangler kv key get scheduler:last-error`). If the cron is still broken after this lands, you'll be back to the "opaque Internal Error" with no stack visible. Worth having a plan ready: if the cron remains broken after this revert, re-add just the KV capture wrapper (the `try { await env.VERIFIED_AGENTS.put(...) }` block from #1000's `worker.ts`) as a standalone diagnostic PR — separate from any legion code — so you can read the exact exception.

**Code quality notes:**
No issues — this is a clean mechanical revert with no new logic introduced.

**Operational note:** We monitor `scheduler:tenero` and `scheduler:competition` lastRunAt via Arc sensors. A scheduler freeze of this duration (~24h) is consistent with an unhandled exception in the `scheduled()` handler aborting the `ctx.waitUntil()` promise — the exact scenario #1000 was trying to instrument. The diagnostic revert approach is the right call here.

PR #1003, opened 2026-06-18
Re-land legion + standalone aibtc-scheduler worker (retire landing-page cron)

biwasxyz:
## What

Closes out the scheduler saga:

1. **Re-land legion** — #999 was reverted (#1001) only to isolate the cron crash; it was **innocent**. The real cause was the deprecated Hiro `x-hiro-api-key` header (fixed in #1002). This reverts the revert, bringing legion back (dashboard, API, lib, migration) including its own `x-api-key` usage and the reverted-vote filter.

2. **Standalone `aibtc-scheduler` worker** (`scheduler-worker.ts` + `wrangler.scheduler.jsonc`) — runs the existing `runScheduledTasks` (Tenero + competition + earnings + legion) on its **own cron**, bound to the **same** D1 (`landing-page`) + KV. Already deployed and verified live (competition 100/100, Hiro + Tenero keys authenticating).

3. **Retire landing-page's own cron** — emptied `triggers.crons` (top-level + env.production). landing-page's trigger kept getting silently orphaned by deploys; `aibtc-scheduler` owns scheduling now.

## Why

A fresh worker's cron fires reliably on this account; landing-page's specifically did not (every deploy re-orphaned it). Isolating the scheduler in its own worker permanently ends that.

## Deploy steps (after merge)
1. Merge → landing-page deploys with empty crons → its trigger is removed.
2. Redeploy the scheduler with the re-landed legion code:
   `npx wrangler deploy --config wrangler.scheduler.jsonc`
   (secrets HIRO_API_KEY/TENERO_API_KEY already set on aibtc-scheduler)
3. Bump scheduler cron back to `*/5` is already the committed value (was temporarily 1-min for verification).

## Verified
- typecheck clean; scheduler worker bundles legion (298 KiB)
- live: competition + Hiro + Tenero all authenticating on the deployed aibtc-scheduler

### DIFF
--- a/app/api/legion/route.ts
+++ b/app/api/legion/route.ts
@@ -1,0 +1,78 @@
+// CACHE_INVARIANTS:POSTURE=public-only-get
+// Read-only Legion dashboard snapshot (Stacks testnet). No auth, no per-caller
+// branching. D1 is the source of truth (written by the 5-min cron); this
+// endpoint reads it via getLegionSnapshot, which layers caches.default +
+// singleflight on top (mirrors the leaderboard). Hiro is only hit on a
+// stale/cold rebuild, never per request.
+
+import { NextRequest, NextResponse } from "next/server";
+import { getCloudflareContext } from "@opennextjs/cloudflare";
+import { createLogger, createConsoleLogger, isLogsRPC } from "@/lib/logging";
+import { getLegionSnapshot } from "@/lib/legion/read";
+import {
+  GOV_CONTRACT,
+  GOV_RULES,
+  TREASURY_CONTRACT,
+} from "@/lib/legion/constants";
+
+export const dynamic = "force-dynamic";
+
+function selfDocResponse() {
+  return NextResponse.json(
+    {
+      endpoint: "/api/legion",
+      method: "GET",
+      description:
+        "Read-only snapshot of the AIBTC Legion (Stacks testnet): treasury, members, and the full lifecycle of every governance proposal — including who voted and whether each proposal is passing. Built server-side on a cron, stored in D1, served from cache. Never accepts writes.",
+      network: "stacks-testnet",
+      contracts: {
+        treasury: TREASURY_CONTRACT,
+        gov: GOV_CONTRACT,
+      },
+      governanceRules: GOV_RULES,
+      queryParameters: {
+        docs: {
+          type: "string",
+          description: "Pass ?docs=1 to return this documentation payload instead of data",
+          example: "?docs=1",
+        },
+      },
+      responseShape: {
+        updatedAt: "number (unix ms)",
+        blockHeight: "number | null (stacks tip at snapshot time)",
+        treasury: "{ balance (sats), govWired, tokenWired }",
+        totalStaked: "number | null (sats)",
+        members: "Array<{ label, address, stake, weightPct, sbtcBalance }>",
+        proposals:
+          "Array<{ id, proposer, recipient, desc, amount, status, votes }> (newest first); status carries metQuorum/metThreshold/voterCount/vetoActivated/concluded/executed; votes is per-agent { voted, vote, amount }",
+        errors: "string[] (per-read failures; partial snapshots are still served)",
+      },
+    },
+    { headers: { "Cache-Control": "public, max-age=3600, s-maxage=86400" } },
+  );
+}
+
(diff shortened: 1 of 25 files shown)


arc0btc (approved):
Re-lands the Legion governance dashboard + moves the scheduler to a standalone Worker to fix Cloudflare cron orphaning on landing-page deploys. Both changes are well-scoped and the architectural reasoning is sound.

**What works well:**
- The stale-while-revalidate read path in `lib/legion/read.ts` is the right call — cold builds (one Hiro fan-out blocking a page render) versus warm reads (D1 row served immediately, revalidate detached via `ctx.waitUntil`) are clearly separated.
- The singleflight guard on both the in-flight rebuild and the background revalidate (`inFlight` / `revalidating` maps) prevents duplicate Hiro fan-outs under concurrent load.
- `prev`-based carry-forward for concluded proposals is a nice optimization: terminal proposals never cost a re-read, and a partial build never blanks out good data.
- `UpdatedAt` in `LegionClient.tsx` correctly initializes to `null` on the server and hydrates `Date.now()` in `useEffect` — no hydration mismatch.
- `scheduler-worker.ts` re-throws on failure, so CF Cron Past Events still records real failures for the new standalone worker.

**[suggestion] `isPassing` hardcodes `voterCount >= 2` instead of `GOV_RULES.minVoters`** (`lib/legion/lifecycle.ts:89`)

```ts
// current
status.voterCount >= 2 &&
// sync with the rules constant
status.voterCount >= GOV_RULES.minVoters &&
```

`GOV_RULES.minVoters` is already `2`, so this is a no-op today — but if the governance rules change the dashboard gate logic will silently diverge from the on-chain constitution.

**[question] `AddressLink.tsx:27` — `CopyButton` with `label=""`**

```tsx
<CopyButton text={address} variant="inline" label="" />
```

If `CopyButton` renders an icon-only button with this label as its accessible name, an empty string would fail the `aria-icon-buttons` rule. Does the component supply its own `aria-label`/`title` fallback? If not, something like `label="Copy address"` or `aria-label="Copy address"` should be explicit here.

**[suggestion] Emoji in outcome text will be announced literally by screen readers** (`app/legion/ProposalCard.tsx:144-147`)

```tsx
// current
"Executed ✅" / "Rejected ❌"

// screen-reader-friendly alternative
<>Executed <span aria-hidden>✅</span></> / <>Rejected <span aria-hidden>❌</span></>
```

Screen readers announce `✅` as "check mark button" and `❌` as "cross mark", which is distracting mid-sentence.

**[suggestion] `HowToParticipate` and `LegionHeader` in the client bundle** (`app/legion/LegionClient.tsx`)

`HowToParticipate` is pure static JSX — no snapshot data, no interactivity. Because `LegionClient.tsx` has `"use client"`, everything it imports is bundled client-side. Passing static sections as `children` from `app/legion/page.tsx` (a Server Component) would keep them server-rendered and out of the JS bundle. Optional here, but worth flagging since the pattern scales.

**[nit] `worker.ts` scheduled handler dead code**

With `"crons": []` in `wrangler.jsonc`, this worker's `scheduled()` will never fire — but the handler (now wrapped in `ctx.waitUntil`) remains. The old version re-threw so CF would record failures; the new version swallows them. Both are moot since the trigger is gone, but a brief comment noting the handler is intentionally inert (cron moved to `aibtc-scheduler`) would prevent future confusion.

**Code quality notes:**
- `createLimiter` in `snapshot.ts` is a clean self-contained pLimit-style limiter — avoids an npm dep for a ~15-line function.
- The `get<T>` defensive accessor is verbose but safe against malformed Clarity responses. If you ever normalize the Clarity decode layer (e.g., typed return from `parseClarityValue`), it could be replaced with a type assertion. No action needed today.
- One implicit behavior: `govWired: govWire != null ? true : ...` treats any non-null read result as "wired" regardless of what the contract returned. For a display-only dashboard this is fine, but worth a comment so the next reader understands it's a connectivity check, not a deep validation.

**Operational note:** We monitor the Cloudflare Durable Objects and D1 row-read quota closely. The Legion snapshot is a single-row upsert/read — zero row-read pressure. The only quota risk is the Hiro fan-out during a cold build (bounded by `LEGION_READ_CONCURRENCY = 6`), which runs in a Worker and doesn't touch DO storage at all. No concerns from our side.

PR #1011, opened 2026-06-23
legion: point dashboard at v3.0 deploy + NYT critique bounty skill

biwasxyz:
## What

Updates the Legion dashboard + agent skill for the v3.0 Phase-1 testnet deploy (legion-agent-03). No contract code here — the NYT rules are an off-chain voting norm.

### Changes
- **lib/legion/constants.ts** — `LEGION_DEPLOYER` → `STBEMQQ…` (v3.0 deploy, supersedes agent-02); add `FEES_CONTRACT` (legion-fees, 8% collector).
- **app/legion/skill.md** — new `propose` signature (content-hash / inscription-height / sources), bond + Rail-A reject codes, proposer-exclusion, eligible-stake quorum, `unstake`, fee collector, and the MCP `0x`-buffer encoding gotcha. Adds a **"Current program: NYT critique bounty"** section (rubric, reply-to-claim eligibility, two-tier reward, `desc` packing).
- **app/legion/HowToParticipate.tsx** — updated `propose` signature + bond / no-self-vote note.

### Untouched (correct as-is)
`snapshot.ts` / `read.ts` / `lifecycle.ts` / `types.ts` key off these constants and read tuple fields by key, so they pick up the new contracts automatically.

### DIFF
--- a/app/legion/HowToParticipate.tsx
+++ b/app/legion/HowToParticipate.tsx
@@ -23,9 +23,14 @@ const STEPS: { title: string; body: React.ReactNode }[] = [
     title: "Propose",
     body: (
       <>
-        <code>legion-gov propose(description, recipient, amount)</code> —
-        description 1–256 ASCII; recipient must not be the gov or treasury
-        contract; amount &gt; 0.
+        <code>
+          legion-gov propose(description, recipient, amount, content-hash,
+          inscription-height, sources)
+        </code>{" "}
+        — description 1–256 ASCII; recipient ≠ gov/treasury; amount &gt; 0.
+        Rail-A gates reject at propose: a unique content-hash, a fresh
+        inscription-height, and ≥2 sources. A 20% bond is locked from your stake,
+        and you cannot vote on your own proposal.
       </>
     ),
   },
--- a/app/legion/skill.md/route.ts
+++ b/app/legion/skill.md/route.ts
@@ -2,6 +2,7 @@ import { NextResponse } from "next/server";
 import {
   GOV_CONTRACT,
   TREASURY_CONTRACT,
+  FEES_CONTRACT,
   SBTC_TOKEN,
   GOV_RULES,
 } from "@/lib/legion/constants";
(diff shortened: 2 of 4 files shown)


arc0btc (approved):
Points the Legion dashboard at the v3.0 Phase-1 testnet deploy, wires in the new `legion-fees` contract, and ships a detailed NYT critique bounty program into the agent skill route. Solid.

**What works well:**
- The comment on `LEGION_DEPLOYER` in `constants.ts` precisely explains what changed (fees, Rail-A prechecks, bond-lock, unstake, proposer-exclusion) — future readers won't have to archaeology the git log.
- The `⚠️ Encode the content-hash hex without a 0x prefix` note in the MCP section is the kind of silent footgun that would show up as `err u420` with no obvious cause. Good catch to surface it explicitly.
- Quorum language update ("eligible (non-proposer) staked") correctly reflects the proposer-exclusion rule — the old wording was technically ambiguous.
- The two-tier reward model (on-chain testnet sBTC automatic, real sBTC manual after authenticity verification) is documented honestly. The voter guidance block makes clear the chain can't verify authenticity and voters must — that's exactly the right framing.
- PR description correctly calls out that `snapshot.ts`, `read.ts`, `lifecycle.ts`, and `types.ts` pick up the new constants automatically. That's a helpful safety note.

## Performance

No async changes. `skill.md/route.ts` is a Route Handler returning a template literal — builds synchronously from imported constants, no waterfall concerns.

**[suggestion] Consider caching the skill route** — the content is effectively static between deploys. Adding `export const dynamic = 'force-static'` or `Cache-Control: max-age=300` would avoid rebuilding the string on every AI agent fetch. Low priority since the computation is trivial, but it would show up cleanly as a static route in the build output.

## Composition

No structural component changes. The `HowToParticipate.tsx` change is text content only — no component API concerns.

## UI/Accessibility

No interactive elements changed. The `<code>` block reformatting in `HowToParticipate.tsx` is cleaner given the longer propose signature — the multi-line format reads better than a single-line overrun.

**[nit]** The `{" "}` spacer after `</code>` in `HowToParticipate.tsx:32` is correct (prevents the em dash from colliding with the closing tag), but a trailing space inside the JSX text would be more idiomatic:
```suggestion
        </code>{" "}
```
(Already how it's written — no change needed. Just noting it's intentional, not a stray character.)

## Code quality notes

**[question]** `GOV_RULES` (quorumPct, thresholdPct, minVoters, vetoPct) is referenced in the skill route but the PR doesn't mention whether v3.0 changes those values. If the on-chain parameters changed with the new deploy, the skill route will advertise stale numbers to agents. Worth confirming these are unchanged before merge.

*Operational note:* We monitor the Legion contracts via sensor. The deployer swap will be picked up automatically on the next sensor cycle — no manual config needed on our end.

PR #1013, opened 2026-06-25
Multi-Legion registry: list, detail-by-id, and provider Legions

biwasxyz:
## What

The `/legion` dashboard was hardcoded to one demand Legion. The platform is now **multi-Legion**: a shared on-chain `legion-registry` lists many Legions, including a new **provider** kind (inference operators who stake a bond and earn sBTC per call). This PR reads the registry, lists every Legion, renders each by id, and branches the UI on kind.

## How it's structured

**Backend (`lib/legion/`)**
- `registry.ts` — `listLegions()` / `getLegion()` read `legion-registry` (`get-count` + `get-legion(id)`); the existing demand deploy is kept as a `source:"fallback"` entry, deduped by treasury.
- `providers.ts` — provider record reads + `get-min-bond`; provider enumeration via a `register`-event scan (the on-chain `Providers` map has no list-all); `buildProviderSnapshot()`.
- `snapshot.ts` — `buildLegionSnapshot()` now reads gov/treasury from the registry entry; `buildRegistrySnapshot()` builds the `/legions` index.
- `d1.ts` + **migration `024`** — per-Legion keyed `legion_snapshots` table (replaces the single-row `legion_snapshot`).
- `read.ts` — caches.default + singleflight + stale-while-revalidate, generalized and keyed per Legion; `getRegistrySnapshot` / `getProviderSnapshot` / `resolveLegionEntry`.
- `cron-runner.ts` — `runLegionNow` writes the registry index + one snapshot per active Legion, branching on kind.

**Frontend / API**
- `/legions` index (cards + kind badges) and `/legions/[id]` detail branching **demand** (proposals/votes) vs **provider** (`ProvidersTable`: bond, model, jobs ok/fail + `HowToProvide`).
- `/legion` → **redirects to `/legions/demand`** (back-compat); `/api/legion` unchanged.
- `/api/legions` + `/api/legions/[id]` (self-documenting).
- `skill.md` extended with registry discovery + provider onboarding; Navbar + sitemap add `/legions`.

## Verification

- Typecheck clean (changed files), lint clean, production build passes.
- Ran the real worker locally (`npm run preview`, workerd + miniflare) against **live testnet**, local D1 only:
  - `/api/legions` → both Legions, **0 errors** (demand: 493,900 sats / 9 proposals; provider #1 `qwen2.5-7b`: 200 sats / 1 provider).
  - `/api/legions/1` (provider) and `/api/legions/demand` (demand) return correct, kind-branched detail; provider contract derived correctly, real provider record.
  - `/legions`, `/legions/1`, `/legions/demand` render; `/legion` 307→`/legions/demand`; unknown id 404; back-compat `/api/legion` 200.
  - Cold-read self-heal wrote all three D1 rows (`__registry__`, `demand`, `1`), exercising the cron write-path shape too.
- All four contracts verified live on testnet before writing the parsers, so types match real on-chain shapes.

## Deploy notes

- Migration **`024` has already been applied to remote D1** (only pending migration; plain `CREATE TABLE`). The table starts empty and is populated lazily by the first read (cold build) or the scheduler.
- Refresh is already wired: the standalone `aibtc-scheduler` worker (`scheduler-worker.ts`, cron `*/5`) runs `runScheduledTasks` → `runLegionNow` every tick against the shared D1, so the rows refresh automatically once both workers are deployed with this code.

### DIFF
--- a/app/api/legions/[id]/route.ts
+++ b/app/api/legions/[id]/route.ts
@@ -1,0 +1,88 @@
+// CACHE_INVARIANTS:POSTURE=public-only-get
+// Read-only detail for one Legion (Stacks testnet). No auth. Resolves the id to
+// its registry entry, then serves the matching snapshot type — demand (treasury
+// + proposals + votes) or provider (bonds + jobs). D1 is the source of truth
+// (cron-written); reads layer caches.default + singleflight (see
+// lib/legion/read.ts). Hiro is only hit on a stale/cold rebuild.
+
+import { NextRequest, NextResponse } from "next/server";
+import { getCloudflareContext } from "@opennextjs/cloudflare";
+import { createLogger, createConsoleLogger, isLogsRPC } from "@/lib/logging";
+import {
+  getLegionSnapshot,
+  getProviderSnapshot,
+  resolveLegionEntry,
+} from "@/lib/legion/read";
+import { REGISTRY_CONTRACT } from "@/lib/legion/constants";
+
+export const dynamic = "force-dynamic";
+
+function selfDocResponse(id: string) {
+  return NextResponse.json(
+    {
+      endpoint: `/api/legions/${id}`,
+      method: "GET",
+      description:
+        "Read-only snapshot of one AIBTC Legion (Stacks testnet). The shape depends on `entry.kind`: a demand Legion returns treasury, members, and the full lifecycle of every governance proposal; a provider Legion returns the treasury, min bond, and every registered inference provider (bond, model, endpoint, jobs ok/fail). Built server-side on a cron, stored in D1, served from cache.",
+      network: "stacks-testnet",
+      registry: REGISTRY_CONTRACT,
+      idScheme:
+        "Registry numeric id as a string ('1', '2', …), or the slug 'demand' for the known demand Legion.",
+      queryParameters: {
+        docs: {
+          type: "string",
+          description: "Pass ?docs=1 to return this documentation payload instead of data",
+          example: "?docs=1",
+        },
+      },
+      responseShape: {
+        entry:
+          "{ id, kind ('demand'|'provider'), owner, treasury, gov, fees, providers, model, uri, active, source }",
+        snapshot:
+          "Demand: { updatedAt, blockHeight, treasury, totalStaked, members, proposals, errors }. Provider: { updatedAt, blockHeight, treasuryBalance, minBond, providers, errors }.",
+      },
+      related: { index: "/api/legions", skill: "/legion/skill.md" },
+    },
+    { headers: { "Cache-Control": "public, max-age=3600, s-maxage=86400" } },
+  );
+}
+
+export async function GET(
+  request: NextRequest,
+  { params }: { params: Promise<{ id: string }> },
+) {
+  const { id } = await params;
+  const { searchParams } = new URL(request.url);
+  if (searchParams.get("docs") === "1") return selfDocResponse(id);
+
+  const { env, ctx } = await getCloudflareContext();
+  const rayId = request.headers.get("cf-ray") || crypto.randomUUID();
+  const logger = isLogsRPC(env.LOGS)
+    ? createLogger(env.LOGS, ctx, { rayId, path: "/api/legions/[id]" })
+    : createConsoleLogger({ rayId, path: "/api/legions/[id]" });
+
+  const entry = await resolveLegionEntry(env, ctx, id, logger);
+  if (!entry) {
+    return NextResponse.json(
+      { error: "legion_not_found", id },
+      { status: 404, headers: { "Cache-Control": "no-store" } },
+    );
+  }
+
+  const snapshot =
+    entry.kind === "provider"
+      ? await getProviderSnapshot(env, ctx, entry, logger)
+      : await getLegionSnapshot(env, ctx, logger, entry.id, entry);
+
+  if (!snapshot) {
+    return NextResponse.json(
+      { error: "snapshot_unavailable", id },
+      { status: 503, headers: { "Cache-Control": "no-store", "Retry-After": "10" } },
+    );
+  }
+
+  return NextResponse.json(
+    { entry, snapshot },
+    { headers: { "Cache-Control": "public, max-age=60, s-maxage=60" } },
+  );
+}
--- a/app/api/legions/route.ts
+++ b/app/api/legions/route.ts
@@ -1,0 +1,69 @@
+// CACHE_INVARIANTS:POSTURE=public-only-get
+// Read-only Legion registry index (Stacks testnet). No auth, no per-caller
+// branching. D1 is the source of truth (written by the 5-min cron); this
+// endpoint reads it via getRegistrySnapshot, which layers caches.default +
+// singleflight on top (mirrors /api/legion). Hiro is only hit on a stale/cold
+// rebuild, never per request.
+
+import { NextRequest, NextResponse } from "next/server";
+import { getCloudflareContext } from "@opennextjs/cloudflare";
+import { createLogger, createConsoleLogger, isLogsRPC } from "@/lib/logging";
+import { getRegistrySnapshot } from "@/lib/legion/read";
+import { REGISTRY_CONTRACT } from "@/lib/legion/constants";
+
+export const dynamic = "force-dynamic";
+
+function selfDocResponse() {
+  return NextResponse.json(
+    {
+      endpoint: "/api/legions",
+      method: "GET",
+      description:
+        "Read-only index of every AIBTC Legion (Stacks testnet): demand Legions (pool + govern an sBTC treasury) and provider Legions (operators stake a bond, serve a model, earn sBTC per call). Sourced from the on-chain legion-registry, built server-side on a cron, stored in D1, served from cache. Fetch /api/legions/{id} for one Legion's full detail.",
+      network: "stacks-testnet",
+      registry: REGISTRY_CONTRACT,
+      queryParameters: {
+        docs: {
+          type: "string",
+          description: "Pass ?docs=1 to return this documentation payload instead of data",
+          example: "?docs=1",
+        },
+      },
+      responseShape: {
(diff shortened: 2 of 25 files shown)


arc0btc (approved):
Multi-Legion registry: list, detail-by-id, and provider Legions — solid architectural lift. The registry-driven discovery model, SWR generalization, and fallback dedup are all well-designed.

**What works well:**
- The `getCachedSnapshot` generic in `lib/legion/read.ts` is exactly the right abstraction — one SWR pattern keyed per Legion instead of copy-pasting it three times. The in-flight singleflight + background revalidate is a proven pattern here.
- `buildProviderSnapshot` mirrors `buildLegionSnapshot`'s graceful-degradation pattern: every read degrades to partial rather than throwing, prev is used as fallback on failures. Consistent and resilient.
- Provider enumeration via event scan + dedupe is the correct call given there's no on-chain list-all, and the 300-event cap prevents unbounded API calls.
- `Promise.all` used correctly throughout — treasury reads, count reads, and provider record fetches are all parallelized. No sequential awaits where they'd be waterfalls.
- Migration 024 is minimal and correct — JSON-in-TEXT keyed by `legion_id` matches the existing 023 precedent, and the `__registry__` key is a clean convention.
- Back-compat via 307 redirect from `/legion` → `/legions/demand` is the right approach.

---

## Performance

**[suggestion] Double `resolve()` in `app/legions/[id]/page.tsx`** ()

`generateMetadata` and the page component both call `resolve(id)` independently. Next.js calls both in the same render pass, so this fires two Hiro reads (or two D1 reads) on every SSR. Wrap with `React.cache()` to deduplicate across the request:

```suggestion
import { cache } from "react";

const resolve = cache(async (id: string): Promise<LegionEntry | null> => {
  try {
    const { env, ctx } = await getCloudflareContext();
    return await resolveLegionEntry(env, ctx, id);
  } catch {
    return null;
  }
});
```

This is the same dedup pattern the leaderboard uses. Without it, every detail page render hits D1 (or Hiro) twice.

---

## Composition

**[nit] `STEPS` array in `HowToProvide`** ()

`STEPS` is defined inside the component body but never depends on props — it's rebuilt on every render. Hoist to module scope:

```suggestion
const STEPS: { title: string; body: React.ReactNode }[] = [
  // ... same content
];

export default function HowToProvide({ minBond, model }: { ... }) {
```

---

## UI/Accessibility

`app/legion/ProvidersTable.tsx:544` — table `<th>` elements are missing `scope="col"`. Screen readers use this to associate headers with data columns. [semantic-html]

```suggestion
<th className="px-5 py-3 font-medium" scope="col">Provider (STX)</th>
<th className="px-5 py-3 font-medium" scope="col">Model</th>
<th className="px-5 py-3 text-right font-medium" scope="col">Bond</th>
<th className="px-5 py-3 font-medium" scope="col">Status</th>
<th className="px-5 py-3 text-right font-medium" scope="col">Jobs (ok / fail)</th>
```

`app/legions/LegionsClient.tsx:802` — `<h3>` for Legion card names is inside a `<Link>` which is inside an `<article>`-equivalent `<div>`. Not a blocker, but consider wrapping each card in `<article>` for landmark semantics.

---

**Code quality notes:**

`lib/legion/providers.ts:1278-1286` — `toNum` and `get` helpers are duplicated from `lib/legion/snapshot.ts`. Both files define the same two utility functions. Worth extracting to a shared `lib/legion/utils.ts` in a follow-up — not blocking here since this is testnet and the duplication is small, but it'll drift if provider types expand. [suggestion]

---

*Operational note:* The SWR + D1 pattern works well for us on the leaderboard with similar fan-out builds. The 5-min stale threshold matching the cron cadence is correct. One thing to watch: the cold-build path (no D1 row at all) blocks the request inline. The PR description says migration 024 is already applied and D1 gets populated lazily on first read — that's fine for testnet, but if D1 ever gets wiped (a real risk on Cloudflare's free tier), the first request to `/legions` will block on the full registry + demand build in series. Not a blocker; worth a comment in `loadOrRebuild` if this path gets hit in prod.

LGTM — the multi-Legion architecture is solid. The `React.cache()` fix is the one I'd address before shipping to avoid the double D1 hit on every detail page render.

arc0btc (approved):
Multi-Legion registry: list, detail-by-id, and provider Legions — solid architectural lift. The registry-driven discovery model, SWR generalization, and fallback dedup are all well-designed.

**What works well:**
- The `getCachedSnapshot` generic in `lib/legion/read.ts` is exactly the right abstraction — one SWR pattern keyed per Legion instead of copy-pasting it three times.
- `buildProviderSnapshot` mirrors `buildLegionSnapshot`'s graceful-degradation: every read degrades to partial rather than throwing, prev used as fallback on failures.
- Provider enumeration via event scan + dedupe is correct given no on-chain list-all. 300-event cap prevents unbounded API calls.
- `Promise.all` used correctly throughout — no sequential awaits where waterfalls would occur.
- Migration 024 minimal and correct. JSON-in-TEXT keyed by `legion_id` matches 023 precedent, `__registry__` key is a clean convention.
- Back-compat 307 redirect from `/legion` → `/legions/demand` is the right call.

---

## Performance

**[suggestion] Double `resolve()` in `app/legions/[id]/page.tsx`**

`generateMetadata` and the page component both call `resolve(id)` independently. Next.js calls both in the same render pass, so this fires two D1 reads (or two Hiro reads on cold) on every SSR. Wrap with `React.cache()` to deduplicate across the request:

```ts
import { cache } from "react";

const resolve = cache(async (id: string): Promise<LegionEntry | null> => {
  try {
    const { env, ctx } = await getCloudflareContext();
    return await resolveLegionEntry(env, ctx, id);
  } catch {
    return null;
  }
});
```

This is the same dedup pattern the leaderboard uses. Without it, every detail page render hits D1 twice.

---

## Composition

**[nit] `STEPS` array in `HowToProvide.tsx` is defined inside the component body**

`STEPS` never depends on props — it is rebuilt on every render. Hoist to module scope so it is allocated once.

---

## UI/Accessibility

`app/legion/ProvidersTable.tsx` — `<th>` elements in the desktop table are missing `scope="col"`. Screen readers use this to associate headers with data columns. Add `scope="col"` to all five header cells. [semantic-html]

`app/legions/LegionsClient.tsx` — Legion cards are `<Link>` wrapping a `<div>`. Not a blocker, but wrapping each in `<article>` would give better landmark semantics for screen readers scanning the page structure.

---

**Code quality notes:**

`lib/legion/providers.ts` — `toNum` and `get` helper functions are duplicated verbatim from `lib/legion/snapshot.ts`. Both files define identical two-line utilities. Worth extracting to `lib/legion/utils.ts` in a follow-up to avoid drift as provider types expand. [suggestion, non-blocking]

---

*Operational note:* The SWR + D1 pattern works well on the leaderboard with similar fan-out builds. One edge to watch: the cold-build path (no D1 row) blocks the first request inline. PR notes migration 024 is already applied and D1 populates lazily — fine for testnet. If D1 gets wiped (possible on free tier), the first `/legions` request blocks on the full registry + demand build in series. Not blocking; a log comment in `loadOrRebuild` would help ops visibility.

LGTM — the multi-Legion architecture is solid. The `React.cache()` fix is the only one I would address before shipping to avoid the double D1 hit on every detail page render.

secret-mars (commented):
Two orthogonal observations arc's review didn't touch — both non-blocking, fine to address in a follow-up:

**1. Sitemap is index-only.** `app/sitemap.ts` adds `https://aibtc.com/legions` (the index) but not the per-Legion detail URLs `/legions/[id]`. With the cron-runner now iterating over every active `entries` from `listLegions`, the same source is available to make `sitemap.ts` dynamic — read the registry once at sitemap build, emit one entry per active Legion. Otherwise the new detail pages (`/legions/demand`, `/legions/<provider-id>`, future entries) don't surface to crawlers and only the bare index gets indexed.

**2. Cron-runner registry step sits outside the per-Legion resilience boundary.** In `runLegionNow` (lib/scheduler/cron-runner.ts), step 1 (`buildRegistrySnapshot` + `writeRegistrySnapshotToD1` + `listLegions` to get `entries`) runs naked — no try/catch. The per-Legion loop (step 2) properly catches per-entry failures and continues. But:

- `buildRegistrySnapshot` itself has the graceful-degradation pattern (every read in the `Promise.all` wraps via `read()`), so per-Legion summary read failures degrade cleanly. ✓
- However, the call to `listLegions(apiKey, logger)` at the top of `buildRegistrySnapshot` (line ~85 of the diff) is OUTSIDE the `read()` wrapper. If the registry contract `get-count` Hiro read throws (auth, transient 5xx, contract-replaced), `buildRegistrySnapshot` throws → `writeRegistrySnapshotToD1` never runs → and crucially the cron's `listLegions(...)` call on line ~289 (the one feeding the per-Legion loop) is a SEPARATE call. If that one throws too (same Hiro burst), the entire per-Legion loop is skipped for this cron tick.

Suggested fix: wrap step 1 in its own try/catch the same way step 2's per-entry loop does, so a registry-side outage doesn't blank-out per-Legion refreshes that could otherwise have run from the in-memory `entries` of a prior cron's cached `__registry__` D1 row.

```typescript
// Step 1 — registry; degrade to prior D1 row on failure
try {
  const registry = await buildRegistrySnapshot(logger, apiKey);
  await writeRegistrySnapshotToD1(db, registry);
  logger.info("legion.registry_written", { ... });
} catch (e) {
  logger.warn("legion.registry_failed", { error: String(e) });
}

// Step 2 — still try to refresh per-Legion details from the registry
// (re-call listLegions, or read the prior __registry__ row from D1)
const entries = await listLegions(apiKey, logger).catch(() => []);
```

Not blocking — the failure mode is "all-or-nothing cron tick" rather than data corruption, and a subsequent cron will retry. But the per-Legion loop's design intent (one Legion's failure ≠ all Legions' failure) is undercut by an unrelated registry-read failure taking the whole tick down.

Verified the migration 024 minimal-correctness arc flagged. LGTM otherwise on the architectural lift.

PR #1018, opened 2026-06-30
feat(legions): match v1 provider model — free join, engage stake, no bond/slash

biwasxyz:
The `/legions` surfaces still described the **old** per-model **bonded**-provider model (bond, slashing, jobs-ok/fail, `legion-providers`). This brings them in line with **v1**:

- Providers **join the gateway for free** — no bond, no deposit.
- Enforcement is **flag + de-route**, not slash.
- An optional **`legion-engage` stake only buys ranking** (refundable minus a 10% exit fee).

## Data layer
- **types**: `ProviderRecord` drops `bond`/`jobsOk`/`jobsFail`, gains `stake`/`health`/`flagged`; `ProviderSnapshot` drops `minBond`, gains `minStake`/`totalStaked`.
- **new `lib/legion/gateway.ts`**: reads the provider directory from the gateway `GET /v1/providers` (best-effort — unreachable gateway → empty list, never an error). Fuzzy-matches providers to a Legion by model.
- **new `lib/legion/engage.ts`**: reads `legion-engage` `get-stake` / `get-min-stake` / `get-total-staked` (best-effort).
- **`providers.ts`**: builds the provider snapshot from the gateway directory + engage-stake overlay, ranked by stake (mirrors the gateway’s `rankByStake`).
- **`snapshot.ts`**: registry-index provider `count` now comes from the gateway directory, not the on-chain `legion-providers` event scan.
- **constants/read/cron**: thread an optional `LEGION_GATEWAY_URL` Worker var (defaults to the testnet gateway in `constants.ts`).

## Scheduler / data
- The **`aibtc-scheduler`** cron (`runLegionNow`) is the sole writer of `legion_snapshots`; it now fetches the gateway directory + engage stakes instead of scanning `legion-providers`. All new reads are best-effort, so a gateway outage degrades to an empty provider list — never a cron failure.
- **No migration**: same D1 table and one-JSON-doc-per-legion shape; only the fields inside `ProviderSnapshot` change. The cron overwrites rows on its next tick; `LegionSummary` (index) and `LegionSnapshot` (demand) are unchanged.

## UI + docs
ProvidersTable (Bond→Stake, drop jobs, flag status) · ProviderClient · HowToProvide (free register + optional stake step) · LegionsClient header · risks page · `skill.md` provider section · `/legions` + detail metadata · `/api/legions` docs.

## Verification
- `tsc --noEmit`: 0 errors in the legion code (pre-existing `__tests__` errors are unrelated).
- `next lint` on the legion dirs: clean.

Pairs with inference PRs #4 (free-join + stake-ranking, merged) and #5 (switch gateway to testnet). Once the gateway serves testnet, the directory payout addresses line up with the testnet `legion-engage` stakes end-to-end.

### DIFF
--- a/app/api/legions/[id]/route.ts
+++ b/app/api/legions/[id]/route.ts
@@ -39,7 +39,7 @@ function selfDocResponse(id: string) {
         entry:
           "{ id, kind ('demand'|'provider'), owner, treasury, gov, fees, providers, model, uri, active, source }",
         snapshot:
-          "Demand: { updatedAt, blockHeight, treasury, totalStaked, members, proposals, errors }. Provider: { updatedAt, blockHeight, treasuryBalance, minBond, providers, errors }.",
+          "Demand: { updatedAt, blockHeight, treasury, totalStaked, members, proposals, errors }. Provider: { updatedAt, blockHeight, treasuryBalance, minStake, totalStaked, providers: [{ address, name, model, endpoint, stake, health, flagged, active }], errors }.",
       },
       related: { index: "/api/legions", skill: "/legion/skill.md" },
     },
--- a/app/api/legions/route.ts
+++ b/app/api/legions/route.ts
@@ -19,7 +19,7 @@ function selfDocResponse() {
       endpoint: "/api/legions",
       method: "GET",
       description:
-        "Read-only index of every AIBTC Legion (Stacks testnet): demand Legions (pool + govern an sBTC treasury) and provider Legions (operators stake a bond, serve a model, earn sBTC per call). Sourced from the on-chain legion-registry, built server-side on a cron, stored in D1, served from cache. Fetch /api/legions/{id} for one Legion's full detail.",
+        "Read-only index of every AIBTC Legion (Stacks testnet): demand Legions (pool + govern an sBTC treasury) and provider Legions (operators join the gateway for free, serve a model, earn sBTC per call; an optional legion-engage stake buys ranking). Sourced from the on-chain legion-registry, built server-side on a cron, stored in D1, served from cache. Fetch /api/legions/{id} for one Legion's full detail.",
       network: "stacks-testnet",
       registry: REGISTRY_CONTRACT,
       queryParameters: {
--- a/app/legion/HowToProvide.tsx
+++ b/app/legion/HowToProvide.tsx
@@ -3,54 +3,31 @@ import { formatSbtc } from "@/lib/legion/format";
 import CopyButton from "../components/CopyButton";
 
 /**
- * Provider-Legion onboarding: get test sBTC, lock a bond + register, serve the
- * model, get paid. The demand-Legion equivalent is HowToParticipate. The exact
+ * Provider-Legion onboarding (v1): run the model, register with the gateway for
+ * FREE, get paid per call, and OPTIONALLY stake legion-engage to rank higher.
+ * No bond, no slash. The demand-Legion equivalent is HowToParticipate. The exact
  * typed contract call lives in /legion/skill.md.
  */
 
 export default function HowToProvide({
-  minBond,
+  minStake,
   model,
 }: {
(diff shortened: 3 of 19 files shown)


arc0btc (approved):
Matches the v1 provider model across every surface — types, data layer, snapshot builder, UI, API self-docs, and skill.md are all updated together. The best-effort fallback pattern (gateway outage → empty list, never a cron failure) is the right call for testnet.

**What works well:**
- `gateway.ts` and `engage.ts` are cleanly scoped: each module owns exactly one concern. The `normalize()` defensive handling in `gateway.ts` is solid.
- `Promise.all` is used correctly in `buildProviderSnapshot` for the 5 initial reads and again for per-provider stake reads — no sequential waterfalls.
- `StatusDot` is a clean replacement for the old `ActiveDot` + `Jobs` pair. The flagged/active/down triage reads clearly.
- UI copy is consistent: every surface (HowToProvide, ProviderClient, ProvidersTable, risks page, legions index, api docs, skill.md) tells the same story about free join + optional stake.
- The `LEGION_GATEWAY_URL` escape hatch in `wrangler.scheduler.jsonc` is a good pattern and appropriately documented.

---

**[question] `health === "up"` — is this the gateway's exact string?**

`lib/legion/providers.ts` derives `active` from:
```ts
active: !p.flagged && p.health === "up",
```
`GatewayProvider.health` is just `string`. If the gateway returns `"healthy"`, `"online"`, or anything other than `"up"`, every provider silently shows as `active: false` in the table and the `stakedCount` panel says "none yet". Since this is your gateway, you probably know the exact value — but worth documenting as a constant or a comment to prevent future drift.

```suggestion
+/** Health string the gateway emits for a routable provider (exact match). */
+const HEALTH_UP = "up";
 ...
-      active: !p.flagged && p.health === "up",
+      active: !p.flagged && p.health === HEALTH_UP,
```

Or document it in `GatewayProvider.health`:
```ts
/** Gateway health status. "up" = routable; "down" = offline; "unknown" = no data. */
health: string;
```

---

**[suggestion] `LEGION_GATEWAY_URL` not in `CloudflareEnv` type**

```ts
return (env as unknown as { LEGION_GATEWAY_URL?: string }).LEGION_GATEWAY_URL;
```

The double cast works but is fragile — if `CloudflareEnv` is regenerated it won't pick up the var automatically. Consider adding `LEGION_GATEWAY_URL?: string` to the `CloudflareEnv` interface (or wherever the other optional Worker vars like `HIRO_API_KEY` live).

---

**[nit] `legionProvidersContract` now appears unused**

After this PR, `providers.ts` no longer calls `legionProvidersContract` — it uses `legionEngageContract` instead. The function has a back-compat comment, but if nothing else references it, consider removing it (or explicitly marking it `@deprecated`) to keep the module surface clean.

---

**[nit] Uncommenting `LEGION_GATEWAY_URL` in `wrangler.scheduler.jsonc` requires a comma**

The commented example sits after `"EARNINGS_INDEX_ENABLED": "true"` with no trailing comma. If a future operator uncomments the line, the JSON will be invalid. A note in the comment would help:
```
// Add a comma after "EARNINGS_INDEX_ENABLED" line before uncommenting:
// "LEGION_GATEWAY_URL": "https://<gateway-host>"
```

---

**Code quality notes:**

- `countProviders` in `providers.ts` fetches the full directory just to count filtered results. For the index cron this is fine (the registry rebuild is already doing one gateway call per Legion), but if the index grows, it'll be N gateway calls. No action needed now — just something to track.
- `normalize()` in `gateway.ts` silently drops providers with no `payoutAddress`. This is correct (can't join to `legion-engage`), but a `logger?.debug` on the drop would help diagnose gateway data issues without cluttering the log.

---

No CRITICAL performance violations. No BLOCKING accessibility findings (aria-labels on CopyButton are correct throughout, `aria-hidden` on decorative arrows is present). Approving.

PR #1019, opened 2026-06-30
feat: surface governed per-model legions + TESTNET badges

biwasxyz:
Surfaces the 7 governed per-model provider legions in the UI and marks the whole surface as testnet.

## Changes
- **Registry / types / snapshot** carry explicit `treasury`/`gov`/`fees` contract ids — per-model legions use suffixed names (`legion-{treasury,gov,fees}-<family>`) so the `{owner}.legion-*` convention no longer holds; the detail page resolves the right contracts from the cached index.
- **Detail route** (`app/legions/[id]/page.tsx`): a provider legion *with* a gov contract renders the governance view (proposals + members + stake). Only an ungoverned provider legion falls back to the directory view. So a model legion shows its proposals.
- **TESTNET pill** on the `/legions` header and the legion detail header.
- **Copy**: provider legions described as per-model guilds (stake bond → ranking + treasury share + governance, 10% exit fee, rest refundable); Admin → Owner.
- **Cache/cron cadence** 5m → 30m.

Demand (news) legion and its proposals are untouched.

### DIFF
--- a/app/legion/LegionHeader.tsx
+++ b/app/legion/LegionHeader.tsx
@@ -57,6 +57,10 @@ export default function LegionHeader({ snapshot }: { snapshot: LegionSnapshot })
           </span>
         </div>
         <div className="flex items-center gap-3">
+          <span className="inline-flex items-center gap-1.5 rounded-full border border-[#F7931A]/40 bg-[#F7931A]/10 px-2 py-0.5 text-[9px] font-semibold uppercase tracking-[0.12em] text-[#F7931A]">
+            <span className="h-1 w-1 rounded-full bg-[#F7931A]" aria-hidden />
+            Testnet
+          </span>
           <WiringDot label="gov" ok={treasury.govWired} />
           <WiringDot label="token" ok={treasury.tokenWired} />
         </div>
--- a/app/legions/LegionsClient.tsx
+++ b/app/legions/LegionsClient.tsx
@@ -73,7 +73,7 @@ function LegionCard({ legion }: { legion: LegionSummary }) {
       </div>
 
       <div className="border-t border-white/[0.06] pt-3 text-[11px] text-white/40">
-        Admin{" "}
+        Owner{" "}
         <span className="font-mono text-white/60" title={legion.owner}>
           {shortAddress(legion.owner, 6, 4)}
         </span>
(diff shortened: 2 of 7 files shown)


arc0btc (approved):
Surfaces the 7 governed per-model legions and marks the surface as TESTNET — the routing pivot (`entry.gov` check) cleanly extends the existing demand/provider split without forking the detail page structure.

**What works well:**
- Backwards-compat fallback in `entryFromSummary` is thoughtful — old cached snapshots (missing `treasury`/`gov`/`fees`) degrade to the naming-convention path, so the cron cadence change and the new fields can ship independently.
- Carrying explicit contract ids through the snapshot index (`buildRegistrySnapshot`) removes the Hiro round-trip on detail-page loads — good for perceived latency.
- Error-fallback block in `[id]/page.tsx` mirrors the main branch exactly. Easy to miss, correctly done.
- Cache TTL 5m→30m makes sense for testnet contracts; aligns with cron cadence.

**[suggestion] Import `TestnetBadge` in `LegionHeader.tsx` instead of duplicating it** (`app/legion/LegionHeader.tsx:60`)

The `TestnetBadge` component is exported from `LegionsClient.tsx`, but `LegionHeader.tsx` defines its own inline version with slightly different sizing (`text-[9px]` / `px-2 py-0.5` / `h-1 w-1` vs the exported `text-[10px]` / `px-2.5 py-1` / `h-1.5 w-1.5`). Two badge implementations means they'll drift visually. Consider importing the shared component:

```suggestion
          <TestnetBadge />
```

(and add `import { TestnetBadge } from "@/app/legions/LegionsClient";` at the top of the file — or move `TestnetBadge` to a shared `components/ui/` path so the import direction isn't awkward.)

**[nit] `||` vs `??` inconsistency in `entryFromSummary`** (`lib/legion/registry.ts:68`)

`treasury` uses `||` (falsy fallback) while `gov` and `fees` use `??` (nullish-only fallback). For contract-address strings, this is functionally equivalent — empty string would be invalid either way — but the inconsistency is easy to misread as intentional.

```suggestion
    treasury: summary.treasury ?? `${summary.owner}.legion-treasury`,
```

**Code quality notes:**

`TestnetBadge` lives in `LegionsClient.tsx` — a "Client" file is an unusual export point for a pure presentational component. If this badge spreads to more surfaces (e.g. the top nav, a sidebar), the import path will feel wrong. Moving it to `components/ui/TestnetBadge.tsx` now costs nothing and avoids the cross-import from `LegionHeader → LegionsClient`.

No CRITICAL performance issues, no BLOCKING accessibility findings. Approved.

Issue #1012, opened 2026-06-23
legion-gov v3.0: 3 findings from full lifecycle testnet verification


secret-mars:
Filing the 3 findings from a full end-to-end testnet verification of the v3.0 Legion contracts (`STBEMQQVSS3K3SQTF2NRZMF82JHMNTHQKQ2J7DW5.legion-{gov,treasury,fees}`) wired by #1011.

Full run report (49 testnet txs, every tx linked, expected/actual matrix, treasury settlement check):
**https://gist.github.com/secret-mars/da47c93aa58ae1a67b99d78517c932ac**

## Finding 1 — `Proposals.createdBtc` field is misleadingly named (contract)

`legion-gov` stores the proposal anchor as `createdBtc` and emits it under that name in `print` events, but the value is set from `stacks-block-height`, not `burn-block-height`. The Rail-A freshness gate then compares `inscription-height` against this stacks-block value:

```clarity
(asserts! (<= (- createdBtc inscription-height) FRESH_WINDOW) ERR_STALE)
```

If an operator/agent passes a real Bitcoin inscription height (the obvious reading of \"createdBtc\"), the freshness check either always-stales or always-futures depending on the burn↔stacks height gap.

**Suggested fix:** rename to `createdStx` in the struct, getter, and print event — *or* switch the lifecycle anchor to `burn-block-height` so the field name matches the value.

## Finding 2 — MCP buffer-encoding gotcha can be fixed at source instead of papered over in skill.md (tooling)

The current `skill.md` correctly warns:

> Content-hash encoding: Provide hex WITHOUT `0x` prefix; `0x`-prefixed values encode as empty buffers, causing collision failures

I hit this myself on first run — 4 of 5 proposes aborted with `ERR_DUP_HASH` because all four sent the same empty buffer. Workaround is exactly as documented (strip `0x`).

Root cause sits in `@aibtc/mcp-server` `call_contract`: passing `{type:"buffer", value:"0x..."}` silently encodes to a zero-length buffer rather than throwing or stripping the prefix. Every new agent will hit this exactly once and lose a tx fee + a content-hash slot before reading the warning.

**Suggested fix:** in the MCP server's buffer arg handler, either (a) auto-strip a leading `0x` so both forms work, or (b) throw on `0x`-prefixed input so the silent-empty failure mode is impossible. Doc note can stay as belt-and-suspenders.

## Finding 3 — 3-block veto window is too tight for human-coordinated veto on testnet (calibration)

[...]

PR #1024, opened 2026-07-06
fix: inbox drift (#995), legion faucet (#1016), challenge hardening (#770), competition sweep gate (#933), tenero token-set observability (#1025)

biwasxyz:
Issue-triage pass: validated each candidate against current code, reproduced in-code, and fixed the live defects. Skipped issues that were already resolved, out-of-repo, features, or docs-only (see triage notes at the bottom).

### #995 — Inbox `unreadCount` drift (phantom unreads) ✅
`GET /api/inbox/[address]` returned `unreadCount` straight from the denormalized `agent_inbox_stats.unread_count` counter (`route.ts:401`). That counter can drift above the actual `read_at IS NULL` row count because the write-path bump/decrement helpers are best-effort (`ctx.waitUntil` + swallowed `.catch()`), and nothing auto-reconciles — surfacing as unreads no PATCH can clear (they exist only in the counter, not as message rows).

**Fix:** read-path self-heal. When the unread view is fetched at offset 0 and the page isn't full, the entire unread set has been enumerated for free, so the true count is known. On a mismatch, the new single-address `recomputeAgentStats()` corrects the stored counter and returns the right value. Respects the team's no-`COUNT(*)`-on-hot-path rule (only fires on detected drift); the corrected counter then propagates to heartbeat and `status=all`.

### #1016 — Legion provider faucet cold-visitor cliff ✅
The originally-reported step-1 was already removed by the v1 provider-model refactor (#1018/#1019), but the provider CTA still promised "faucet + exact commands" while the `HowToProvide` section it anchors to had no faucet guidance. Added the repo-canonical faucet instruction (call `faucet` on the sBTC token) to the optional-stake step, matching sibling `HowToParticipate.tsx`.

### #770 — `/api/challenge` rate-limit hardening ✅
- Fail-closed now uses `shouldFailClosed(env)` instead of `if (env.DEPLOY_ENV)` — an empty-string `DEPLOY_ENV` was failing **open**.
- 503 on binding outage now sends `Retry-After`.
- IP bucket no longer collapses every IP-less caller into one shared `"unknown"` key (self-DoS).
- `x-forwarded-for` fallback is dev-only — `cf-connecting-ip` is the only trustworthy source on a CF Worker.

### #933 — Competition Hiro sweep 429-starving reputation/identity ✅
The sweep runs on the scheduler worker sharing the single `HIRO_API_KEY` with reputation/identity lookups, with no gate. Widened cadence 15min → 60min (4x fewer Hiro calls) and added a `COMPETITION_SWEEP_ENABLED` kill-switch (defaults **ON**; set `"false"` to disable during sunset). Agent self-submit via `POST /api/competition/trades` remains the primary ingestion path.

### #1025 — `getActiveTokenIds` masked D1 failures as a 3-token fallback ✅
The function swallowed D1 errors in a bare `catch` and returned `STATIC_TOKEN_IDS` — byte-identical to a legitimate static-only result — so a DB failure silently staled every non-core token's price with no signal. Return is now `{ tokenIds, degraded }`: a real query failure sets `degraded: true` (missing binding stays `false`), the Tenero cron logs `tenero.token_set_d1_failed`, and `/api/competition/tracked-tokens` reports `source: "db-error"`. Drive-by: refreshed stale `SchedulerDO` / "5 min cadence" doc references.

## Tests
New/updated regression tests for #995, #933, and #1025. All touched-area tests pass; typecheck + lint clean.

## Triage notes (no code change)
- **#1012** — all 3 findings are in the `legion-gov` Clarity contracts / `@aibtc/mcp-server`, not this repo.
- **#809** — RESOLVED/OBSOLETE: P&L column removed in #978 rework, `SchedulerDO` deleted for a cron worker, sBTC asset-id already corrected to `::sbtc-token`.
- **#781** — DOCS-ONLY and the docs already exist (`docs/edge-cache-pr-checklist.md` §7); zone-CDN `s-maxage` behavior is intentional.

### DIFF
--- a/app/api/challenge/route.ts
+++ b/app/api/challenge/route.ts
@@ -17,6 +17,7 @@ import {
   getAvailableActions,
 } from "@/lib/challenge";
 import { verifyBitcoinSignature } from "@/lib/bitcoin-verify";
+import { shouldFailClosed } from "@/lib/env";
 import { updateAgentInD1 } from "@/lib/d1/agents-mirror";
 import {
   publicKeyFromSignatureRsv,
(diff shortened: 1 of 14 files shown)


arc0btc (approved):
Four independent, well-scoped fixes with regression tests for each — this is exactly the kind of issue-triage pass we like to see (validated against live code, not just the original report).

**What works well:**
- **#995 (inbox self-heal)**: the "prove full enumeration, then trust the count" logic (`offset === 0 && receivedMessages.length < limit`) is the right way to get a correct count without reintroducing a `COUNT(*)` hot path — matches the team's own no-COUNT(*) rule and reuses the existing partial indexes (`idx_inbox_to_btc_sent_at`, `idx_inbox_outbox_from_btc_sent_at`) so the recompute query is indexed, not a scan. Three regression tests cover the drift/no-drift/full-page-can't-prove cases precisely.
- **#770 (challenge hardening)**: `shouldFailClosed(env)` instead of `if (env.DEPLOY_ENV)` closes a real gap — an empty-string `DEPLOY_ENV` would've evaluated truthy-checked-as-falsy in the old code and failed open in what should be a closed environment. Skipping the rate limit entirely when no IP is available (instead of collapsing into a shared `"unknown"` bucket) is the correct call — a shared bucket is a self-inflicted DoS vector.
- **#933 (competition sweep gate)**: `disabledCompetitionSummary()` keeps the disabled path's return shape identical to a real run, so `runCompetitionNow` callers and the KV state writer don't need a special case — nice bit of interface discipline. Widening 15min → 60min for a catch-up sweep with a primary self-submit ingestion path is a sound trade for the shared `HIRO_API_KEY` budget.
- **#1016**: correctly identified that the originally-reported gap was already closed by the v1 refactor and scoped the fix to the actual remaining gap (missing faucet copy) rather than re-fixing something already fixed.

**Code quality notes:**
- `recomputeAgentStats` mirrors the existing `rebuildAllStats` column mapping (`to_btc_address AS btc_address`) and upsert pattern — good reuse of an established convention rather than inventing a new shape.
- No blocking issues found across functionality, security, performance, or architecture fit. All four fixes are scoped to their reported issue with no drive-by changes.

**[question]** The self-heal in `route.ts` only fires when `statusFilter === "unread"`. If a client requests `status=all` or `status=read` on a drifted account, the returned `unreadCount` field could still reflect the stale counter (only the `unread` view path proves full enumeration for free). That looks like an intentional scope decision to avoid extra queries on other views rather than an oversight — just flagging in case `all`/`read` drift is also worth self-healing in a follow-up, since `unreadCount` is returned regardless of which view was requested.

Operationally: we don't run this D1 schema ourselves, but the partial-index reuse and the "only recompute on proven drift" design are the right shape for a low-traffic self-heal — approving.

Issue #1021, opened 2026-06-30
Scale: stop full-re-walking the legion registry every cron tick (chainhook + thin index)


biwasxyz:
## Problem

The `/legions` registry index is rebuilt by **fully re-walking the registry every cron tick** (`buildRegistrySnapshot` → `listLegions` → `get-count` + `get-legion` per id, then `get-balance` + a count per entry). This is O(N) Hiro calls every 5 minutes and does not scale.

Concretely, per rebuild ≈ **2 Hiro calls × N + 1 count/entry**:
- At the current `MAX_LEGIONS = 100` cap → ~200 Hiro calls / 5 min.
- At 1000 legions (cap lifted) → ~2,000 Hiro calls / tick ≈ **24k/hour** → Hiro rate limits, slow rebuilds, wasted cost.

There is also a latent correctness bug: `getCount` does `Math.min(n, 100)` (`lib/legion/registry.ts:31,114`), so **beyond 100 legions the extra entries silently never appear** in the index.

## Current shape
- `lib/legion/snapshot.ts` → `buildRegistrySnapshot` (full walk + per-entry `get-balance`/count)
- `lib/legion/read.ts` → 5-min cron + SWR, single D1 row `__registry__` (upsert overwrite)
- `lib/legion/registry.ts` → `MAX_LEGIONS = 100`
- No chainhook/event path for the registry (chainhooks exist for competition only).

## Proposed fix

1. **Event-driven via chainhook.** Subscribe to the `legion-registry` `print` events (`register` / `set-uri` / `set-active` / `transfer-entry` all emit). On an event, upsert only that one legion row. Removes the periodic full walk entirely.
2. **Thin index.** The list view only needs `id/owner/model/uri/active` — all from the single `get-legion`. Defer `get-balance` + proposal/provider counts to the detail page (lazy). Kills the 2× per-entry multiplier.
3. **Remove / raise the 100 cap** once the walk is no longer O(N) per tick; paginate or rotate if any polling remains.

## Not urgent
Fine for current scale (7 model legions + 1 demand). This is a roadmap item for when the registry grows past ~100 entries. No demo impact.

biwasxyz:
Resolved by #1072. The legions page is now event-driven: Hiro Chainhooks deliver to `POST /api/legions/chainhook`, which writes D1 `legion_events`, and there is no cron re-walk. `lib/legion/registry.ts`, `snapshot.ts` and the `MAX_LEGIONS` cap were removed.

PR #1027, opened 2026-07-06
perf(cost): cut cron Hiro fan-out, coalesce heartbeat KV writes, drop redundant D1 scans

biwasxyz:
Cost-optimization pass from a 3-axis audit (D1 rows-scanned, KV ops, external Hiro/subrequest calls). The codebase is already well-optimized (maintained counters, edge-cached GETs, ratelimits bindings), so these are the remaining real offenders. No user-visible behavior change beyond the bounded staleness noted per item.

## 🔴 Legion cron — biggest win (external Hiro cost)
The Legion snapshot was the **one ungated scheduler task** — it ran its large per-member (stake+balance), per-proposal (status), and per-voter (record) Hiro fan-out on **every 15-min tick**, on the **shared `HIRO_API_KEY`** that also feeds identity/reputation (~475k reads/mo for a single legion). This is literally why the cron fired every 15 min.
- **Gate it to hourly** via a `scheduler:legion` `lastRunAt` check, exactly like Tenero/competition/earnings.
- **Widen the cron `*/15` → `*/30`** — earnings (30-min) still runs each tick; the hourly tasks gate themselves. Nothing needs a 15-min wakeup now.
- **Dedupe the double `listLegions()` walk**: reuse the entries `buildRegistrySnapshot` already resolved (`entryFromSummary`) instead of a second `get-count + N×get-legion` round-trip per run. (overlaps #1021)
- **Net: ~4× fewer Legion Hiro reads**; stops it starving the shared key. `/legions` pages read the D1 snapshot behind `caches.default`, so reader freshness is unchanged.

## 🔴 Heartbeat KV write — highest-volume KV write
Heartbeat wrote the full `btc:` agent record to KV on **every check-in** (every agent, ~every 5 min), but the only changing fields are `lastActiveAt`/`lastCheckInAt`, and **D1 (written every check-in) is the source of truth** for all aggregate consumers (agents list / activity / leaderboard rebuild from D1).
- **Coalesce**: skip the KV write while the stored `lastActiveAt` is < 1h old → ~**12× fewer** heartbeat KV writes (once/5min → ~once/hour per agent).
- **Tradeoff**: a direct `btc:` lookup can see a ≤1h-stale `lastActiveAt`; D1 always has the fresh value. No aggregate reads `lastActiveAt` from KV.

## 🟡 D1 redundant scans
- **`listSubmissionsForBounty`**: dropped the per-request `COUNT(*)` scan; use the parent bounty's maintained `submission_count` (every caller already loads the bounty).
- **`tracked-tokens`**: raised cache `60s → 300s` in front of `getActiveTokenIds`' unbounded `swaps` GROUP BY (the token set shifts slowly; the cron scan is hourly).

## 🟡 KV N+1
- **`vouch/[address]`**: resolve referee display info from the cached agent list (**1 read + Map lookups**) instead of one `kv.get(btc:)` per referee; and **skip the read entirely** when the agent has vouched for nobody (the common case).

## Deliberately left (bounded / low-frequency — follow-up candidates)
- **capabilities** N+1 — re-validates capabilities against the source `btc:` record and is behind a 1h edge cache (amplitude capped).
- **inbox sent-view** N+1 and **referral-code** POST N+1 — bounded page size / low frequency.
These want a shared "resolve many agents" helper; out of scope here to keep the PR safe.

## Testing
+4 tests (heartbeat coalesce ×2, vouch batching ×2). **All 1516 app/api + lib tests pass; non-test typecheck clean.** This PR produces a Cloudflare Workers **preview URL** — worth spot-checking `/legions`, `/api/vouch/[addr]`, `/api/bounties/[id]/submissions`, and a heartbeat POST against real bindings before merge.

## Not included
Full KV→D1 authoritative migration (making D1 the source of truth for durable records) is a separate phased effort under #652 — this PR only *reduces* KV/D1/Hiro cost without changing which store is authoritative.

### DIFF
--- a/app/api/bounties/[id]/route.ts
+++ b/app/api/bounties/[id]/route.ts
@@ -53,7 +53,8 @@ export async function GET(
     db,
     bounty.id,
     20,
-    0
+    0,
+    bounty.submissionCount
   );
 
   // Winner block — populated whenever the bounty has acceptedAt (i.e. on
--- a/app/api/bounties/[id]/submissions/route.ts
+++ b/app/api/bounties/[id]/submissions/route.ts
@@ -39,7 +39,13 @@ export async function GET(
   const limit = clampInt(url.searchParams.get("limit"), 20, 1, 100);
   const offset = clampInt(url.searchParams.get("offset"), 0, 0, 100_000);
 
-  const { submissions, total } = await listSubmissionsForBounty(db, id, limit, offset);
+  const { submissions, total } = await listSubmissionsForBounty(
+    db,
+    id,
+    limit,
+    offset,
+    bounty.submissionCount
+  );
   const nextOffset = offset + submissions.length < total ? offset + submissions.length : null;
 
   return NextResponse.json(
--- a/app/api/competition/tracked-tokens/route.ts
+++ b/app/api/competition/tracked-tokens/route.ts
@@ -82,7 +82,11 @@ export async function GET(request: NextRequest) {
     },
     {
       headers: {
-        "Cache-Control": "public, max-age=60, s-maxage=60",
+        // 5 min. The active set is derived from an unbounded full-scan GROUP BY
+        // over the swaps table (getActiveTokenIds), so cache-misses are the
+        // expensive path; the set only shifts as agents trade new tokens, which
+        // is slow, so a longer TTL cuts scan frequency ~5x with no real staleness.
+        "Cache-Control": "public, max-age=300, s-maxage=300",
       },
     }
   );
--- a/app/api/heartbeat/__tests__/route.test.ts
+++ b/app/api/heartbeat/__tests__/route.test.ts
@@ -78,6 +78,7 @@ vi.mock("@/lib/name-generator", () => ({
 vi.mock("@/lib/heartbeat", () => ({
   CHECK_IN_MESSAGE_FORMAT: "AIBTC Check-In | {timestamp}",
   CHECK_IN_RATE_LIMIT_SECONDS: 60,
+  HEARTBEAT_KV_WRITE_COALESCE_MS: 60 * 60 * 1000,
   buildCheckInMessage: vi.fn().mockReturnValue("AIBTC Check-In | 2026-01-01T00:00:00.000Z"),
   validateCheckInBody: vi.fn().mockReturnValue({ data: {} }),
 }));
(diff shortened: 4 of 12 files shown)


arc0btc (approved):
Solid cost-optimization pass — the audit approach (D1 rows-scanned / KV ops / Hiro subrequests) is the right lens, and every fix ties back to a concrete measured offender rather than guessing. Traced the four changes against the actual call sites, not just the diff:

**What works well:**
- `listSubmissionsForBounty`'s `knownTotal` param is a clean win — verified `insertSubmission` bumps `submission_count` transactionally in the same D1 batch as the insert (`lib/bounty/d1-helpers.ts:521`), so the maintained counter is trustworthy and both callers (`bounties/[id]/route.ts`, `bounties/[id]/submissions/route.ts`, `bounties/[id]/page.tsx`) already load the bounty first. Good default-preserving fallback (`if (total === undefined)`) for any caller that doesn't pass it.
- Heartbeat KV coalescing: confirmed `updatedAgent` only ever differs from the stored `agent` in `lastActiveAt`/`lastCheckInAt` on this path (`app/api/heartbeat/route.ts:486`) — no other field mutation gets silently dropped by skipping the write. The `!Number.isFinite(prevActiveMs)` fallback correctly forces a write for first-ever/unparseable timestamps.
- Legion dedup: verified `buildRegistrySnapshot` already resolves full `LegionSummary[]` (`lib/legion/snapshot.ts:236`) with every field `entryFromSummary` needs (treasury/gov/fees/model/uri/active/source) — reusing it instead of a second `listLegions()` walk is a genuine win, not a shortcut that drops data.
- Gating pattern for `legionDue` matches the existing `teneroDue`/`competitionDue` convention exactly (`now + 1_000` slack, same lastRunAt-in-KV shape) — consistent, not a one-off variant.
- Good test coverage for the two behavior-changing paths (heartbeat coalesce skip/write, vouch batched-vs-skipped cache read).

**[question] Heartbeat KV/pubkey-capture race** (`app/api/heartbeat/route.ts:435-439` vs the new coalesced write)
The opportunistic `persistBtcPubkeyIfMissing` write via `ctx.waitUntil` and the main coalesced `kv.put` can still race today — whichever finishes last wins. That race predates this PR (the old code had the same two independent writers), so not a regression, but worth confirming it's an accepted pre-existing tradeoff rather than something this PR should've addressed while touching this code path.

**[nit]** `wrangler.scheduler.jsonc` comment says "cost: 2x fewer wakeups" for the cron widen, which is correct, but combined with Legion's ~4x fewer Hiro reads and the KV coalesce's ~12x, might be worth a one-line rollup in the PR description of the *combined* effect on the shared `HIRO_API_KEY` budget — mostly for whoever reviews the Hiro dashard next week and wants the "why" without re-deriving it.

**Code quality notes:**
- No reuse/simplification issues — `entryFromSummary` and `getCachedAgentList` are both existing helpers being *reused* here, which is exactly right (avoids introducing a parallel lookup path).
- The `HEARTBEAT_KV_WRITE_COALESCE_MS` constant and its doc comment are a nice pattern — self-documenting the tradeoff at the definition site rather than only at the call site.

**Operational context:** we don't operate this repo's cron/KV directly, so no first-hand incident data to add here — but the failure mode being closed (a single ungated 15-min task starving a shared external API key from unrelated features) is a pattern we watch for in our own dispatch scheduling, and gating-by-last-run-timestamp-in-KV is the same fix shape we'd reach for.

No blocking issues. This is scoped entirely to `app/api/`, `lib/`, and cron config — no `components/`/page UI changes, so the React performance / composition / accessibility checklist doesn't apply to this diff.

**Approve.**

PR #1028, opened 2026-07-08
perf(cost): cache profile sBTC balance, index competition results by agent

biwasxyz:
Two remaining metered/rate-limited read leaks surfaced by the post-#1027 cost audit (the D1 / KV / external-API sweep). KV came back clean; these are the only two hot-path findings worth acting on.

## 1. Uncached external balance calls on every profile render
`lib/balances/btc.ts` · `app/agents/[address]/page.tsx`

The agent-profile page is dynamic (no `revalidate`), so **every view — crawlers and OG bots included — made two uncached external calls**: Hiro `/balances` (sBTC, L2) under the shared `HIRO_API_KEY` (metered against quota), and mempool.space (native BTC, L1, public but rate-limited). Identity + BNS on the same page were already KV-cached; these balance calls were the leaks left on the path.

Both legs are now cached in KV (30-minute TTL), one key per leg:
- `cache:sbtc-balance:{stxAddress}` (L2 / Hiro)
- `cache:btc-balance:{btcAddress}` (L1 / mempool.space)

Per leg:
- **Cache hit → zero upstream calls.** Repeat views of the same profile within the window reuse the cached number.
- **Only authoritative answers are cached** (a real balance, including a genuine 0). A transient upstream failure returns 0 **uncached**, so the next view retries.
- **Best-effort write** — a KV write failure never fails the balance read.

Corrected the file's stale "L1 is public and free, no caching" comment.

## 2. Full-table scan on `/api/competition/status`
`migrations/025_competition_round_results_agent_index.sql`

`getLatestFinalizedRoundResultForAgent` filters `competition_round_results` by `stx_address` alone, but both existing indexes (PK `(round_id, stx_address)` and `idx_..._rank (round_id, rank)`) lead with `round_id` — so SQLite full-scanned the table. It gains one row per eligible agent per finalized round, so the scan cost grows permanently. This is the endpoint agents poll for standings (edge cache only `s-maxage=10`), so most distinct-address requests reach D1.

New index `idx_competition_round_results_agent(stx_address)` turns the scan into a point-seek.

## Tests
- `lib/balances/__tests__/btc.test.ts` (new, 7 tests): for each leg — cache hit skips upstream, fresh answer (incl. 0) written with TTL, transient failure not cached — plus a no-KV-binding pass-through.
- Typecheck: no new errors (the 150 pre-existing test-file errors are unchanged from `main`). Lint clean.

## Deploy note
Migration 025 applies via the standard `wrangler d1 migrations apply landing-page --remote --env production` path (index-only, `IF NOT EXISTS`, no data change).

### DIFF
--- a/app/agents/[address]/page.tsx
+++ b/app/agents/[address]/page.tsx
@@ -336,7 +336,8 @@ export default async function AgentProfilePage({
     const btcBalance = await fetchBtcBalance(
       agentWithIdentity.btcAddress,
       agentWithIdentity.stxAddress,
-      env.HIRO_API_KEY
+      env.HIRO_API_KEY,
+      kv
     );
 
     return (
--- a/lib/balances/__tests__/btc.test.ts
+++ b/lib/balances/__tests__/btc.test.ts
@@ -1,0 +1,125 @@
+/**
+ * L2 (sBTC) balance cache — the cost fix that collapses repeat profile views
+ * into one metered Hiro call. Locks three behaviors:
+ *   1. cache hit returns without calling Hiro
+ *   2. a real answer (incl. genuine 0) is written to KV with a TTL
+ *   3. a transient upstream failure is NOT cached (next view retries)
+ */
+
+import { describe, it, expect, vi, beforeEach, afterEach } from "vitest";
+import { fetchBtcBalance } from "../btc";
+
+const STX = "SP2J6ZY48GV1EZ5V2V5RB9MP66SW86PYKKNRV9EJ7";
+const BTC = "bc1qexampleexampleexampleexampleexampleex";
+
+function mockKv(initial: Record<string, string> = {}) {
+  const store = new Map(Object.entries(initial));
+  return {
+    get: vi.fn(async (k: string) => store.get(k) ?? null),
+    put: vi.fn(async (k: string, v: string) => {
+      store.set(k, v);
+    }),
+    delete: vi.fn(),
+    list: vi.fn(),
+    getWithMetadata: vi.fn(),
+    __store: store,
+  } as unknown as KVNamespace & { __store: Map<string, string> };
+}
+
+/** Build a Hiro /balances response with the given sBTC balance string. */
+function balancesResponse(sbtcBalance: string) {
+  // Asset id must match SBTC_ASSET_ID built in btc.ts from SBTC_CONTRACTS.mainnet.
+  return {
+    ok: true,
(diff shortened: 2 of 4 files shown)


arc0btc (approved):
Two clean, well-scoped fixes to metered-read leaks from the post-#1027 cost audit — cache the L2 sBTC balance on profile SSR, and add a leading index so the competition-status endpoint stops full-scanning `competition_round_results`.

**What works well:**
- `fetchL2Sats`'s cache-write discipline is exactly right: only an authoritative answer (including a genuine `0`) gets written to KV, while a `!res.ok` failure returns `0` uncached so the next view retries against Hiro instead of pinning a bad answer for 90s. The comment calls this out explicitly, which is the kind of non-obvious invariant worth documenting.
- The KV write is wrapped in try/catch so a cache-write failure can't turn a successful balance fetch into a 500 — matches the file's existing "degrade to 0, never throw" philosophy for this whole function.
- `kv` is threaded through as an optional param with the no-KV path preserved (test 4 in `btc.test.ts` locks this), so this doesn't silently break any other caller of `fetchBtcBalance`.
- The migration comment for #025 traces the exact query (`getLatestFinalizedRoundResultForAgent`) and correctly diagnoses that both existing indexes on `competition_round_results` (PK `(round_id, stx_address)`, `idx_competition_round_results_rank(round_id, rank)`) lead with `round_id`, so a `WHERE stx_address = ?1` filter with no `round_id` predicate can't use either — I checked the migration 017 source and the live query in `lib/competition/finalize/read.ts` and the diagnosis is accurate. `CREATE INDEX IF NOT EXISTS` makes the migration safely re-runnable.
- Test coverage on the cache is solid: hit/miss/failure-not-cached/no-KV-fallback are all exercised, including asserting the exact TTL passed to `kv.put`.

**Code quality notes:**
- No reuse or simplification issues — `L2_BALANCE_CACHE_PREFIX`/`_TTL_SECONDS` as named constants is the right amount of structure for a two-line cache shim, nothing over-built here.

**[question]** 90s TTL — is that tuned to expected crawler/OG-bot re-visit cadence, or a round number? Not blocking either way, just curious if there's a data point behind it since the L1 leg is uncached for comparison.

Operationally: this is the right shape for the fix — matches the "cache only authoritative reads, never cache failure" pattern we use elsewhere for metered upstream calls, and the index fix is a textbook leading-column diagnosis. Approving.

PR #1042, opened 2026-07-21
fix(deps): patch brace-expansion DoS (CVE-2026-13149)

arc0btc:
## Summary
- Fixes dependabot alert #134 (GHSA-3jxr-9vmj-r5cp, CVE-2026-13149) — exponential-time DoS in `brace-expansion`'s `expand()`.
- All 4 transitive installs are devDependency-only (eslint/typescript-estree, @node-minify/core, glob, minimatch tooling used at build time) — no production runtime path passes untrusted input to `expand()`, so exploitability here is low, but the fix is cheap and correct so no reason to accept the risk.
- Added scoped `overrides` in `package.json`, each pinned to the patched release **within its original major** to avoid an unintended cross-major jump for any single consumer:
  - top-level `minimatch` → `brace-expansion` 1.1.13 → 1.1.16
  - `@typescript-eslint/typescript-estree`'s `minimatch` → `brace-expansion` 2.0.3 → 2.1.2
  - `@node-minify/core`'s nested `glob > minimatch` → `brace-expansion` 2.0.3 → 2.1.2
  - `glob`'s `minimatch` → `brace-expansion` 5.0.5 → 5.0.7
- Regenerated `package-lock.json` via `npm install --package-lock-only`; no dependency ranges in `package.json`'s direct deps changed.

## Test plan
- [x] Verified via lockfile inspection that each brace-expansion instance lands on its patched version without a major-version jump for any consumer
- [ ] CI: typecheck / build / vitest run green

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3067,9 +3067,9 @@
       }
     },
     "node_modules/@node-minify/core/node_modules/brace-expansion": {
-      "version": "2.0.3",
-      "resolved": "https://registry.npmjs.org/brace-expansion/-/brace-expansion-2.0.3.tgz",
-      "integrity": "sha512-MCV/fYJEbqx68aE58kv2cA/kiky1G8vux3OR6/jbS+jIMe/6fJWa0DTzJU7dqijOWYwHi1t29FlfYI9uytqlpA==",
+      "version": "2.1.2",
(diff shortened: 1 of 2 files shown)


biwasxyz:
Superseded by #1052, which is the same brace-expansion patch and merges cleanly. This branch is CONFLICTING against main.

PR #1046, opened 2026-07-23
docs(claude-md): correct leaderboard section, document earnings ledger

biwasxyz:
## Why

Two gaps found while auditing open issues against the codebase.

**1. The leaderboard section was stale.** `/leaderboard` was rewritten as a verified on-chain earnings board (#978, PRs #979-#992), but `CLAUDE.md` still described it as the trading-comp ranking with Trades / Volume / P&L columns, chip-based sort, and "P&L computed client-side from `/api/competition/trades` data + Tenero prices (5-min localStorage cache)." None of that is true now: the page is SSR from D1, and `LeaderboardClient.tsx` makes zero Tenero calls.

**2. The earnings ledger had no coverage at all.** Before this PR, `grep -i earning CLAUDE.md` returned nothing. `lib/earnings/` (10 modules), migrations `020`-`022`, the 30-min cron sweep, and both public read endpoints were undocumented, even though every comparable subsystem (Inbox, Identity, Bounty, Competition Finalize) has its own section.

## What changed

**Leaderboard**
- UX dual-interface line and the Pages (UX) entry rewritten: SSR via `getEarningsBoard()`, migration `022` partial index, 5-min `caches.default` layer plus in-flight singleflight, 5000-row backstop, client paginating at 50/page.
- Split out `LeaderboardClient.tsx` and `EarningsMethodologyModal.tsx` as their own entries.
- Added `ClubBadge.tsx` and `EarningsActivity.tsx` to Components (UX). Both were missing, and the new leaderboard text references `ClubBadge`.
- Disambiguated `/api/leaderboard` in the endpoint table. It is the agent directory ranking and has nothing to do with the `/leaderboard` page. This is a real trip hazard for anyone reading the doc cold.

**New "Earnings Ledger" section**, placed after Competition Finalize to match the existing subsystem ordering. Covers the indexer-only design, the three in-scope assets and the `verified_at` earnings floor, the txid-based classification order (including why `x402_endpoint` is inert and `exchange_or_external` unpopulated), anti-gaming precedence and the three heuristics, index-time pricing and why Tenero forces it, the sweep cadence constants including the Cloudflare 6-connection cap on `EARNINGS_FETCH_CONCURRENCY`, the four D1 tables and why each index exists, and both public read endpoints with their cache TTLs and param ranges.

Also adds an Earnings row group to the API Endpoints table.

## Verification

Docs-only, no code touched. Every value was read off the source rather than reconstructed from the issue text:

- Cadence and concurrency constants: `lib/earnings/constants.ts:13-24`
- Cursor key and its D1 table: `lib/earnings/d1.ts:115` (`competition_state`, key `earnings_scheduler_cursor`)
- Classification priority and the inert `x402_endpoint` branch: `lib/earnings/classify.ts`
- Override precedence and the `agent_peer`-only heuristic scope: `lib/earnings/anti-gaming.ts:198`
- Cache TTLs and param ranges: `app/api/agents/[address]/earnings/route.ts:16-18`, `app/api/stats/earnings/route.ts:14-15`
- Leaderboard page sizing and sort chips: `app/leaderboard/page.tsx:127`, `LeaderboardClient.tsx:24-33`

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -31,7 +31,7 @@ npm run cf-typegen   # Generate Cloudflare Workers TypeScript types
 
 Every feature is designed for two audiences simultaneously:
 
-- **UX (User Experience)** — Browser-based pages for humans (`app/page.tsx`, `app/agents/`, `app/leaderboard/` (trading-comp ranking with P&L and chip-based sort), `app/guide/`)
+- **UX (User Experience)** — Browser-based pages for humans (`app/page.tsx`, `app/agents/`, `app/leaderboard/` (verified on-chain earnings ranking), `app/guide/`)
 - **AX (Agent Experience)** — API-first endpoints for AI agents. Every API route self-documents on GET (returns usage instructions as JSON). Agents discover the platform via the discovery chain.
 
 ### Agent Discovery Chain
@@ -110,7 +110,13 @@ During registration (POST /api/register), after both signatures are verified, th
 | Route | Methods | Purpose |
 |-------|---------|---------|
 | `/api/levels` | GET | Self-documenting level system reference |
-| `/api/leaderboard` | GET | Ranked agents with `?level`, `?limit`, `?offset` params |
+| `/api/leaderboard` | GET | Ranked agents with `?level`, `?limit`, `?offset` params. Note: this is the agent directory ranking, **not** the `/leaderboard` page (that renders the earnings board, see below) |
+
+### Earnings (see Earnings Ledger below)
+| Route | Methods | Purpose |
+|-------|---------|---------|
+| `/api/agents/[address]/earnings` | GET | Per-agent earnings rollup, source breakdown, paginated line items |
+| `/api/stats/earnings` | GET | Platform-aggregate earnings + top-100 ranking (`?window=7d\|30d\|lifetime`) |
 
 ### Claims & Rewards
 | Route | Methods | Purpose |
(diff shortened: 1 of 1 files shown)


arc0btc (approved):
Doc-only PR — corrects the CLAUDE.md leaderboard description (was describing the old trading-comp P&L leaderboard; `/leaderboard` now renders verified on-chain earnings per issue #978) and adds a full Earnings Ledger section. No code changes, so the React performance/composition/a11y checklist doesn't apply here — reviewed as documentation accuracy.

**What works well:**
- Verified every file/route/migration named in the new section actually exists in the repo (`lib/earnings/*.ts`, migrations `020`–`022`, `app/components/ClubBadge.tsx` + `EarningsActivity.tsx`, `app/leaderboard/{LeaderboardClient,EarningsMethodologyModal}.tsx`, `docs/earnings-ledger-architecture.md`) — nothing here is aspirational or stale.
- The classification-order table and anti-gaming heuristics section are exactly the kind of detail that saves a future reader from re-deriving tricky invariants (e.g. "self-funded only excludes on two confident `ok` lookups," "ring excludes both legs"). That's the right altitude for a CLAUDE.md.
- Good catch clarifying `/api/leaderboard` (agent directory ranking) vs `/leaderboard` (earnings board) — those names collide easily and this disambiguation will save a support cycle down the line.
- Concurrency/cadence constants table (`EARNINGS_FETCH_CONCURRENCY` ≤ 5 for the CF Worker connection cap, etc.) documents an operational constraint that isn't obvious from the code alone.

**[nit]** The updated `app/leaderboard/page.tsx` bullet is dense (SSR path, caching layer, singleflight map, TTL alignment) — all accurate per the code references, just flagging in case a future trim pass wants to push some of that detail into `docs/earnings-ledger-architecture.md` and keep the file-tree bullet list scannable.

No blocking issues. Approving.

PR #1050, opened 2026-07-29
docs(discovery): add earnings ledger coverage, fix competition drift

biwasxyz:
Found while updating #815 (the trading-competition rules reference). Auditing the discovery chain against the code turned up one gap and several statements that had gone wrong.

## The gap: earnings ledger is invisible to agents

The earnings ledger (#978) shipped `/api/agents/{address}/earnings` and `/api/stats/earnings`, but "earnings" appeared **zero times** across `llms.txt`, `llms-full.txt`, `openapi.json`, `agent.json`, and the `/docs` index. An agent following the discovery chain could not find either endpoint, and nothing explained what the Club tier chips on profiles mean. CLAUDE.md requires discovery docs to be updated alongside endpoints; that did not happen for #978.

## The drift: competition statements that mislead

| Where | Said | Actually |
|---|---|---|
| `llms.txt` | `aibtc.com/leaderboard` has Trades / Volume / Unrealized P&L chips, default Trades desc | That page is the **earnings** board; chips are Earnings / Payers / Latest |
| `llms.txt` | P&L is mark-to-current, "both legs re-price on every render" | Scoring runs against a per-round frozen price snapshot |
| `llms.txt`, `llms-full.txt` | catch-up sweep runs every 15 min | Once a day (`COMPETITION_INTERVAL_MS`, `lib/scheduler/cron-runner.ts:72`) |
| `llms.txt` | no mention of `/api/competition/rounds*` | Standings live there; an agent stopping at the quick-start cannot find its placement |
| `llms.txt` Pages list | `/leaderboard` is "Ranked agents by level (API: /api/leaderboard)" | Two unrelated systems; CLAUDE.md flags this exact confusion |

The cadence one matters most in practice: at a 24h lag, telling agents that self-submitting a txid "just gets you scored sooner" understates it. The docs now say submitting is the reliable path.

Also indexed the `competition-finalize` topic doc, which existed in `TOPICS` and in the 404 hint but was missing from `/docs`.

## Changes

- **`llms.txt`** — new Earnings section (what counts, what is excluded as self-dealing, pricing, Club tiers); rounds endpoints; corrected ranking, P&L methodology, sweep cadence, and both leaderboard references.
- **`llms-full.txt`** — new Earnings Ledger reference section with response shapes, the classification table, anti-gaming rules, and pricing; cadence corrections; disambiguated the directory leaderboard.
- **`openapi.json`** — added both earnings paths with full response schemas.
[...]

PR #1051, opened 2026-07-30
fix(levels): remove unbacked "leaderboard bonus" reward promise

biwasxyz:
Found while auditing the discovery chain in #1050.

## The problem

Genesis (Level 2) has advertised a "leaderboard bonus" since the level system was written. Nothing implements one. There is no bonus field, no scoring adjustment, and no code path in `lib/` that reads level when ranking anything.

Neither leaderboard has a concept it could apply to:

- `/api/leaderboard` sorts by level, then registration date.
- The `/leaderboard` page ranks by verified on-chain earnings, which by design only move when a counterparty actually pays the agent. A level-based bonus would defeat the whole point of the ledger.

The other three items in that list (Genesis badge, vouching, trading competition eligibility) are all real and stay.

## Why it matters beyond copy

This was not docs-only. It shipped in live API response bodies:

| Surface | Field |
|---|---|
| `POST /api/register` | `nextStep.reward`, and the onboarding `steps[].reward` |
| `GET /api/claims/viral` | `unlocks` |
| `GET /api/levels` | `howToLevelUp["1 → 2 (Genesis)"].reward` |
| `/.well-known/agent.json` | `progression.levels[].reward` and the claims skill description |

So an agent completing its viral claim was told, in a machine-readable field, that it had earned something that does not exist. `app/api/register/route.ts` was already self-inconsistent about it: `nextStep.description` omitted the bonus while `nextStep.reward` on the very next line promised it.

The prose versions were the most explicit. From `app/skill.md/route.ts`:

> On success: you are now **Level 2 (Genesis)**. You can vouch for new agents, submit trades to the trading competition, wear the Genesis badge, and **earn a leaderboard bonus**.

## Changes

13 occurrences across 9 files. Canonical string is `lib/levels.ts:58`; removed there and at every call site, including the prose variants in `skill.md`, `llms.txt`, `llms-full.txt`, and the schema.org `HowTo` block in `layout.tsx`.

Same class as `634e401 fix(claims): remove Genesis/viral reward promise copy`.

## Verification

- `npm run test` — 100 files, 1535 passed, 5 skipped. No test asserted on the string.
- `npm run build` — compiled successfully
- `grep -rn "leaderboard bonus"` — zero remaining

## Follow-up worth filing

Four routes hardcode this reward string rather than importing `LEVELS` from `lib/levels.ts`, which is why a one-line fix became thirteen. `app/api/levels/route.ts` already imports `LEVELS` and *still* hardcodes it. That duplication is the mechanism by which this drifted in the first place, and it is worth deduping on its own.

### DIFF
--- a/app/.well-known/agent.json/route.ts
+++ b/app/.well-known/agent.json/route.ts
@@ -183,7 +183,7 @@ export function GET() {
           level: 2,
           name: "Genesis",
           unlock: "Tweet + claim via POST /api/claims/viral",
-          reward: "Genesis badge + vouching + trading competition eligibility + leaderboard bonus",
+          reward: "Genesis badge + vouching + trading competition eligibility",
         },
       ],
       checkEndpoint: "GET /api/verify/{address}",
@@ -505,7 +505,7 @@ export function GET() {
           "Reach Genesis (Level 2) by tweeting about your registered AIBTC agent. " +
           "Requires a valid claim code (from registration or POST /api/claims/code). " +
           "Include the code in your tweet, then POST btcAddress and tweetUrl to " +
-          "/api/claims/viral. Unlocks: Genesis badge + vouching + trading competition eligibility + leaderboard bonus. " +
+          "/api/claims/viral. Unlocks: Genesis badge + vouching + trading competition eligibility. " +
           "Successful claim upgrades you to Level 2 (Genesis).",
         tags: ["genesis", "x", "viral", "level-up"],
         examples: [
--- a/app/api/claims/viral/route.ts
+++ b/app/api/claims/viral/route.ts
@@ -331,7 +331,7 @@ export async function GET(request: NextRequest) {
           },
         },
       },
-      unlocks: "Genesis badge + vouching + trading competition eligibility + leaderboard bonus",
+      unlocks: "Genesis badge + vouching + trading competition eligibility",
       documentation: {
         registerFirst: "https://aibtc.com/api/register",
         fullDocs: "https://aibtc.com/llms-full.txt",
--- a/app/api/levels/route.ts
+++ b/app/api/levels/route.ts
@@ -58,7 +58,7 @@ export function GET() {
           action: `Tweet about your agent mentioning 'AIBTC', your claim code, your agent name, and tag ${X_HANDLE}`,
           endpoint: "POST /api/claims/viral",
           body: { btcAddress: "your-btc-address", tweetUrl: "https://x.com/you/status/..." },
-          reward: "Genesis badge + vouching + trading competition eligibility + leaderboard bonus",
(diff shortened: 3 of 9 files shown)


arc0btc (approved):
Removes the "leaderboard bonus" reward promise everywhere it was advertised, since nothing implements one — no bonus field, no scoring adjustment, no leaderboard code path that reads level. Good catch, and the writeup is thorough about *why* it matters (it wasn't just prose — it shipped in machine-readable `reward`/`unlocks` fields that agents parse programmatically).

**What works well:**
- Verified both leaderboard surfaces (`/api/leaderboard` sorts by level+date, `/leaderboard` page ranks by verified on-chain earnings) confirm there's genuinely no bonus mechanism to preserve — this isn't a "forgot to implement" gap, it's dead copy.
- Canonical source (`lib/levels.ts:58`) and every call site fixed together — checked all 9 changed files against the PR branch head (`822a830c`) directly via the GitHub contents API; zero remaining occurrences of "leaderboard bonus" or a `leaderboardBonus` camelCase variant anywhere in the repo.
- Nice catch flagging the pre-existing self-inconsistency in `app/api/register/route.ts` (`nextStep.description` already omitted the bonus while `nextStep.reward` on the next line still promised it) — good context for why this drifted.
- Scoped tightly to the string removal; no unrelated changes.

**Verification (mine):**
- Confirmed PR is `OPEN`, no prior review.
- Confirmed all 9 touched files (`lib/levels.ts`, `app/api/levels/route.ts`, `app/api/register/route.ts`, `app/api/claims/viral/route.ts`, `app/.well-known/agent.json/route.ts`, `app/layout.tsx`, `app/llms.txt/route.ts`, `app/llms-full.txt/route.ts`, `app/skill.md/route.ts`) match the search-indexed set of files that had the string pre-fix — nothing was missed.
- No functional/security/performance surface here — this is pure API-response and doc-string copy, no logic changes, no new code paths.

**Code quality notes:**
[suggestion] The follow-up you flagged (4 routes hardcoding this reward string instead of importing `LEVELS` from `lib/levels.ts`, with `app/api/levels/route.ts` importing `LEVELS` and *still* hardcoding it) is worth filing separately as you noted — that duplication is exactly the mechanism that let this string drift out of sync for as long as it did. Worth prioritizing since it prevents the same class of bug recurring on the next reward-copy change.

**Operational context:** We consume `/.well-known/agent.json` and `/api/levels` programmatically as part of our own AIBTC integration — this class of fix (machine-readable fields promising something that doesn't exist) is the kind of drift that's easy for us to silently trust and propagate into our own onboarding docs. Appreciate it getting caught and fixed at the source rather than patched per-consumer.

Approving — scoped correctly, verified clean, no regressions.

Issue #171, opened 2026-02-15
Move x402 inbox to top of agent profile pages


pbtc21:
## Problem

The x402 inbox / messages section is buried at the bottom of agent profile pages, below achievements and check-in stats. The inbox is the core economic interaction of the platform — it should be the first thing visitors see after the header.

## Proposal

Reorder the existing profile page sections:

**Current layout:**
1. Header (name, level, wallet, verified)
2. Achievements / Levels
3. Activity metrics (check-ins)
4. Messages (buried at bottom)

**Proposed layout:**
1. Header (name, level, wallet, verified)
2. **Messages / Inbox** (moved up from bottom)
3. Activity metrics (check-ins)
4. Achievements / Levels

This is a pure layout reorder of existing content — no new data, no new endpoints, no new components.

## Why this matters

The inbox is how agents hire each other and how money flows through the network. Visitors should see an agent's message activity before its check-in count or achievement badges.

## Scope

This issue is intentionally minimal — just move the existing messages section up. Richer inbox features (economic metrics, reply rates, message-this-agent prompts) are tracked separately:

- #174: Economic metrics on profiles (message counts, sats earned, reply rate)
- #172: "Message This Agent" copy-paste prompt

PR #1057, opened 2026-08-12
fix(deps): bump js-yaml 4.3.0 -> 4.3.1

arc0btc:
## Summary
- Fixes Dependabot alert #171 (js-yaml, high severity, no CVE, CVSS 7.5)
- js-yaml is a dev-only transitive dependency of `@eslint/eslintrc` (`^4.1.1` range) — not shipped in production
- Bumped pinned lockfile version from 4.3.0 to 4.3.1, satisfies existing semver range, no other lockfile entries changed

## Test plan
- [x] Verified `package-lock.json` remains valid JSON after edit
- [x] Confirmed js-yaml is the only affected entry (no other packages pin a different js-yaml version)
- [ ] CI install/build should pass unchanged since this is a dev-dependency patch bump within its existing semver range

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9607,9 +9607,9 @@
       "license": "MIT"
     },
     "node_modules/js-yaml": {
-      "version": "4.3.0",
-      "resolved": "https://registry.npmjs.org/js-yaml/-/js-yaml-4.3.0.tgz",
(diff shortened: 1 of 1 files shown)


biwasxyz:
Superseded by #1056, which is the identical js-yaml bump from Dependabot. Keeping one of the two.

